PR #8, opened 2020-01-03
Pdf generation

stuij:
- changed .rst files to make them fit to generate pdfs from with rst2pdf
- also did some stylistic changes to the .rst files to make them more alike

### DIFF
--- a/abi/aapcs64/aapcs64.rst
+++ b/abi/aapcs64/aapcs64.rst
@@ -3,21 +3,22 @@
    CC-BY-SA-4.0 AND Apache-Patent-License
    See LICENSE file for details
 
-*****
-|arm|
-*****
+Procedure Call Standard for the Arm® 64-bit Architecture (AArch64)
+******************************************************************
 
-.. |arm| raw:: html
-
-   <div align="center">
-      <img src="Arm_logo_blue_150MN.png" /><br/>
-      Procedure Call Standard for the Arm® 64-bit Architecture (AArch64)
-   </div>
-   <br />
+.. class:: issued
 
 Date of Issue: 31\ :sup:`st` October 2019
 
-================================================================================
+.. class:: logo
+
+.. image:: ../Arm_logo_blue_150MN.png
+
+.. section-numbering::
+
+.. raw:: pdf
+
+   PageBreak oneColumn
 
 Preamble
 ========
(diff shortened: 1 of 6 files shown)


stuij:
Just found out that Github actions won't work with legacy accounts, so we'll need to revert to something like Travis for the actual generation of the docs.

abi/rst2pdf-abi.style
```diff
@@ -0,0 +1,132 @@
+"pageSetup" : {
```
fpetrogalli: I wonder if it is worth moving anything "PDF related" under `abi/tools/pdf`, or similar? In there, it would be useful to have also a script that can be used to invoke rst2pdf

abi/aapcs64/aapcs64.rst
```diff
@@ -43,6 +44,9 @@ Please report defects in this specification to the `issue tracker page
 on GitHub
 <https://github.com/ARM-software/software-standards/issues>`_.
 
+.. raw:: pdf
+
+   PageBreak
```
fpetrogalli: Are the page breaks needed? The headers are numbered, does it really matter if we don't start chapters on new pages?

stuij: Maybe it's my years on years familiarity with dead-tree formats, or maybe I just feel a strong need to adhere to printed text layout, but In my opinion page breaks make a pdf a lot cleaner. It gives you a mental break of sorts.

abi/rst2pdf-abi.style
stuij: Agreed, this would be nicer. I'll do this.

abi/aapcs64/aapcs64.rst
fpetrogalli: I'd like to see a version without the page breaks before deciding :)

abi/rst2pdf-abi.style
stuij: done. see the updated commit.

abi/aapcs64/aapcs64.rst
stuij: GitHub seems broken. I can't seem to add the pdf on Windows or on Linux. I guess I'll have to send you it over email for now.

I must say, after re-evaluating, I'm still not a fan at all of no page breaks in pdfs.

fpetrogalli: Yep, so, if you are not a fan at all of no page breaks, you get my blessing :)

PR #10, opened 2020-01-22
clarify frame chain behaviour when PAC is enabled

stuij:
The text describing the frame chain in the AAPCS64 was written before PAC was added to the architecture.  For security purposes, when PAC is enabled, the return addresses need to be signed, but the AAPCS does not state this.

If we lacked such a requirement then the frame records would potentially contain a set of addresses that were unverifiable for the purposes of security and might be exploitable in terms of forming a ROP attack.

Code reading the frame chain obviously also has to be aware of this and deliberately strip the encoding in order to identify the real return address (it generally will not be able to unsign the address as the seed – the SP value – needed for this is not recorded in the frame).


rearnsha:
LGTM

PR #12, opened 2020-01-24
[abi][vfabia64] Specify the value of inactive lanes in output.

fpetrogalli:
The issue has been discussed in
https://github.com/ARM-software/software-standards/issues/2

### DIFF
--- a/abi/vfabia64/vfabia64.rst
+++ b/abi/vfabia64/vfabia64.rst
@@ -1489,6 +1489,12 @@ Masked signature generation for ``[not]inbranch`` clause.
 |``#pragma omp declare simd notinbranch`` |No     |Yes     |Yes   |No      |
 +-----------------------------------------+-------+--------+------+--------+
 
+Return value and masking
+------------------------
+
+The inactive lanes of return values of masked functions is ``undef``.
+
+
 Advanced SIMD masking
 ---------------------
 
--- a/abi/vfabia64/vfabia64.rst
+++ b/abi/vfabia64/vfabia64.rst
@@ -1492,9 +1492,9 @@ Masked signature generation for ``[not]inbranch`` clause.
 Inactive lanes
 --------------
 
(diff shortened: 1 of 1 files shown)


abi/vfabia64/vfabia64.rst
```diff
@@ -1489,6 +1489,12 @@ Masked signature generation for ``[not]inbranch`` clause.
 |``#pragma omp declare simd notinbranch`` |No     |Yes     |Yes   |No      |
 +-----------------------------------------+-------+--------+------+--------+
 
+Return value and masking
+------------------------
+
+The inactive lanes of return values of masked functions is ``undef``.
```
rsandifo-arm: `undef` sounds like LLVM terminology and isn't used elsewhere in the doc.  Maybe just "are undefined"?

Should we document this for arguments too?

fpetrogalli: Done.

abi/vfabia64/vfabia64.rst
```diff
@@ -1489,6 +1489,13 @@ Masked signature generation for ``[not]inbranch`` clause.
 |``#pragma omp declare simd notinbranch`` |No     |Yes     |Yes   |No      |
 +-----------------------------------------+-------+--------+------+--------+
 
+Inactive lanes
+--------------
+
+The values of the inactive lanes of the input parameters and the
```
rsandifo-arm: Sorry to nit-pick, but maybe:

In a masked vector function, the contents of the inactive lanes of the input parameters and the inactive lanes of the return value are undefined.

fpetrogalli: Done

fpetrogalli:
@jakubjelinek, we are in the process of adding the behavior introduced in this patch to the vector function ABI for AArch64. Do you foresee any problems in term of `libgomp` or the OpenMP standard?

PR #18, opened 2020-02-11
change styling for aapcs64 changelog as it's 3 columns now

stuij:


fpetrogalli:
Conflicts! :)

I don't think you need anyone's review for such changes, as long as the PDFs are building correctly.

So when you are done in fixing the conflict, I am happy for you to merge this directly.

PR #22, opened 2020-04-03
Extend the AAPCS64 to support SVE types and registers

rsandifo-arm:
This patch imports the SVE parts of the AAPCS64 from:

   https://developer.arm.com/docs/100986/0000

Most of the text is copied verbatim.  However, I took the opportunity
to replace the term "Scalable Vector Aggregate" with "Pure Scalable Type",
which includes individual predicates and vectors as well as aggregates.
This avoids having to refer to "Scalable Vector Types, Scalable Predicate
Types and Scalable Vector Aggregates" in various places.

I also tried to clarify exactly which composite types are pure
scalable types.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -205,7 +205,7 @@ Change History
 |            |                    |    Model`_, `APPENDIX Support for Advanced SIMD Extensions`_)    |
 |            |                    |                                                                  |
 |            |                    | 5. Update C++ mangling to reflect existing practice              |
-|            |                    |    (`C++ Mangling`_).                                            |
+|            |                    |    (`APPENDIX C++ Mangling`_).                                   |
 |            |                    |                                                                  |
 |            |                    | Minor changes:                                                   |
 |            |                    |                                                                  |
@@ -296,9 +296,16 @@ SIMD and floating point
    the floating point instruction set, the SIMD instruction set and the
    register set shared by them.
 
+SVE
+   The Arm architecture's Scalable Vector Extension.
+
 T32
    The instruction set named Thumb in the Armv7 architecture; T32 uses
    16-bit and 32-bit instructions.
+
+VG
+   The number of 64-bit “vector granules” in an SVE vector; in other words,
+   the number of bits in an SVE vector register divided by 64.
 
 ILP32
    SysV-like data model where int, long int and pointer are 32-bit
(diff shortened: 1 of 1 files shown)


aapcs64/aapcs64.rst
```diff
@@ -521,6 +538,35 @@ Elements in a short vector are numbered such that the lowest numbered element (e
...
 
+Scalable Vectors
+----------------
+
+Like a short vector (see `Short Vectors`_), a scalable vector is a
+machine type that is composed of repeated instances of one fundamental
+integral or floating-point type. The number of bytes in the vector is
+always VG×8, where VG is a runtime value determined by the execution
+environment. VG is an even integer greater than or equal to 2; the ABI
+does not define an upper bound. VG is the same for all scalable vector
```
fpetrogalli, line 549:
```suggestion
does not define an upper bound. VG is the same for all scalable vectors
```

aapcs64/aapcs64.rst
```diff
@@ -521,6 +538,35 @@ Elements in a short vector are numbered such that the lowest numbered element (e
...
+and predicate types.
```
fpetrogalli, line 550: I might be missing something, but what is a predicate type? If with that you mean scalable vector, maybe it is better to tick with the name used in the next section?

aapcs64/aapcs64.rst
```diff
@@ -521,6 +538,35 @@ Elements in a short vector are numbered such that the lowest numbered element (e
...
+register, the least significant bit of element 0 occupies bit 0
+of the corresponding short vector register. Note that the layout of the
+vector in a scalable vector register does not depend on whether the
+system is big- or little-endian.
+
+Scalable Predicates
+-------------------
+
+A scalable predicate is a machine type that is composed of VG bytes,
+with one bit for each byte of a scalable vector.
```
fpetrogalli, line 565: May I suggest to add an example showing a scalable vector with VGx8 bytes and the correspondent scalable predicate of VG bytes, with bytes corresponding to bits? It could help answering the question that will very likely raise when reading the specs "why is it one bit for each byte of a scalable vector". Or does this goes beyond this specs?

aapcs64/aapcs64.rst
```diff
@@ -623,6 +669,46 @@ Homogeneous Short-Vector Aggregates (HVA)
...
+
+* an aggregate in which every member is a Pure Scalable Type.
+
+As with Homogeneous Aggregates, these rules apply after data layout is
+completed and without regard to access control or other source language
+restrictions.  However, there are several notable differences from
+Homogeneous Aggregates:
+
+* A Pure Scalable Type may contain a mixture of different Fundamental
+  Data Types.  For example, an aggregate that contains a scalable vector
```
fpetrogalli, line 692:
```suggestion
  Data Types. For example, an aggregate that contains a scalable vector
```

aapcs64/aapcs64.rst
```diff
@@ -623,6 +669,46 @@ Homogeneous Short-Vector Aggregates (HVA)
...
+  of 8-bit elements, a scalable predicate, and a scalable vector of
+  16-bit elements is a Pure Scalable Type.
+
+* Alignment and padding do not play a role when determining whether
+  something is a Pure Scalable Type.  (In fact, a Pure Scalable Type
```
fpetrogalli, line 697:
```suggestion
  something is a Pure Scalable Type. (In fact, a Pure Scalable Type
```

aapcs64/aapcs64.rst
rsandifo-arm: "scalable vector and predicate types" == scalable vector types and scalable predicate types.  But yeah, I agree that kind of shorthand probably isn't appropriate here.  I'll write it out in full.

aapcs64/aapcs64.rst
rsandifo-arm: Yeah, I think that's out-of-scope.  The way that the bits relate to bytes when vectors and predicates are used together is really an architecture-level thing.

aapcs64/aapcs64.rst
rsandifo-arm: Fixed by expanding "scalable vector and predicate types" to "scalable vector types and scalable predicate types".

aapcs64/aapcs64.rst
rsandifo-arm: Done.

aapcs64/aapcs64.rst
rsandifo-arm: The paragraph now only talks about the link between predicates and vectors in terms of their size, to avoid the previous (unintended) implication that it was saying how each bit maps to each byte.

aapcs64/aapcs64.rst
```diff
@@ -623,6 +672,46 @@ Homogeneous Short-Vector Aggregates (HVA)
...
+A type is a Pure Scalable Type if (recursively) it is:
+
+* a Scalable Vector Type;
+
+* a Scalable Predicate Type;
+
+* an array that contains a constant (nonzero) number of elements and whose
+  Base Type is a Pure Scalable Type; or
+
+* an aggregate in which every member is a Pure Scalable Type.
```
stuij, line 687: I'd remove the semicolons and the or in this list. Looks cleaner.

```suggestion
* a Scalable Vector Type

* a Scalable Predicate Type

* an array that contains a constant (nonzero) number of elements and whose
  Base Type is a Pure Scalable Type

* an aggregate in which every member is a Pure Scalable Type
```

aapcs64/aapcs64.rst
```diff
@@ -635,7 +724,7 @@ The base standard defines a machine-level calling standard for the A64 instructi
 Machine Registers
 -----------------
 
-The Arm 64-bit architecture defines two mandatory register banks: a general-purpose register bank which can be used for scalar integer processing and pointer arithmetic; and a SIMD and Floating-Point register bank.
+The Arm 64-bit architecture defines two mandatory register banks: a general-purpose register bank which can be used for scalar integer processing and pointer arithmetic; and a SIMD and Floating-Point register bank. In addition, the architecture defines an optional set of scalable vector registers that overlap the SIMD and Floating-Point register bank, accompanied by a set of scalable predicate registers.
```
stuij, line 727: Not your change, but I can't help my inner language-stickler to comment: That semicolon shouldn't be there. I don't think that's how semicolons are used.

aapcs64/aapcs64.rst
```diff
@@ -716,6 +805,40 @@ The FPCR is used to control the behavior of the floating-point unit. It is a glo
...
+The Arm 64-bit architecture also defines an optional set of thirty-two
+scalable vector registers, z0-z31. Each register extends the
+corresponding SIMD and Floating-Point register so that it can hold the
+contents of a single Scalable Vector Type (see `Scalable Vectors`_).
+That is, scalable vector register z0 is an extension of SIMD and
+Floating-Point register v0.
+
+z0-z7 are used to pass scalable vector arguments to a subroutine, and to
+return scalable vector results from a function. If a subroutine takes
+arguments in scalable vector or predicate registers, or if it is a
```
stuij, line 820: I'm tripping over this wording. I'd just be verbose about  it:

```suggestion
arguments in scalable vector registers or predicate registers, or if it is a
```

aapcs64/aapcs64.rst
```diff
@@ -716,6 +805,40 @@ The FPCR is used to control the behavior of the floating-point unit. It is a glo
...
+
+The Arm 64-bit architecture defines an optional set of sixteen scalable
+predicate registers p0-p15. These registers are available if and only if
+the scalable vector registers are available (see `Scalable Vector Registers`_).
+Each register can store the contents of a Scalable Predicate Type
+(see `Scalable Predicates`_).
+
+p0-p3 are used to pass scalable predicate arguments to a subroutine and
+to return scalable predicate results from a function. If a subroutine
+takes arguments in scalable vector or predicate registers, or if it is a
```
stuij, line 837: idem
```suggestion
takes arguments in scalable vector registers or predicate registers, or if it is a
```

aapcs64/aapcs64.rst
```diff
@@ -1468,10 +1613,74 @@ Table 7: Short vector extended types
...
 +-----------------+-------------------+--------------------------+-----------+
 
-C++ Mangling
-------------
+APPENDIX Support for Scalable Vectors
+=====================================
+
+As an optional extension, the AArch64 architecture supports a number of
+scalable-vector operations. To facilitate accessing these operations
+from C and C++ a number of extended types need to be added to the
```
stuij, line 1621: I'd add a comma.
```suggestion
from C and C++, a number of extended types need to be added to the
```

aapcs64/aapcs64.rst
```diff
@@ -1468,10 +1613,74 @@ Table 7: Short vector extended types
...
+language.
+
+Following the conventions used for adding types to C99, and the
+conventions established in `APPENDIX Support for Advanced SIMD Extensions`_,
+a number of additional types (internal types) are defined
+unconditionally. To facilitate use in applications a header file
+``arm_sve.h`` is also defined that maps these internal types onto more
```
stuij, line 1628: Also what? I'd remove the also.
```suggestion
``arm_sve.h`` is defined that maps these internal types onto more
```

aapcs64/aapcs64.rst
rsandifo-arm: I'd prefer to keep it as-is to be honest.  This is a relatively common style and makes the conjunction ("or") more obvious.  But maybe I've just been reading too many legal documents :-)

aapcs64/aapcs64.rst
rsandifo-arm: That's probably true, but can we fix it separately?  This is quite a big change as it is :-)

aapcs64/aapcs64.rst
rsandifo-arm: OK, will change.  Note that this is deliberately apeing the style for the Advanced SIMD stuff.

aapcs64/aapcs64.rst
rsandifo-arm: OK.  This too is a carry-over from the Advanced SIMD stuff.

aapcs64/aapcs64.rst
rsandifo-arm: Done.

aapcs64/aapcs64.rst
rsandifo-arm: Done.

aapcs64/aapcs64.rst
stuij: sure, fine :)

aapcs64/aapcs64.rst
stuij: yup, of course.

stuij:
LGTM :) Tnx!

PR #23, opened 2020-05-21
Add requirements for stack space with MTE tags.

mmalcomson:
There are two salient points we need to ensure the stack conforms to:

1. There is a single "stack background tag" that is set on the unused
   part of the stack.

2. All stack space that is not allocated at public interfaces must have
   their tag set to that "stack background tag".

Note:
  This second requirement implicitly means that landing pads may not
  change the stack pointer before calling back into _Unwind_Resume
  unless they handle setting the tags on the stack space that is now
  below the stack pointer.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -285,6 +285,9 @@ Q-o-I
    by systems conforming to it.  Q-o-I is often used to describe the
    tool-chain-specific means by which a standard requirement is met.
 
+MTE
+   The Arm architecture's Memory Tagging Extension.
+
 SIMD
    Single Instruction Multiple Data – A term denoting or qualifying:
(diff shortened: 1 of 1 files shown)


mmalcomson:
> I'd rather think this condition should be stronger, i.e. at all times, not only at a public interface.
> We have always assumed, for example, that an interrupt handler could write to the un-allocated portion of the stack, that would be an issue if the compiler did not enclose all tag manipulation between SP increments/decrements.

That's a good point -- thanks for the catch!

I will update the PR -- though the text I added doesn't mention at a public interface it's placed in a "public interface" section.

mmalcomson:
I have moved the additional text into the "Universal stack constraints" section and updated the commit message.
I believe mentioning "at all time" in the text is moot given the section it's now in.

PR #27, opened 2020-09-04
collate sections common to the different documents in a single dir

stuij:
I know that in spirit we should have different directories for the different documents. However it would help maintainability greatly if we can put common pieces of the text in a common directory. This would especially help the parts of the text where we have version numbers or changing dates.

I also moved around some parts of the vfabi64 to be in line with the rest of the ABI documents.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -6,23 +6,7 @@
 Procedure Call Standard for the Arm® 64-bit Architecture (AArch64)
 ******************************************************************
 
-.. class:: version
-
-2020Q2
-
-.. class:: issued
-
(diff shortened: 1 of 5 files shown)


stuij:
This would also help converting new documents, as it's quite a bit less confusing to know what sections to change where, especially for new contributors.

stuij:
Hmm, not sure if I'm sold on this myself to be honest. Another option is to just source in the version and date sections, which look pretty out of place in the documents right now and are Arm-release specific anyways.

stuij:
I'm going to abandon this. I thought of a less ideal version from an engineering standpoint, but a workable one keeping in mind we want to keep the different documents separate for license reasons.

PR #28, opened 2020-09-22
[PAuthABI] Add initial version of the Pauth ELF extensions

smithp35:
The PAuth ABI is an alpha specification for using the Armv8.3-A Pauth
extension without needing to be backwards compatible with the ABI for the
64-bit Arm Architecture.

This document describes the extensions to ELF. It contains relocation
directives and an encoding for the signing schema needed for the run-time
to sign pointers. This document does not contain a mapping from any source
language to signing schema, this will be provided in a separate document.


stuij:
LGTM

PR #41, opened 2020-10-02
Make encoding of signing schema and evaluation of relocation concrete

smithp35:
Moved encoding options to an appendix. Made the proposal concrete and wrote out the arm64e rules for evaluating an AUTH relocation.


pauthabielf64/pauthabielf64.rst
```diff
@@ -542,243 +586,191 @@ by the run-time environment. There are a number of possible encoding
...
+| auth | address diversity | reserved | key      | reserved | discriminator | reserved for addend |
++------+-------------------+----------+----------+----------+---------------+---------------------+
 
-In ELF we have the following places where we can encode this
-information via a combination of.
+* ``auth`` is a single bit to denote that the value represents metadata
 
-* The relocation code.
+  * Review Question: Do we need this? It should only be present when
+    there is an AUTH variant dynamic relocation.
```
pcc, line 614: I don't think we need it. I never had a need for it in my prototype.

pauthabielf64/pauthabielf64.rst
```diff
@@ -542,243 +586,191 @@ by the run-time environment. There are a number of possible encoding
...
+this should be sufficient.
 
-* The constant discriminator value.
+.. class:: pauthabiencoding
 
-* Whether to combine address diversity with the discriminator.
++------+-------------------+----------+----------+----------+---------------+---------------------+
+| 63   | 62                | 61       | 60:59    | 58:48    |  47:32        | 31:0                |
++======+===================+==========+==========+==========+===============+=====================+
+| auth | address diversity | reserved | key      | reserved | discriminator | reserved for addend |
```
pcc, line 608: Should the specification require these reserved bits to be set to 0 and recommend that consumers of the field require the reserved bits to be set to 0? This should allow them to be safely repurposed later.

pauthabielf64/pauthabielf64.rst
```diff
@@ -542,243 +586,191 @@ by the run-time environment. There are a number of possible encoding
...
-following GOT generating relocations must be used. As the linker
-generates the pointer in the GOT the contents of the place of the
-relocation is an instruction so the signing schema cannot be encoded
-in the contents of the place. The relocation addend is used instead.
+This is the equivalent of the arm64e ARM64_RELOC_AUTHENTICATED
+relocation. It can also be used as a dynamic relocation.
 
-``ADDEND(A)`` extracts the addend field from the bottom 32-bits of the
-64 bit RELA addend field (signextend64, A & 0xffffffff)
+AUTH variant GOT Generating Relocations
```
pcc, line 679: I still think we don't need to specify these because it is possible for the compiler to create these GOT entries by itself and rely on the linker to deduplicate them.

pauthabielf64/pauthabielf64.rst
```diff
@@ -397,100 +399,130 @@ General Principles
...
-* If the GOT is signed there seems little advantage to the explicit
-  schema, with the disadvantage of added complexity in the tools to
-  support the new relocations. There is a possibility of a hybrid
-  model where data pointers follow an implicit signing schema and
-  function pointers in the GOT use the explicit schema. Feedback is
-  requested on whether it is necessary to sign the GOT and whether an
-  implicit signing schema would be sufficient.
+
+Signed GOT entries use the ``IA`` key for symbols of type ``STT_FUNC``
+and the ``DA`` key for symbols of , with the address of the GOT entry
```
rjmccall, line 462: Incomplete sentence

pauthabielf64/pauthabielf64.rst
smithp35: Thanks, I'll update and remove it.

pauthabielf64/pauthabielf64.rst
smithp35: That seems reasonable, I expect that in the early stages we'll be recompiling everything but it won't hurt to be more specific. I'll add some text.

pauthabielf64/pauthabielf64.rst
smithp35: The existing GOT generating relocations could be used in this context but this does require the compiler and linker to agree on whether the GOT is signed, and what subset of the GOT is signed. In a draft I had 3 options:

- No signing of the GOT (default)
- Code pointers in GOT are signed
- Code and Data pointers are signed
We then need a mechanism to communicate the option from the compiler to the static linker and there isn't a good existing option, a prototype could use a command line option but ideally this would be automatic. 

The .note.gnu.property isn't ideal for this purpose as this was intended to record architectural properties and doesn't handle anything other than feature bits that are subject to And or Or operations. 

Something like AArch32 BuildAttributes could work but that would mean specifying them.

Using AUTH GOT generating relocations handles the communication without needing some kind of binary marking.
- Unsigned GOT = usual GOT generating relocations
- Code pointers = mix of AUTH and usual GOT generating relocations
- Code and Data pointers = AUTH GOT generating relocations

Hope I haven't misunderstood your comment?

I think we may be lucky and the initial adoption doesn't need any kind of signed GOT. Experiments could use a command line option.

pauthabielf64/pauthabielf64.rst
smithp35: Thanks for pointing that out, I'll fix in the next update. I'm proposing IA for STT_FUNC and DA for everything else (STT_OBJECT, STT_NONE , STT_SECTION etc.)

pauthabielf64/pauthabielf64.rst
```diff
@@ -503,21 +540,36 @@ instructions is:
...
 
-As the signing schema used for the GOT can affect code generation, and
-there is a use case for both a signed and unsigned GOT the ELF ABI
-defines mechanisms to do both. It is expected that platforms will
-mandate a single choice, although per link-unit choices are
-possible. It is not possible to mix mechanisms within the same
-link-unit as there is at most one GOT entry per symbol.
+The PAuth ABI adds an additional Processor specific section type
+
+.. table:: ELF Section Types
```
stuij, line 563: in the generated pdf the columns of the table are just off, causing the name and variable columns to cut off the text

pauthabielf64/pauthabielf64.rst
```diff
@@ -542,167 +594,360 @@ by the run-time environment. There are a number of possible encoding
...
+field of the relocation.
+
+The top 32-bits of the contents of the place is used to encode the
+signing schema for both static and dynamic relocations. This permits
+platforms using relocation compression or SHT_REL dynamic relocations
+to encode relocation addends in the bottom 32-bits. Given that the
+maximum size of link-units using the small code-model is 4 gigabytes
+this should be sufficient.
+
+.. class:: pauthabiencoding
```
stuij, line 611: if you're not going to use the class, you might as well delete it. also it's not adhering to the unofficial scheme: nameofdoc-description

pauthabielf64/pauthabielf64.rst
```diff
@@ -542,167 +594,360 @@ by the run-time environment. There are a number of possible encoding
...
+
+* ``address diversity`` is a single bit that when set, denotes that
+  the pointer has address diversity. The place (relocation target
+  address) will be blended with the discriminator value.
+
+* ``key`` determines the key to be used. Armv8.3-A specifies 5 keys, 4
+  of which can be used by PAUTHELF64. The generic key ``APGA`` is not
+  repesented at the ELF ABI level.
+
+  .. class:: pauthkeyencoding
```
stuij, line 627: same as above comment

pauthabielf64/pauthabielf64.rst
```diff
@@ -713,7 +958,7 @@ in the contents of the place. The relocation addend is used instead.
 ``ENCD(value, schema)`` is the encoding of the signing schema into the
 GOT slot.
 
-.. class:: pauthabielf64-signing-schema
+.. class:: pauthabielf64-signing-schema-alternative
```
stuij, line 961: the class isn't actually used. since it looks like it has the same dimensios as the other table, perhaps the same class, `pauthabielf64-signing-schema`, will render it equally well?

pauthabielf64/pauthabielf64.rst
```diff
@@ -3,7 +3,7 @@
    CC-BY-SA-4.0 AND Apache-Patent-License
    See LICENSE file for details
 
-.. |release| replace:: 0.1
+.. |release| replace:: 0.2
 .. |date-of-issue| replace:: 1\ :sup:`st` October 2020
 .. |copyright-date| replace:: 2020
```
stuij, line 9: I see I forgot to change the footer markup in this doc. I'll do that myself at some point.

pauthabielf64/pauthabielf64.rst
smithp35: Thanks for pointing that out, I think I've managed to find a way of getting to render correctly by using `` `` quotes.

pauthabielf64/pauthabielf64.rst
smithp35: I've deleted all of the .. class:: as I don't think I use any of them.

pauthabielf64/pauthabielf64.rst
smithp35: As above I've deleted the class.

pauthabielf64/pauthabielf64.rst
smithp35: I can't find any documentation on what .. class:: actually does so given my limited understanding I'm going to remove them all for now.

stuij: We use class to tell rst2pdf what the width of the columns should be. There's a style-sheet used when building: `tools/rst2pdf/rst2pdf-abi.style`. I think I should save us both some time and fix up the added tables myself in a follow-up commit.

PR #43, opened 2020-10-19
add DWARF for the Arm 64-bit Architecture document

walkerkd:


aadwarf64/aadwarf64.rst
```diff
@@ -0,0 +1,668 @@
...
+
+   +-----------------------------+----------+
+   | Sub-operation               | Code     |
+   +=============================+==========+
+   | ``DW_SUB_OP_AARCH64_sign``  | ``0x00`` |
+   +-----------------------------+----------+
+
+The ``DW_SUB_OP_AARCH64_sign`` sub-operation takes a single operand
+encoded as an unsigned LEB128 operand. This value specifies a pointer
+key signing operation given in `aarch64 dwarf pointer signing operations`. The top
```
stuij, line 601: I would either rephrase `aarch64 dwarf pointer signing operations` to `the table below` or I would make it a proper link. Currently when reading this it's unclear what it refers to.

Also, I'm not sure what the quotes do around `aarch64 dwarf pointer signing operations`. They don't do anything in the html doc in any case.

philgee-oss: I assume that aarch64 dwarf pointer signing operations should be in italics. if so, then it needs to be wrapped with `:title:`aarch64 dwarf pointer signing operation`` or `*aarch64 dwarf pointer signing operation*`

stuij: `aarch64 dwarf pointer signing operation` is meant to refer to the table below. In the previous documents on developer.arm.com it is a link to the table. Just rendering it differently still keeps it vague what the term is supposed to refer to. So I would still suggest either making it a link to the table again or rephrasing it to make clear you're referring to the table.

walkerkd: Made it a link and slightly modified the wording to reflect that it is referencing a table.

stuij (approved):
LGTM
Thanks so much!

PR #46, opened 2020-10-27
Add DWARD register 32/PC.

DanielKristofKiss:
PC was a resereved register because it is not directly accessible
on the architecture but in special cases it need to be restored
in alternative ways.

### DIFF
--- a/aadwarf64/aadwarf64.rst
+++ b/aadwarf64/aadwarf64.rst
@@ -362,7 +362,8 @@ integers.
    +----------------+--------------------------+----------------------------------------------------+
    | 31             | SP                       | 64-bit stack pointer                               |
    +----------------+--------------------------+----------------------------------------------------+
-   | 32             | Reserved                 | \-                                                 |
+   | 32             | PC                       | 64-bit program counter                             |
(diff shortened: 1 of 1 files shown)


aadwarf64/aadwarf64.rst
```diff
@@ -362,7 +362,8 @@ integers.
    +----------------+--------------------------+----------------------------------------------------+
    | 31             | SP                       | 64-bit stack pointer                               |
    +----------------+--------------------------+----------------------------------------------------+
-   | 32             | Reserved                 | \-                                                 |
+   | 32             | PC                       | 64-bit program counter                             |
```
nsz-arm, line 365: how is this handled by existing unwinders? will they crash or ignore the related dwarf operations.

DanielKristofKiss: LLVM/libunwind and savannah/libunwind already accepts it as any other register. gcc/libgcc doesn't and I got pushback on the change without spec change.

walkerkd: As there have been no obections to the change, and the one question asked has been answered I am going to accept this pull request.

PR #51, opened 2020-11-12
CLIBABI32, RTABI32: redraw interoperability diagram in SVG.

statham-arm:
The previous PNG was low-resolution, and also low-quality (the lines
were antialiased in a way that suggested it had been rescaled at least
once in its lifetime). I've redrawn it in Inkscape, as a nicely
scalable SVG file.

rst2pdf can import SVG as long as you have the 'svglib' Python module
available. Unfortunately, its SVG import doesn't seem to implement the
feature Inkscape uses to put arrowheads on the ends of lines. So I've
had to work around that by a round-trip conversion through EPS (which
lacks that feature, forcing Inkscape to convert it to something more
primitive).

Each ABI directory now contains two SVG files. The one with '-source'
in the name, e.g. rtabi32/rtabi32-inter-operation-source.svg, is the
preferred form for modification: that's the one that will look
sensible if you load it into Inkscape and try to edit it. The other
one, e.g. clibabi32/clibabi32-inter-operation.svg, is a plain SVG
conversion suitable for use by rst2pdf, and is the one referred to by
the 'figure' command in the RST.

I've committed the converted versions into source control, to avoid
requiring a complicated Makefile and introducing Inkscape as a
build-dependency. But for anyone who needs to edit the documents, I've
provided a script tools/inkscape/export-svg.sh to do the conversion
between the two formats.


stuij:
For rst2pdf to be able to process SVG files, the Python svglib module needs to have been installed. Could you update tools/rst2pdf/install.sh to add it to the pip install line?

statham-arm:
I had managed to completely miss the existence of that script. Ahem.

stuij:
Yea, totally understandable as it's hiding at the back in a corner.
LGTM. Thanks!

PR #53, opened 2020-11-23
Add Thread ID register numbers

walkerkd:


walkerkd:
This changes addes register number for the Thread ID registers for wach of the ELO to EL3 levels.
This was choosen over the choice of adding a single entry TPIDR_mode similar to the existing ELR_mode register number entry as it should allow easier implementation as it allows direct 1-to-1 mapping between the register used in the code and the register used in the DWARF description.

stuij:
LGTM. Thanks!

PR #59, opened 2020-12-10
[aapcs64-morello] add Morello extensions to PCS for AArch64 document

amilendra:


aapcs64-morello/aapcs64-morello.rst
```diff
@@ -0,0 +1,858 @@
...
+AAPCS64 and AAPCS64-cap have the same design goals as described in the AAPCS64_ document.
+
+An additional design goal for the capability-aware AAPCS64 is to support interworking with legacy AAPCS64 code, following only the Procedure Calling Standard described in the AAPCS64_                 document.
+
+Conformance
+-----------
+
+The AAPCS64 and AAPCS64-cap have the same conformance rules as defined in AAPCS64_, with the additional requirement that:
+
+- At each call where the control transfer instruction is subject to a BL-type relocation at static link time, rules on the use of r16 and r17 (CIP0, CIP1, IP0, IP1) are observed (`Design Goals`_).
```
stuij, line 266: I think this links to the wrong section. In the original document this links to "Use of CIP0 and CIP1 by the linker"

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -0,0 +1,858 @@
...
+    +-----------+----------+----------------------------------------------------------------------------------------------------+
+    | r8        |          | The indirect result location register.                                                             |
+    +-----------+----------+----------------------------------------------------------------------------------------------------+
+    | r0-r7     |          | Parameter/result registers.                                                                        |
+    +-----------+----------+----------------------------------------------------------------------------------------------------+
+
+
+The first eight registers, r0-r7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).
+
+Registers r16 (IP0/CIP0) and r17 (IP1/CIP1) may be used by a linker as a scratch register between a routine and any subroutine it calls (for details, see `Capabilities`_). They can also be used within a routine to hold intermediate values between subroutine calls.
```
stuij, line 396: It looks like the `Cababilities`_ link here is wrong. Perhaps done through a script?

amilendra: I did it manually. Looks correct to me because it goes to the 5.2 Capabilities section both in my local pdf and github rendering as expected.

stuij: But in the pdf of the original release up on developer, this isn't a link and reads: (for details, see aapcs64-section5-3-1)

Which I think must refer to section 6.3.1 of the aapcs64 spec up on Github: https://github.com/ARM-software/abi-aa/blob/master/aapcs64/aapcs64.rst#id55 (I wouldn't link to that though, I think), which actually deals with IP0/IP1, as opposed to the Capabilities section in this document.

amilendra: Ah yes, sorry I was looking at the table in section 5.1 which also links correctly to Capabilities.

The link you point out seems to be a bug in the original spec which linked to the non-existent aapcs64-section5-3-1 section, when it should have been using _aapcs64-morello-section5-3-1, which is for the section "Use of CIP0 and CIP1 by the linker". Will correct it.

stuij:
LGTM. Thanks!

PR #62, opened 2020-12-16
[AAPCS64] Put back missing instructions on how to interpret the tables

smithp35:
The Parameter passing rules contains 3 tables containing rules for each
of the 3 stages. In the original word document there were instructions
on how to interpret the rules. These have been missed out in the .rst
document.

Add the text back from the word document verbatim.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -986,7 +986,9 @@ For a caller, sufficient stack space to hold stacked argument values is assumed 
 
   +-----------------------------------------------------------------------------------------------------------------+
   | Stage A - Initialization                                                                                        |
-  +======================+==========================================================================================+
+  +=================================================================================================================+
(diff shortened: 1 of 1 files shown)


fpetrogalli (commented):
Shouldn't you create a new row in the `Change History` table for the `Next` Issue with `TDB` release date and a description of the change?

smithp35:
It depends on whether the change is considered large enough to warrant it. Certainly in the past the change-log hasn't been exhaustive although I'm happy to add one if you'd like.

fpetrogalli:
Oh, if the changelog is not supposed to get down to that level of details, I am happy for you to leave the patch as it is.

Ultimately, I'll let @stuij give a final word on the patch, but given that the paragraphs are quoted verbatim from https://developer.arm.com/documentation/100986/latest, I think you are good to go. LGTM 👍

stuij:
Yes, I'm also fine with not adding to change log.

LGTM :o)

PR #63, opened 2021-01-19
Minor proofreading & typos

pbarrio:
I just created all the minor changes into a single commit, but let me know if you want to separate them or remove some of the changes.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -549,7 +549,7 @@ generate one type of PLT entry in a link-unit.
 Section Types
 =============
 
-The PAuth ABI adds an additional Processor specific section type
+The PAuth ABI adds an additional Processor-specific section type
 
 .. table:: ELF Section Types
 
@@ -559,7 +559,7 @@ The PAuth ABI adds an additional Processor specific section type
   | ``SHT_AARCH64_AUTH_RELR`` | ``0x70000004`` | Section type for compressed signed relative relocations |
   +---------------------------+----------------+---------------------------------------------------------+
 
-The value is in the AArch64 Processor specfic range. The value is
+The value is in the AArch64 Processor-specific range. The value is
 subject to change if there is a clash with any section types added by
 AAELF64_.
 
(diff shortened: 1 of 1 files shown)


pauthabielf64/pauthabielf64.rst
```diff
@@ -754,13 +754,13 @@ Description:
 
    PageBreak
 
-Reocation Compression
+Relocation Compression
 =====================
```
smithp35, line 758: You'll need to extend the = to include the extra space (CI caught it).

pauthabielf64/pauthabielf64.rst
```diff
@@ -777,13 +777,15 @@ Static Linking
 ==============
 
 The static linker cannot create signed pointers, just as it cannot run
-constructors for static variables, but the C-runtime that runs before
-main can. The static linker must communicate the details of how to
+constructors for static variables, but the runtime that runs before
```
smithp35, line 780: How about "but the startup code that runs before the program (``main`` in C/C++) can?

Avoids "runtime that runs" which sounds a bit redundant.

There isn't really a good definition that I can point to. https://en.wikipedia.org/wiki/Entry_point uses the term library startup code and main as the start of the user program.

pauthabielf64/pauthabielf64.rst
```diff
@@ -777,13 +777,15 @@ Static Linking
...
+the user code (e.g. in C/C++, before ``main()``) can. The static
+linker must communicate the details of how to
 create the signed pointers by embedding the information in the ELF
-file. The format of the information is platform ABI as it is a
-contract between the static-linker and the C-runtime. One simple
+file. The format of the information is dependent on the platform ABI,
```
smithp35, line 784: I have a mild preference for "is defined by" than "is dependent on".

smithp35 (approved):
I've got a couple of alternative suggestion for a couple of the changes, but otherwise LGTM.

pbarrio:
All comments addressed.

stuij:
lgtm, merged

PR #67, opened 2021-02-15
[aapcs32] Clarify rule C.4 when there is an overaligned HFA

smithp35:
Rule C.4 stated that the NSAA is rounded up to the maximum of 8 and
the natural alignment of the HFA.

The original intent, and the GCC implmentation limits the alignment to
16 as this is the maximum alignment requirement of an AArch64 instruction.
Change rule C.4 to make it explicit that the alignment of NSAA is 16-bytes.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -1054,7 +1054,7 @@ For a caller, sufficient stack space to hold stacked argument values is assumed 
 
 .. class:: aapcs64-stage
 
-  +-----------------------+----------------------------------------------------------------------------------------+
+  +----------------------------------------------------------------------------------------------------------------+
   | Stage C – Assignment of arguments to registers and stack                                                       |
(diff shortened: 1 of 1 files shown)


momchil-velikov (approved):
LGTM

rearnsha (changes requested):
NSAA is a counter, not specifically bytes, so it only needs to say multiple of 8 (or 16).
'a multiple' is ambiguous, there are many multiples that would satisfy that requirement.  It should read 'next multiple' to avoid that ambiguity.

smithp35:
> NSAA is a counter, not specifically bytes, so it only needs to say multiple of 8 (or 16).
> 'a multiple' is ambiguous, there are many multiples that would satisfy that requirement. It should read 'next multiple' to avoid that ambiguity.

Thanks, I've updated a new revision.

rearnsha (approved):

PR #71, opened 2021-02-22
[AAELF64] Reserve relocation codes for PAuthABI

smithp35:
The Alpha PAuthABI defines a small number of static and dynamic relocations
in the private vendor experiment space. When an upstream implementation
lands we will want to use non-experimental codes. Step 1 is to reserve
a range of static and dynamic relocations in AAELF64 so we can guarantee
no clash with any relocations defined in the PAuthABI.


smithp35:
Managed to mess up the commits. Will resubmit under a different branch.

PR #76, opened 2021-02-25
Make diagrams and formulas look prettier

statham-arm:
The main commit here redraws all the remaining low-res blurry PNG diagrams as SVG. Also, a bonus commit correcting a couple of rogue `:math:` formulas I spotted in the process, which provoked rst2pdf warnings and must have been overlooked during previous conversions.


advnote132/README.md
```diff
@@ -1,5 +1,5 @@
 <div align="center">
-   <img src="Arm_logo_blue_150MN.png" />
+   <img src="Arm_logo_blue_RGB.svg" />
```
stuij, line 2: The advnote132, bsabi32 and dbgoverl32 documents are missing svgs. I'm guessing because they weren't merged yet when the first version of this patch were created.

statham-arm: Oops! I wasn't awake yesterday. I remembered to delete the bitmaps from those three directories, but not to add the SVG in its place.

aapcs64/aapcs64.rst
```diff
@@ -1787,13 +1790,13 @@ set to zero, permitting a simplified implementation of ``va_arg`` which simply a
 
 Although this standard does not mandate a particular stack frame organisation beyond what is required to meet the stack constraints described in `The Stack`_, the following figure illustrates one possible stack layout for a variadic routine which invokes the ``va_start`` macro.
 
-.. figure:: aapcs64-variadic-stack.png
+.. figure:: aapcs64-variadic-stack.svg
```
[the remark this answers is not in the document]
statham-arm: No, that wasn't on purpose, and I'm quite confused, because I _thought_ I was being careful about those prime signs!

I agree that they're right the original way round; in particular, the FP' stored on the stack in the new stack frame on the RHS ought to be the same as the one before the call on the LHS.

stuij:
LGTM. Thanks!

stuij:
hopefully last github ABI notification test

PR #79, opened 2021-03-04
sharpen up definition of the extent of the stack

stuij:
Currently, the SP is specified to lie between stack-base inclusive, and
stack-limit exclusive. Having the stack-limit be exclusive to the stack pointer
is confusing at best, so for clarity this patch changes the definition to be
stack-limit inclusive.

### DIFF
--- a/aapcs32/aapcs32.rst
+++ b/aapcs32/aapcs32.rst
@@ -1044,7 +1044,7 @@ Universal stack constraints
 
 At all times the following basic constraints must hold:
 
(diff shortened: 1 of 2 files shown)


rearnsha:
Given that there's no way of reading the stack limit in the process this is largely academic.  However, the original intent was that both values would be aligned to the stack alignment and this change breaks that.  So I'm not sure that is a good idea.

Wilco1:
Agreed, we should go ahead and finally merge this.

PR #80, opened 2021-03-05
[Morello] Describe the purecap TLS implementation in the ELF appendix.

sbaranga-arm:


aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+Static TLS block resolver
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If the TLS variable is in the static block, while resolving the
+``R_MORELLO_TLSDESC`` relocation, the dynamic linker will place in the two GOT
+slots associated with this variable:
+
+- A capability to the static TLS block resolver function at offset 0.
+- The offset of the variable in the static TLS block at offset 16 (8 bytes).
+- The size of the variable at offset 24 (8 bytes).
```
jrtc27, line 898: Isn't that an implementation detail of the runtime you're using, not part of the ABI? AFAIK the only requirement is that the first capability point to the resolver?

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -588,6 +592,19 @@ Relocations needed to define the traditional TLS models are undefined.
     |       |                                  |                            | descriptor function for ``S + A``.                        |
     +-------+----------------------------------+----------------------------+-----------------------------------------------------------+
 
+The General Dynamic access sequence must be output in the following form to
+allow correct linker relaxation:
```
jrtc27, line 596: What's the LE sequence?

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+
+  adrp c0, :tlsdesc:sym
+  ldr c1, [c0, :tlsdesc_lo12:sym]
+  add c0, c0, :tlsdesc_lo12:sym
+  .tlsdesccall sym
+  blr c1
+  nop
+
+The compiler will emit a bounds setting instruction on c0 using the size
+returned in x1.
```
jrtc27, line 927: How does that work with weak undef symbols? Are you relying on the fact that SCBNDS doesn't trap in Morello? If so that's really ugly, and another reason other than "the runtime should set bounds in the resolver not rely on the caller being correct" (otherwise you can just manually call a resolver from assembly and get back an overly-permissive capability) why this is not a good idea.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+and its size in x1. The resolver function has a custom calling convention that
+must preserve all registers except c0 and c1.
+
+Considerations:
+
+- Any dynamically loaded modules will be placed outside of the bounds of the
+  thread pointer, so a resolver function cannot return an offset from the
+  thread pointer, but rather needs to return a pointer (capability).
+- To minimize reading of ``CTPIDR_EL0``, the resolver functions take a
+  copy of ``CTPIDR_EL0`` as an argument and preserve it.
```
jrtc27, line 884: So, you differ from AArch64 regardless of which way you do this. For the static resolver AArch64 doesn't use tpidr_el0 at all, and for the dynamic resolvers the value needs to be subtracted. From a code size perspective it's better to make the resolvers read it, and that's also a cleaner interface.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+^^^^^^^^^^^^^^
+
+The thread pointer is a capability, held in ``CTPIDR_EL0``. The thread pointer
+needs to have the read, write, read capability and write capability permissions
+and bounds such that the entire TLS static block is accessible.
+
+Resolver functions
+^^^^^^^^^^^^^^^^^^
+
+A resolver function takes arguments in c0 (address of the TLS GOT slot), and c2
```
jrtc27, line 873: This is an extremely weird calling convention. I can see _why_ you've done it (so that ctpidr_el0 can be read into c2 once and preserved), but I do not think that is a worthwhile optimisation given how little TLS is used when it makes such a confusing calling convention.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  blr c1
+  nop
+
+The compiler will emit a bounds setting instruction on c0 using the size
+returned in x1.
+
+General Dynamic to Local Exec relaxation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The relaxed sequence will use pairs of movz/movk instructions to materialize
```
jrtc27, line 932: I don't think that really matters?

aaelf64-morello/aaelf64-morello.rst
jrtc27: This is also repeated below; I don't think there should be more than one copy

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -827,3 +844,91 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+the size and offset. The offset to the thread pointer (c2) and the result is
+returned in c0, while the size is returned in x1.
```
jrtc27, line 934: This sentence doesn't make any sense to me

aaelf64-morello/aaelf64-morello.rst
sbaranga-arm: Undef weak symbols should be handled using a special resolver that writes 0 to x0 and x1 (the text needs to updated).
This is indeed specific to Morello (scbnds won't trap).
If we would set the bounds in the resolver I guess the caller could still get the module base symbol if they wanted to get an over-permissive capability. I agree that setting the bounds in the resolver would reduce the permissions of the capabilities that are in registers (for the general case), so this needs further consideration.

aaelf64-morello/aaelf64-morello.rst
sbaranga-arm: I don't think this has too much impact on code size (ctpidr_el0 would typically be read once in a function, so the code size impact shouldn't be different from AArch64), but moving it to the resolver might have an impact on the dynamic instruction count. I think how clean the interface is here is subjective (from my perspective having to read ctpidr_el0 in resolvers is less clean).

aaelf64-morello/aaelf64-morello.rst
sbaranga-arm: The other reason is that if we would read CTPIDR in the resolver then we would also need to read it when doing the GD->LE relaxation (in the LE sequence), and that would add another NOP to the GD sequence.

jrtc27: Now that x1 is gone as a return value you can instead pass CTPIDR in c1 at least and avoid the hole

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -687,6 +703,9 @@ Dynamic Morello relocations
     | 59397 | ``R_MORELLO_TLSDESC``   | ``TLSDESC(S+A)``                        | Identifies a TLS descriptor to be filled.|
     |       |                         |                                         |                                          |
     +-------+-------------------------+-----------------------------------------+------------------------------------------+
+    | 59398 | ``R_MORELLO_DTPREL128`` | ``DTPREL(S)``                           |                                          |
```
ruben-arm, line 706: Would this be `R_MORELLO_TLS_TPREL128`?

aaelf64-morello/aaelf64-morello.rst
sbaranga-arm: We still need a scratch register though. That could be c2, but then that leaves a gap in the preserved registers, which is less nice. I would rather leave this like it is (also avoids some large changes in both the spec and implementations).

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -865,3 +929,191 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  adrp c0, :tlsdesc:sym
+  ldr c1, [c0, :tlsdesc_lo12:sym]
+  add c0, c0, :tlsdesc_lo12:sym
+  nop
+  .tlsdesccall sym
+  blr c1
+
+The General Dynamic access sequence must be output in the following form to
+allow correct linker relaxation:
+
```
mmalcomson, line 1076: Nit:  we say here "in the following form", but the form is above the paragraph ;-)

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -865,3 +929,195 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  scbnds c0, c0, x1
+  ret c30
+
+Local Exec
+^^^^^^^^^^
+
+The capability to the TLS variable is derived from ``CTPIDR_EL0``. There
+are no requirements on how this is performed or the registers used, except
+that the sequence doesn't produce a dynamic relocation. A possible
+instruction sequence could be:
```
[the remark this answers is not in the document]
jrtc27: Because that would be a text relocation?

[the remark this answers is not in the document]
mmalcomson: Heh -- of course ;-)  (thanks)

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -865,3 +929,195 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+General Dynamic to Initial Exec relaxation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The relaxed sequence is:
+
+.. code-block:: text
+
+  adrp c0, :gottprel:sym
+  add c0, [c0, :gottprel_lo12:sym]
+  ldp x0, x1, [c0]
```
mmalcomson, line 1090: I think typo having the square-braces in the `add` instruction syntax.
Similar in the GD -> LE relaxed sequence below.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -550,6 +564,30 @@ Static Morello relocations
...
+  | 57356      | R\_MORELLO\_MOVW\_SIZE\_G1\_NC | SIZE       | Set a MOV[KZ] immediate field to bits [31:16] of X. No overflow check               |
+  +------------+--------------------------------+------------+-------------------------------------------------------------------------------------+
+  | 57357      | R\_MORELLO\_MOVW\_SIZE\_G2     | SIZE       | Set a MOV[KZ] immediate field to bits [47:32] of X; check that 0 <= X < 2\ :sup:`48`|
+  +------------+--------------------------------+------------+-------------------------------------------------------------------------------------+
+  | 57358      | R\_MORELLO\_MOVW\_SIZE\_G2\_NC | SIZE       | Set a MOV[KZ] immediate field to bits [47:32] of X. No overflow check               |
+  +------------+--------------------------------+------------+-------------------------------------------------------------------------------------+
+  | 57359      | R\_MORELLO\_MOVW\_SIZE\_G3     | SIZE       | Set a MOV[KZ] immediate field to bits [63:48] of X (no overflow check needed)       |
+  +------------+--------------------------------+------------+-------------------------------------------------------------------------------------+
```
mmalcomson, line 587: I think it would be nice to either mention a restriction that these relocations are only for symbols directly (and disallowed on an offset), or to specify what we do with an offset in these `size` relocations as something like "the remainder".
(My preference is to disallow with an offset).

Stock-AArch64 allows offsets on TLS accessing relocations (like `add x0, x0, #:tprel_hi12:a+48`) -- at least in the BFD linker.

If implemented naively the `size` relocations just returning the size of a symbol would mean that the pointer we end up with after a Local-Exec sequence could access outside the object we wanted and not access the start of the object.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -865,3 +929,195 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  ``sym``, while sizeof(sym) denotes the size of the symbol ``sym``. These are
+  not valid assebler directives.
+
+The relaxation is performed by:
+
+- changing the ``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` relocation on the symbol
+  ``sym`` to a ``R_MORELLO_ADR_PREL_PG_HI20`` with the symbol ``_sym_data``
+
+- changing the ``R_MORELLO_TLSIE_ADD_LO12`` relocation on symbol the ``sym`` to
+  a ``R_AARCH64_ADD_ABS_LO12`` relocation with the symbol ``_sym_data``.
```
mmalcomson, line 1057: Is this requiring the addition of a symbol `_sym_data` by the linker, or is that symbol just there as an aid for explanation?
I expect the latter, but it would be nice to add a phrase to make it clearer.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -865,3 +929,195 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  mrs c1, CTPIDR_EL0
+  add c0, c1, x0, uxtx
+  scbnds c0, c0, x8
+
+Initial Exec to Local Exec relaxation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The linker will generate 16 bytes in the ``.rodata`` section, containing
+the offset in the static TLS block in the first 8 bytes and the size of the
+symbol in the next 8 bytes:
```
mmalcomson, line 1036: W.r.t. requiring that the linker has generated 16 bytes in the .rodata section.
As it corresponds to strange linker scripts which put .rodata input sections in a differently named output section:
Is this an input section name or output section name?

Also, is there a particular need to specify this?
I'd imagine any section marked read-only would work (though I can't say I know for certain).

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -610,21 +648,29 @@ Relocations needed to define the traditional TLS models are undefined.
...
+    | 57601 | ``R_MORELLO_TLSDESC_LD128_LO12``        | ``G(GTLSDESC(S+A))``       | Set the LD/ST immediate field to bits [11:4] of X.        |
+    |       |                                         |                            | No overflow check. Check that X&15 = 0.                   |
+    +-------+-----------------------------------------+----------------------------+-----------------------------------------------------------+
+    | 57602 | ``R_MORELLO_TLSDESC_CALL``              | None                       | For relaxation only. Must be used to identify a ``BLR``   |
+    |       |                                         |                            | instruction which performs an indirect call to the TLS    |
+    |       |                                         |                            | descriptor function for ``S + A``.                        |
+    +-------+-----------------------------------------+----------------------------+-----------------------------------------------------------+
+    | 57603 | ``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` | ``Page(G(GTPREL(S)))       | Set the immediate value of an ADRP to bits [31:12] of X.  |
```
acoplan-arm, line 666: Is there a reason that the TLSIE relocations don't allow for an addend on the symbol? Just asking for my own understanding. Thanks

jrtc27: Presumably because setting the size gets weird in that case; it needs to be done as an add after the scbnds

acoplan-arm: I see, that makes sense, thanks

aaelf64-morello/aaelf64-morello.rst
acoplan-arm, line 1057: Should this be `R_AARCH64_ADD_ABS_LO12_NC` instead of `R_AARCH64_ADD_ABS_LO12`? I can't see the latter in the aaelf64 doc.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -866,3 +934,202 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  ``sym``, while sizeof(sym) denotes the size of the symbol ``sym``. These are
+  not valid assebler directives.
+
+The relaxed sequence is:
+
+.. code-block:: text
+
+  adrp c0, _sym_data
+  add c0, [c0, :lo12:_sym_data]
+  ldp x0, x1, [c0]
```
mmalcomson, line 1133: `add` square braces typo.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -866,3 +934,202 @@ veneer is used. The BX changes the execution state from A64 to C64:
...
+  add c0, c0, :gottprel_lo12:sym
+  ldp x0, x1, [c0]
+  add c0, c2, x0
+  scbnds c0, c0, x1
+
+
+General Dynamic to Local Exec relaxation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The linker will generate 16 bytes in the ``.rodata`` section, containing
```
mmalcomson, line 1110: Just since we've adjusted the earlier phrase from "the `.rodata` section" to "a read-only section" it would be nice to apply the same change here.

PR #81, opened 2021-03-08
[DWARF32] Deleted duplicated TPIDRURO register number entry.

walkerkd:
A spurious entry for the TPIDRURO register was added to the
register number table in the document initially uploaded to github.


stuij (approved):
LGTM

PR #82, opened 2021-03-08
Codify the de-facto alignment rules when passing HAs on the stack

momchil-velikov:
Change-Id: If77d1cc01a4e14a0388ca82bd2c122d014095ee4


smithp35 (approved):
Commit message says HA rather than HFA, which would be nice to correct but not essential.

Other than that LGTM.

PR #84, opened 2021-03-12
[aapcs64] Use unicode and ≤ and ≥ instead of <= and >= [NFC]

smithp35:
The document is inconsistent about use of the mathematical char and the
C style <= and >=. Standardise on using mathematical chars when not in
source code.

No functional changes.


rearnsha (approved):
LGTM, we should probably check AAPCS32 as well, though.

smithp35:
aapcs32 uses ≤ and ≥ exclusively.

PR #85, opened 2021-03-19
[aapcs32] Clarify what does it mean for a VFP CPRC argument to be "correctly  aligned"

momchil-velikov:


smithp35 (approved):
Thanks for the update. LGTM. I think this captures the result of a recent discussion with reansha.

PR #88, opened 2021-04-01
Various typo fixes in aaelf64.rst

mmalcomson:
The description of the changes are below.
I believe there is only one modification that is not purely superficial (i.e. that can change how the document would be read).
That's the change of "until the relocation after has been resolved" to  "until after the relocation has been resolved".
I think that's a clearly correct change -- but calling it out as the main place for review.

Put code-quotes around S when used to represent the symbol address, and
A when used to represent an addend to a symbol.

Some spelling mistakes:
  requried  -> required
  presense  -> presence
  mechinasm -> mechanism

Remove a double-space character.

Change
  Addresses stored in the place of these relocations may not be used for
  pointer comparison until the relocation after has been resolved.
To
  Addresses stored in the place of these relocations may not be used for
  pointer comparison until after the relocation has been resolved.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -915,7 +915,7 @@ The following nomenclature is used in the descriptions of relocation operations:
 
 - ``G(expr)`` is the address of the GOT entry for the expression expr.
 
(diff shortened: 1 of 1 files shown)


stuij (approved):
LGTM

PR #90, opened 2021-04-06
Licensing

stuij:
Added license files to the tools directory. Added license info to the readme

### DIFF
--- a/tools/common/new-release.py
+++ b/tools/common/new-release.py
@@ -1,19 +1,19 @@
 #!/usr/bin/env python
 
-#  Copyright (c) 2021, Arm Limited
-#  SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2021, Arm Limited
+# SPDX-License-Identifier: Apache-2.0
 #
-#  Licensed under the Apache License, Version 2.0 (the "License");
-#  you may not use this file except in compliance with the License.
-#  You may obtain a copy of the License at
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#    http://www.apache.org/licenses/LICENSE-2.0
 #
-#  Unless required by applicable law or agreed to in writing, software
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


tools/common/new-release.py
```diff
@@ -0,0 +1,81 @@
...
+        # becomes:
+        # Copyright (c) 2003-2007, 2012, 2018, 2020-2021
+        (r"(opyright.* )(\d+)(?!.*\d)", lambda x: \
+                                                x.group(1) + year \
+                                            if x.group(2) == year \
+                                                else x.group(1) + x.group(2) + "-" + year),
+        # replace last year with this year, if it's part of a range:
+        # Copyright (c) 2003-2007, 2012, 2018-2021
+        # becomes:
+        # Copyright (c) 2003-2007, 2012, 2018-2021
```
danh-arm, line 45: minor: This is the same as line 43. Has this script already been run over this file!

tools/common/new-release.py
```diff
@@ -0,0 +1,81 @@
...
+
+def convert(text, quarter, year, date_of_issue):
+    re_pairs = [
+        (r"(\|release\| replace:: )(.*)", lambda x: x.group(1) + year + quarter),
+        (r"(\|date-of-issue\| replace:: )(.*)", lambda x: x.group(1) + date_of_issue),
+        # In copyright statements:
+        # append this year to last year with a hyphen if it's not a range:
+        # Copyright (c) 2003-2007, 2012, 2018, 2020-2021
+        # becomes:
+        # Copyright (c) 2003-2007, 2012, 2018, 2020-2021
```
danh-arm, line 37: minor: This is the same as line 35. Has this script already been run over this file!

tools/common/new-release.py
```diff
@@ -0,0 +1,81 @@
...
+
+import argparse
+import os
+import re
+
+description = (
+    "Replace dates in ABI files with dates suitable for a new release.")
+
+
+def convert(text, quarter, year, date_of_issue):
```
danh-arm, line 29: minor: In my experience, "touch all" scripts like this to update copyright years can be disruptive to downstream consumers if there are a lot of otherwise unchanged files. It's also not strictly necessary. That's probably not an issue for this project, though.

danh-arm (approved):

tools/common/new-release.py
stuij: ah, haha, yes, I think it has :)

tools/common/new-release.py
stuij: idem: I will fix them both.

tools/common/new-release.py
stuij: Yes, technically I would have preferred a central variable (of course), but the decision was made to have a strict separation between the ABI files in their own directory so they could be easier consumed individually. In this case I'd rather stick to not breaking that contract. I agree that your sketched problem is most probably not an issue here.

PR #91, opened 2021-04-09
PACBTI-M unwinding for DWARF

momchil-velikov:
Functions, compiled with Return Address Authentication will need to
keep a pointer authentication code, used to validate integrity of the
return address upon function exit. This patch defines a
pseudo-register `RA_AUTH_CODE`, which denotes the authentication
code. It is an unsigned integer with the same size as a
general-purpose register.

The intent is for producers to use that pseudo-register with the
ordinary CFI instructions DW_CFA_offset, DW_CFA_register, etc.

If the value of ``RA_AUTH_CODE`` is used to authenticate the return
address, the authentication shall use CFA as a modifier.

Change-Id: If285812db90eecc626c8c5acf43ac2d779c6a797


stuij (commented):
LGTM

PR #93, opened 2021-04-09
[aaelf64] Fix typos in TLS descriptor relocations' comments

pratlucas:
The comments for the `R_<CLS>_TLSDESC_LD_PREL19` and
`R_<CLS>_TLSDESC_ADR_PREL21` TLS relocations incorrectly mentioned a
range check that stated `-220 <= X < 2^20` instead of the expected
`-2^20 <= x < 2^20`. This patch fixes it.


aaelf64/aaelf64.rst
```diff
@@ -1452,35 +1452,35 @@ Thread-local storage descriptors
...
-  +------------+------------+---------------------------------+----------------------------------+----------------------------------------------------------------------------------------------+
+  +------------+------------+---------------------------------+----------------------------------+-------------------------------------------------------------------------------------------------------+
+  | ELF64 Code | ELF32 Code | Name                            | Operation                        | Comment                                                                                               |
+  +============+============+=================================+==================================+=======================================================================================================+
+  | 560        | 122        | R\_<CLS>\_TLSDESC\_ LD\_PREL19  | G(GTLSDESC(S+A)) - P             | Set a load-literal immediate to bits [20:2]; check -2\ :sup:`20` <= X < 2\ :sup:`20`; check X & 3 = 0.|
```
fpetrogalli, line 1458: Just a comment that you don't have to address if you disagree: as a personal preference I would have just increased of one line the height of the two lines being touched, avoiding to extend all the tables separators.

smithp35:
In terms of content, I can confirm that the 220 is a Typo, in the original 1.0 https://developer.arm.com/documentation/ihi0056/b?lang=en uses "-2^20 <= x < 2^20" it is likely that the mistake was introduced in the conversion to RST.

aaelf64/aaelf64.rst
pratlucas: Done. Looks better indeed.

mmalcomson (approved):
I've just double checked the PDF this generates looks fine at this area.
Everything seems OK to me.

PR #95, opened 2021-04-12
prepare documents for 2021q1 release

stuij:


philgee-oss:
aaelf32
-------
2021Q1 entry not appearing in the PDF Change history

addenda32
---------
Two Issue columns, other docs have one.

clibabi32
---------
No 2021Q1 entry in the Change history

cppabi32
---------
No 2021Q1 entry in the Change history

pauthabielf64
-------------
"Change history" title missing (other docs have it)

No 2021Q1 entry in the Change history

rtabi32
-------
No 2021Q1 entry in the Change history

semihosting
-----------
No 2021Q1 entry in the Change history

vfabia64
--------
No 2021Q1 entry in the Change history

stuij:
I've changed the addenda32 and pauthabielf64 to reflect your suggestions.

As for no 2021Q1 entry in the Change history: Just like last release we are publishing all relevant ABI docs for a release, regardless of if they have changes in them. This makes it easier on consumers of the documents to just get the latest, instead of finding what version might be the latest. I think adding an entry in all these documents with 'no change' will clutter up the document without supplying any value.

stuij:
aapcs64 space: no that's not necessary. The doc will render this correctly.


Change history entries: we have discussed either only putting changed documents in a new release like we've done on developer.arm.com or putting all documents in a new release. People thought it was a good idea to change the procedure.

I would prefer not amending the Change History table as part of these changes as it adds more work for us to amend all of the 20+ documents every time we do a new release. It's not totally trivial to do this in an automated way, and I don't necessarily see the value to add extra entries that don't add any information. Would it be ok to change our way of working for the ABI? If not, what is the value of the requirement?

Thanks,
/Ties

philgee-oss:
It is good practise to add an entry in the change history for the following reasons:

- Users would query why there is no entry in the Change History that corresponds to the version on the document, especially when the history seems to jumps a revision
- If there is a version change, then without an entry in the change history users would not know if there has been a technical change or not. So, adding a change entry that states "No technical changes." would prevent users from wasting time looking for any technical changes.

stuij:
Would a possible solution be a note in the readme, stating that if there is no entry in the change history for a revision, than the document hasn't changed for that revision. That, together with the knowledge that we publish all documents every release, might be enough? Or perhaps a note in just above the change table?

philgee-oss:
Having a note in the readme and above the table might be OK. You'll have to see if you get any user feedback on whether that is sufficient.

stuij:
Checking the changed files in this pull request again, I'm seeing all (or at least an approximation) of the documents now have a not above the table: https://github.com/ARM-software/abi-aa/pull/95/files

philgee-oss:
The problem is that all except aadwarf32 have the original note text you wrote:

`If there isn't an entry in the change history table for a release, there have been no changes made to the content of the document.`

Instead of my version:

`If there is no entry in the change history table for a release, there are no changes to the content of the document for that release.`

stuij:
Oh, I'm so sorry, that made me sad. That was exactly what I was trying to avoid :(

I should now have changed them all.

stuij:
The bpabi32 doc On github has as copyright date range 2005-2020. You're suggesting 2006-2008, 2012, 2018, 2021, and in the 2.10 version it says  2004-2005, 2007, 2008, 2012. Which one can we trust?

Looking at the release dates in the doc I guess it should be 2005, 2007, 2008, 2018, 2021. Agreed?

philgee-oss:
It should be 2005, 2007, 2008, 2012, 2018, 2021.

stuij:
yup, done :)

philgee-oss:
aadwarf32 ne3ds to be:

2003-2007, 2012, 2018, 2020, 2021

PR #96, opened 2021-04-13
Specify placement of `BTI` at exception handlers and after `setjmp` calls

momchil-velikov:


ehabi32/ehabi32.rst
```diff
@@ -1427,6 +1431,18 @@ The details are as follows:
...
 
+   If Branch Target Identification mechanism is enabled, the unwinder
+   may elect to transfer control to the routine installed in VRS[r15]
+   via a jump instruction that requires the jump target to be
+   identified as a valid destination for indirect jumps.
+
+   .. note::
+
+    The intent is that the routine installed in VRS[r15] should start with a
+    ``bti`` instruction (or equivalent), when it might operate in an environment
```
vhscampos, line 1442: IMHO your phrase reads better if the comma is removed.

vhscampos (changes requested):

ehabi32/ehabi32.rst
momchil-velikov: Agree, it's even grammatically incorrect to place a comma in front of a dependent clause.

clibabi32/clibabi32.rst
```diff
@@ -1581,6 +1585,11 @@ When \_AEABI\_PORTABILITY\_LEVEL != 0, the required definition of
 jmp\_buf cannot be used to create jmp\_buf objects. Instead, a jmp\_buf
 must be passed as a parameter or allocated dynamically.
 
+If Branch Target Identification mechanism is enabled, the
```
smithp35, line 1588: As an alternative suggestion using "BTI setting" and "BTI clearing" from the v8-M Arm ARM version B.o

If the Branch Target Identification mechanism is enabled, the longjmp may transfer control to the destination stored in the jmp\_buf, using a BTI setting instruction that requires a BTI clearing instruction at the destination.

ehabi32/ehabi32.rst
```diff
@@ -1427,6 +1431,18 @@ The details are as follows:
 
     In general an unwinder must load *all* the machine registers listed in the VRS.
 
+   If Branch Target Identification mechanism is enabled, the unwinder
```
smithp35, line 1434: A similar suggestion to the previous one:

If the Branch Target Identification mechanism is enabled, the unwinder may transfer control to the routine installed in VRS[r15] using a BTI setting instruction that requires a BTI clearing instruction at the destination.

smithp35 (commented):
I think the gist of the changes is that the ABI permits an implementation to use a BTI setting instruction in longjmp and the unwinder.

I've made a suggestion on wording to more closely match what is in the Arm ARM.

clibabi32/clibabi32.rst
momchil-velikov: I'm happy with such a wording, even though I personally interpret it pretty much as a requirement, rather than as a permission. Something like "Don't place BTIs on your own risk" :D

smithp35 (approved):
Thanks for the updates. LGTM.

PR #100, opened 2021-05-20
[AAELF64] Reserve relocation codes for PAuthABIELF64

smithp35:
The Alpha PAuthABIELF64 defines a small number of static and dynamic
relocations in the private vendor experiment space. When an upstream
implementation lands we will want to use non-experimental codes.
Step 1 is to reserve a range of static and dynamic relocations in
AAELF64 so we can guarantee no clash with any relocations defined in
the PAuthABI. When PAuthABIELF64 gets to Release state then
relocations can be added to the AAELF64 and the reservation can be
removed.


stuij (approved):
LGTM

PR #101, opened 2021-05-21
[aaelf64-morello] ELF markers to identify Morello purecap binaries

amilendra:
1) Assign e_flags bit 16 (0x00010000) to identify CHERI purecap binaries.
2) Add platform agnostic CHERI Note Section to specify abi variant.

### DIFF
--- a/aaelf64-morello/aaelf64-morello.rst
+++ b/aaelf64-morello/aaelf64-morello.rst
@@ -13,6 +13,8 @@
 .. _MORELLO_ARM: https://developer.arm.com/documentation/ddi0606/latest
 .. |tlsdesc-url| replace:: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
 .. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
+.. |cherielf-url| replace:: https://github.com/CTSRD-CHERI/cheri-elf-gabi/blob/main/gabi.md
+.. _CHERI_ELF: https://github.com/CTSRD-CHERI/cheri-elf-gabi/blob/main/gabi.md
 
 Morello extensions to ELF for the Arm\ :sup:`®` 64-bit Architecture (AArch64)
 *****************************************************************************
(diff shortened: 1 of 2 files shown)


aaelf64-morello/aaelf64-morello.rst
```diff
@@ -252,6 +256,37 @@ the same name in `AAELF64`_:
...
+  | (:code:`0x00010000`)            | implemented using capabilities (Pure-capability ABI)         |
+  +---------------------------------+--------------------------------------------------------------+
+
+Sections
+--------
+
+Special Sections
+^^^^^^^^^^^^^^^^
+
+A Morello toolchain can emit ELF Note sections in accordance to [CHERI_ELF_].
```
jrtc27, line 288: Should Morello prove successful, do you want to reserve the ability to reuse this flag to be its real non-prototype successor? As it stands this bit will forever be wasted once the Morello program comes to an end.

jrtc27: i.e. should this really be `EF_AARCH64_CHERI_PURECAP`?

bsdjhb: That is my only concern as well and the reason I had originally suggested an ELF note with a name of "Morello" as it could be easily jettisoned in the future.  Checking e_flags bits is of course vastly simpler to deal with than looking for an ELF note, so if the bit is potentially forward looking I'm happy with having a flag in e_flags.

amilendra: It is difficult to make predictions at this stage. We can either reuse or rename the flag for a future architecture.
However we don't have any strong opinions on the name so will go with the more generic `EF_AARCH64_CHERI_PURECAP`.

amilendra: @jrtc27, @bsdjhb
Are you happy with these changes?

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -252,6 +256,37 @@ the same name in `AAELF64`_:
...
+.. _Morello-specific e_flags:
+
+.. class:: aaelf64-morello-elf-flags
+
+.. table:: Morello-specific e_flags
+
+  +-----------------------------------+--------------------------------------------------------------+
+  | Value                             | Meaning                                                      |
+  +-----------------------------------+--------------------------------------------------------------+
+  | :code:`EF_AARCH64_CHERI_PURECAP`  | The ELF file contains uses an ABI where all pointers are     |
```
jrtc27, line 278: "contains uses"

jrtc27 (approved):

bsdjhb (approved):
Thanks.

PR #103, opened 2021-07-15
[Morello] Add the morello descriptor ABI specification.

sbaranga-arm:


descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+
+  ``R_MORELLO_DESC_FUNC_RELATIVE`` represents an optimization of ``R_MORELLO_DESC_JUMP_SLOT``.
+  It can be used when the symbol resolves to the current shared object or executable. ``S``
+  must be the ``Null`` symbol (Index 0). The address and permissions must be written
+  to the fragment. Creates a function capability.
+
+  ``R_MORELLO_DESC_IRELATIVE`` is used by the linker when transforming ``IFUNC`` s. The
+  rest are the same as ``R_MORELLO_DESC_RELATIVE``
+
+All of the above relocations are against the ``PT_MORELLO_DESC segment``. The fragment encoding
```
jrtc27, line 717:
```suggestion
All of the above relocations are against the ``PT_MORELLO_DESC`` segment. The fragment encoding
```

descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+    +------------+----------+----------------------------------------------------------------------------------------------------+
+
+
+Global entry points
+-------------------
+
+Each function has an additional entry point (the global entry point) that is used to manage
+the value of the Capability Private Data Addressing Register. The global entry point is a thunk
+generated by the linker and is used for indirect calls or calls from a different DSO.
+The global entry point is located in the same DSO as the associated function.
```
jrtc27, line 311: What section?

descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+    |                           |                                  |                                 |
+    +---------------------------+----------------------------------+---------------------------------+
+
+
+When creating function capabilities the runtime or dynamic linker will allocate space for two capabilities:
+
+- The first capability is the value of the Capability Private Data Addressing Register corresponding
+  to the DSO of the function. The address and base of this capability are the start of the
+  ``.desc.data.rel.ro`` section. This capability is read-only and has the bounds of the
+  ``.desc.data.rel.ro`` section.
```
jrtc27, line 754: And GOT, given 6.1?

descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+
+- ``.got``
+
+- ``.data``
+
+- ``.bss``
+
+
+The value of the Capability Private Data Addressing Register (c28) points to the start of the ``PT_MORELLO_DESC``
+segment, is read-only and can be used to access the ``.desc.data.rel.ro`` and ``.got`` sections.
```
jrtc27, line 424: Does it need to cover `.desc.data.rel.ro` if all descriptors are accessed indirectly via the GOT?

descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+the runtime of dynamic linker.
+
+PLT sequence
+------------
+
+The PLT sequence for the Morello Descriptor ABI is:
+
+.. code-block:: text
+
+    adrp c16, :got:foo
```
jrtc27, line 436: I think you mean adrdp? Otherwise this is PC-relative so code and data are 1:1 again.

jrtc27 (commented):
There's an undocumented implicit requirement that CCTLR.ADRDPB be 1 so that C28 is used, no?

How does all of this interact with TLS so you get per-compartment TLS rather than having it shared between copies of the same compartment?

jrtc27:
How do you handle function pointer equality in the presence of dlsym? What actually goes in `.desc.data.rel.ro`?

descabi-morello/descabi-morello.rst
```diff
@@ -0,0 +1,769 @@
...
+This specification provides the Morello Descriptor ABI specification for the
+Arm 64-bit Architecture (AArch64), and is expected to be used along with MORELLO_ELF_
+and MORELLO_PCS_.
+
+Morello Descriptor ABI
+======================
+
+The Morello Descriptor ABI is a variant of the pure capability ABI which decouples code and
+global data addresses, allowing the same code to be used with multiple instances of global data.
+This allows compartments to share code in the same address space, while having separata data
```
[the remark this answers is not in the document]
sbaranga-arm: Yes, there is an assumtion that CCTLR.ADRDPB should be set. I'll make that explicit.

Different compartments should have different CTPIDR_ELx/c28/global data so that should produce different values per compartment for the same TLS symbol.

The TLS details are no really covered in this though, I'll send an update for that (there should be a small deviation for purecap when the GD resolver needs to be called).

descabi-morello/descabi-morello.rst
sbaranga-arm: It ends up as adrdp when the linker resolves the relocation. It might be better to just say adrdp here though to avoid confusion.

sbaranga-arm:
At the moment it's the job of the dynamic linker to make sure only one pair of capabilities is allocated per function pointer while resolving the relocations, and that provides function pointer equality. There's likely scope for improvement here though.

descabi-morello/descabi-morello.rst
sbaranga-arm: Currently .desc.data.rel.ro contains capabilities that are needed for objects which don't have an associated GOT slot (this is inherited from purecap). We could use a different section for these capabilities and not have to cover .desc.data.rel.ro. This would minimize privilege, but I wonder if it adds much security-wise.

descabi-morello/descabi-morello.rst
sbaranga-arm: Not sure we need a specific section here, as long as it these end up in the pcc bounds.

smithp35:
A comment on the status of the patch. sbaranga-arm is no longer available to take this forward. If there is still interest in the descriptor ABI we'll need a volunteer to take this on, most likely raising another PR with the same text as I don't think there's a way of transferring ownership with GitHub.

jrtc27:
I don't know if anyone at Arm is invested in this any more to drive it forwards. Given that an implementation (probably of what's written here?) is present in Morello LLVM it might be worth reserving the relocations it's using, but I'm not aware of anyone actually using the implementation. Were we to look at using a descriptor ABI in our compartmentalisation work we would probably start from scratch, with this as a helpful reference, since from memory this specification didn't quite meet the needs that we'd have of such a thing. But I don't know if we actually plan to do so or not, it's just something that's come up in conversations in the past.

PR #105, opened 2021-07-21
[PAuthABI] Use a program property to mark PAuth compatibility

pbarrio:
A new program property GNU_PROPERTY_AARCH64_FEATURE_PAUTH in
section .note.gnu.property replaces the previous proposal of a
full section .note.AARCH64-PAUTH-ABI-tag. This is easier
to implement, at least in the Bionic dynamic linker: the property
(and the existing .note.gnu.property section) can be trivially
extracted from a DSO, and is already happening for other properties
(mainly GNU_PROPERTY_AARCH64_FEATURE_1_AND to check for BTI and
PAC-ret). In contrast, a new section would require traversing all
the sections in the ELF and checking for the section name, which is
more complicated to implement and possibly slower.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -14,7 +14,7 @@
 .. _LSB: https://refspecs.linuxfoundation.org/LSB_1.2.0/gLSB/noteabitag.html
 .. _SCO-ELF: http://www.sco.com/developers/gabi/
 .. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
-
+.. _LINUX_ABI: https://github.com/hjl-tools/linux-abi/wiki
 .. footer::
 
    ###Page###
@@ -253,6 +253,8 @@ This document refers to, or is referred to by, the following documents.
   | TLSDESC_                                                                                | http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf | TLS Descriptors for Arm. Original proposal document                      |
   +-----------------------------------------------------------------------------------------+-------------------------------------------------------------+--------------------------------------------------------------------------+
   | `GABI_SHT_RELR <https://groups.google.com/d/msg/generic-abi/bX460iggiKg/YT2RrjpMAwAJ>`_ | ELF GABI Google Groups                                      | Proposal for a new section type SHT_RELR                                 |
+  +-----------------------------------------------------------------------------------------+-------------------------------------------------------------+--------------------------------------------------------------------------+
+  | LINUX_ABI_                                                                              | https://github.com/hjl-tools/linux-abi/wiki                 | Linux Extensions to gABI                                                 |
   +-----------------------------------------------------------------------------------------+-------------------------------------------------------------+--------------------------------------------------------------------------+
 
 Terms and Abbreviations
@@ -832,10 +834,20 @@ language to signing schema is expected to evolve over time. Even if
 the low-level ELF extensions remain constant, a change to the
 high-level language mapping may result in incompatible ELF files.
 
-Every relocatable object, executable and shared library that uses the
-PAuth ABI ELF extensions must have a section named
-``.note.AARCH64-PAUTH-ABI-tag`` of type ``SHT_NOTE``. This section is
-structured as a note section as documented in SCO-ELF_.
+This document defines a default ELF marking schema and a base
+compatibility model. Platforms may define their own ELF marking and
+compatibility model that replace or extend the default ones. `Appendix
+Alternative ELF Marking Using GNU Program Properties`_ defines an
+alternative marking schema for platforms that support the
+``.note.gnu.property`` section.
+
+Default Marking Schema
+----------------------
+
+A new section named ``.note.AARCH64-PAUTH-ABI-tag`` of type
+``SHT_NOTE`` is defined. This section is structured as a note section
+as documented in SCO-ELF_, and its attribute flag ``SHF_ALLOC`` must
+be set.
 
 The name field (``namesz`` / ``name``) contains the string "ARM". The
 type field shall be 1, and the ``descsz`` field must be at least 16.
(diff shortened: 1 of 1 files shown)


pauthabielf64/pauthabielf64.rst
```diff
@@ -853,34 +853,36 @@ The (platform id, version number) of (0, 0) is reserved as an invalid
...
   pointers were signed in a compatible way with the default signing
   rules for tuple (platform id, version number).
 
-* The static linker may fault the combination of relocatable
-  objects that contain ``.note.AARCH64-PAUTH-ABI-tag`` sections with
-  incompatible (platform id, version number) tuples. If an ELF file is
+* The static linker may fault the combination of relocatable objects
+  that contain ``GNU_PROPERTY_AARCH64_FEATURE_PAUTH``
+  with incompatible (platform id, version number) tuples. If an ELF file is
   produced, the output ``.note.AARCH64-PAUTH-ABI-tag`` must have the
```
DanielKristofKiss, line 878: Here we also should say ``GNU_PROPERTY_AARCH64_FEATURE_PAUTH``.

DanielKristofKiss (changes requested):

pauthabielf64/pauthabielf64.rst
pbarrio: Ah, yes, thanks!

pauthabielf64/pauthabielf64.rst
```diff
@@ -832,19 +834,17 @@ language to signing schema is expected to evolve over time. Even if
...
+PAuth ABI ELF extensions must contain this property. The format of
+the data in ``pr_data`` is at least two 64-bit words, the first being
+a platform identifier, and the second being a version number specific
+to the platform identified in the first word. Consequently, the
+``pr_datasz`` field must be at least 16. When ``pr_datasz`` is larger
+than 16, the remainder of the contents of ``pr_data`` are specific
+to the (platform id, version number).
 
 This ABI does not determine the format of the platform identifier. Arm
 reserves the platform id 0 for a bare-metal platform.
```
pbarrio, line 850: @smithp35 would it be reasonable to specify some format, or at least add the values in a table so that platforms can add their identifiers here as they get allocated? I.e.
|| Platform || ID ||
| Baremetal | 0 |
| | <ID 1> |
| ... | ... |

Keeping the allocations clearly in one place might help platforms avoid stepping on each other's toes.

smithp35: I was expecting the majority of holders to use short null terminated strings such as "Arm\0" as there is 64-bits to play with. This would limit the possibility of clashes. However no objections to introducing a table. With the ABI in Github it should be easy to submit pull requests to update the table.

(On holiday today, so will be a bit slow responding)

pbarrio: Peter, thank you for the review. Feel free to respond after you are back from holidays :) 

How about a table like this:

|| Platform || ID (hex) || ID (human-readable) ||
| Baremetal | 0x0 | 0 |
| Platform 1 | 0x506c617431000000 | "Plat1" |

Then we can define in this ABI the rule by which two platforms are different. For example:

* The rule is that the full 64 bytes must be the same. "Arm\0\0\0\0\0" != "Arm\0Leg\0". I prefer this rule as it allows platforms to decide on the human-readable representation (string, integer, whatever) as long as the hex value is different to the rest of the platforms.

* The rule is that the field represents a null-terminated string. "Arm\0\0\0\0\0" and "Arm\0Leg\0" are the same vendor.

Defining the rule in this ABI would make it easier for tools to implement the comparison consistently for all platforms.

WDYT?

apazos:
Hi Pablo, your suggestion to use a new program property GNU_PROPERTY_AARCH64_FEATURE_PAUTH is good, and aligns with how BTI feature is handled too.

rearnsha:
We should **NOT** be putting .note.gnu.property features into the base ABI.  It's not our namespace.  What's more the 'gnu' term may cause problems with some partners.

rearnsha:
Use of GNU properties in the generic Arm ABI is not appropriate, IMO.  It's too specific to one platform.

pbarrio:
> Use of GNU properties in the generic Arm ABI is not appropriate, IMO. It's too specific to one platform.

@rearnsha, thank you for the comment. What is too platform-specific, 1) using GNU properties in the generic Arm ABI or 2) the specific property I am proposing?

We already use GNU properties in the generic Arm ABI, e.g. https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#special-sections. Do you see a difference between the new property and, say, GNU_PROPERTY_AARCH64_FEATURE_1_AND?

rearnsha:
We should not be using any GNU properties in the the platform-agnostic parts of the ABI.  If they're needed in platform-specific executables for Linux, that's OK.  But another solution needs to be found in object files.

smithp35:
Just come back from holiday. Will have a think about what the alternatives are. Off the top of my head:
* Use the raw SHT_NOTE section from the existing specification in relocatable objects but when targeting Linux/Android then write a GNU property section.
* Have a mirror of .note.GNU.property in the Arm namespace (not particularly keen on this).
* Some other mechanism, either a specific one for PAuthABI or a generic extensible one like BuildAttributes.

pauthabielf64/pauthabielf64.rst
```diff
@@ -858,6 +873,11 @@ related to the version number of this document. It is controlled by
 the object-producer based on the signing schema that have been used
 for pointers.
 
+If a file contains a section named ``.note.AARCH64-PAUTH-ABI-tag``,
+it must observe the entirety of the rules in this default marking
+schema. Generating such section with a platform-specific schema is
+forbidden.
+
```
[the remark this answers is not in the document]
pbarrio: I think I've resolved this but I've put the information a bit all over the place:

> A static linker may use the base compatibility model for relocatable objects, and a separate platform specific ELF marking for executables and shared libraries.
I think this is implicit from the other paragraphs, where we state that both the marking schema and the compatibility model can be platform-specific. Do we need further clarification in the bullet point?

I also split the 5th bullet point into sub-points to make it more readable.

pbarrio:
Thank you Peter, that was very useful! All changes made now. Please have a look at the second change you proposed: I've structured it a bit differently and you may have other ideas. The rest of the changes I've made almost verbatim.

smithp35 (approved):
Thanks for the update Pablo, I'm happy with the updated wording.

I think that this addresses the concerns of us exclusively using the GNU property in relocatable objects.

pbarrio:
Ok one more update after I received some questions elsewhere, which I'm posting here for visibility:

>  seem to remember there were some initial issues in the GNU property tags with alignment - has that been thought about and resolved - it would be sensible to get this right from the get-go.

Addressed with a clarification of alignment requirements when using GNU program properties and the use of `pr_padding`.

> the source of the specification for the tuple numbers seems a bit vague - How do I know which numbers to use?

Added a table with the platform IDs currently in use. Platforms are expected to start adding their choice of IDs to that table in order to prevent clashes.

I've also moved the baremetal platform from 0x0 to 0x1 in order to use 0x0 to represent an invalid platform. I have some uneasy feeling about using a valid platform ID (0x0 representing baremetal) in a tuple (0x0, 0x0) that means "invalid file". I also though (but haven't added to the patch) that, when using 0x0 as platform ID, we could use the second field of the tuple (version number) to contain a "reason for failure" so that consumers of the file could figure out why the file is invalid and provide good diagnostics at a minimum.

smithp35 (approved):
LGTM thanks for the update.

pbarrio:
Ok last change I hope.

* I've replaced the requirement to use (0,0) as an "invalid" combination by just using a platform id = 0. This was reserved as "Invalid" in my previous iteration of the patch yesterday. This also means that we could in theory use the "version id" field for other purposes when the platform id is 0 - if we ever want to do that.

* Slightly reworded the compatibility model to explicitly allow files without pointers to be marked with the section: "the presence of a ``.note.AARCH64-PAUTH-ABI-tag`` means that, *if the file contains pointers*, they were signed in a compatible way with the default signing...". This makes it easier for toolchains and loaders to reason about file compatibility, and for loaders to leave PAC on, instead of conservatively having to turn it off. It's still the platform's choice, though, to mark them or just trust files without any marking.

Let me know if you have any comments on these changes. I'll ask the ABI owner to merge the patch later next week.

Thanks everyone for all the reviews!

smithp35 (approved):
I'm happy with the changes. Will need to ask an editor nicely to merge the changes.

pbarrio:
@stuij, could you merge this pull request?

PR #106, opened 2021-08-17
Add relocation relaxation

Wilco1:
Add initial description of relocation relaxations. Make it clear that linkers may optimize existing relocations (or pairs of relocations) to improve code quality. A simple relaxation is replacing the common case of addition with zero with a move or nop. A more complex one is removing unnecessary GOT indirections from non pre-emptable symbols:

adrp x0, :got: symbol
ldr x0, [x0, :got_lo12: symbol]

and change into:

adrp x0, symbol
add x0, x0, :lo12: symbol

This works if these relocations are grouped in pairs and both instructions in each pair read/write the same register.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1249,6 +1249,88 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
 
 For "``MOVW``" type relocations it is the assembler’s responsibility to encode the hw bits (bits 21 and 22) to indicate the bits in the target value that the immediate field represents.
 
+Relocation optimization
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Linkers may optionally optimize instructions affected by relocation. Relocation optimizations improve the efficiency of relocated instructions without changing their visible behaviour. There are several classes of relocation optimizations:
+
+- A single relocation optimization may change an instruction after relocation into an equivalent, more efficient form.
+
+  - Several relocations may result in an addition with zero, which may be optimized as follows:
+
+    ::
+
+      ADD   x0, x1, 0   // eg. R_<CLS>_TLSLE_ADD_TPREL_HI12
+      ADD   x2, x2, 0   // or R_<CLS>_ADD_ABS_LO12_NC
+
+      // after optimization:
+
+      MOV   x0, x1
+      NOP
+
+  - The relocation ``R_<CLS>_ADR_PREL_PG_HI21`` may emit a MOV with zero immediate for undefined weak symbols.
+
+  - The following TLS relocations may be optimized if the symbol is not a pre-emptable definition and the TLS offset fits in 16 bits:
+
+    ::
+
+      ADRP  x0, :gottprel: symbol            // R_<CLS>_TLSIE_ADR_GOTTPREL_PAGE21
+      LDR   x1, [x0, :gottprel_lo12: symbol] // R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC
+      LDR   x2, :gottprel: symbol            // R_<CLS>_TLSIE_LD_GOTTPREL_PREL19
+
+      // after optimization:
+
+      NOP
+      MOV   x1, :tprel_g0: symbol            // R_<CLS>_TLSLE_MOVW_TPREL_G0
+      MOV   x2, :tprel_g0: symbol            // R_<CLS>_TLSLE_MOVW_TPREL_G0
+
+    If a linker supports optimizing ``R_<CLS>_TLSIE_ADR_GOTTPREL_PAGE21``, it must also support optimizing ``R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC``.
+
+- A sequence of relocated instructions may be optimized if all of the following conditions are true:
+
+  - The relocations apply to consecutive instructions in the order specified.
+  - The relocations use the same symbol.
+  - The relocated instructions have the same source and destination register.
+  - The relocations do not appear separately or in a different order.
+
+  In this case each set of relocations is independent and may be optimized. The following sequences are defined:
+
+  - Large GOT indirection
+
+    A GOT indirection may be optimized into PC-relative addressing:
+
+    ::
+
+      ADRP  x0, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
+      LDR   x0, [x0 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC
+
+      // after optimization:
+
+      ADRP  x0, symbol                  // R_<CLS>_ADR_PREL_PG_HI21
+      ADD   x0, x0, :lo12: symbol       // R_<CLS>_ADD_ABS_LO12_NC
+
+    This sequence may be optimized if it meets all of the following conditions:
+
+    - ``symbol`` is not a pre-emptable definition.
+    - ``symbol`` is not of type ``STT_GNU_IFUNC``.
+    - ``symbol`` is within range of the ``R_<CLS>_ADR_PREL_PG_HI21`` relocation.
+    - The addend of both relocations is zero.
+
+    The optimized sequence does not require a GOT entry. A linker may avoid creating a GOT entry if no other GOT relocations exist for the symbol.
+
+  - PC-relative addressing
+
+    ``ADR`` may replace ``ADRP/ADD`` if ``symbol`` is within +-1MiB range:
+
+    ::
+
+      ADRP  x0, symbol                  // R_<CLS>_ADR_PREL_PG_HI21
+      ADD   x0, x0, :lo12: symbol       // R_<CLS>_ADD_ABS_LO12_NC
+
+      // after optimization:
+
+      NOP
+      ADR   x0, symbol                  // R_<CLS>_ADR_PREL_LO21
 
 Proxy-generating relocations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^


MaskRay:
This may be called linker optimization where the total length does not decrease.  Linker relaxation means the total length can decrease (AVR/RISC-V), which is not the case here.

smithp35 (commented):
I think it may be better to call this GOT relaxation rather than linker relaxation. Although we are changing relocation codes we still write the same instructions, just to a different location.

It will be worth clarifying:
* Do consecutive pairs mean consecutive pairs of instructions or consecutive pairs of relocations?
* Does the static linker still have to create the GOT entry for symbol or can/must it skip the creation of the GOT entry?
* Although covered by pre-emptable as I don't think there is a pre-emptable reference, it may be worth saying there must be a non-pre-emptable definition of the destination.

I think X86 has a GOTRELAX relocation that does something similar, albeit explicitly. I think it did run into some teething problems https://lkml.org/lkml/2020/8/11/975 however I don't think that would apply to this relaxation as both sequences are position independent.

Wilco1:
I use the term relaxation when the linker changes an instruction opcode (rather than update an immediate field) - note your link also uses relaxation, so I think that is the correct term. Relaxations don't just apply to GOT relocations, I already mentioned relaxing ADD with zero to NOP, and ADRP+ADD could relax to ADR+NOP etc.

- It means both - maybe using "adjacent instructions" would be better?
- No GOT entry is required if all pairs can be relaxed and there are no other relocations that create a GOT entry. However whether a linker is able to avoid it is QoI.
- And you also have to deal with function symbols and PLTs which are tricky (eg. if the GOT entry points to a default PLT then you can still relax eventhough the destination symbol is not locally defined).

smithp35:
> And you also have to deal with function symbols and PLTs which are tricky (eg. if the GOT entry points to a default PLT then you can still relax eventhough the destination symbol is not locally defined).

That's an interesting case. If all input objects have been compiled with -fpic or -fpie then in there shouldn't need to be a canonical PLT and it is probably best to not to create one just to relax the instructions. If a canonical PLT is going to be created anyway, perhaps a program is a mix of -fpie and non-pie then it will be worth relaxing.

I'm not sure if there is anything that we need to write down here though. As this is effectively QoI as well.

Wilco1:
In principle compilers could always emit the address of a function using GOT relocations. That way you can avoid a PLT indirection in many cases. If you still need a canonical PLT, relaxation means there is no overhead from using GOT relocations.

aaelf64/aaelf64.rst
```diff
@@ -1249,6 +1249,10 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
 
 For "``MOVW``" type relocations it is the assembler’s responsibility to encode the hw bits (bits 21 and 22) to indicate the bits in the target value that the immediate field represents.
 
+Relocation relaxation
+^^^^^^^^^^^^^^^^^^^^^
+
+Linkers may perform relocation relaxations to optimize instructions affected by relocation. Relocations that result in addition with zero may emit a faster register move or nop instead. The relocations ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` may be relaxed if all such relocations for a symbol are consecutive pairs, and both relocations write the same register. In this case each pair is independent and can be safely relaxed. If the symbol is not pre-emptable, the addend is zero and the destination is within ADRP range, the relocations relax to ``R_<CLS>_ADR_PREL_PG_HI21`` and ``R_<CLS>_LDST64_ABS_LO12_NC``.
```
smithp35, line 1255: I think it would be useful to split this up and make some of the conditions more formal. I've made a suggestion below.

Linkers may optionally perform relocation relaxations to optimize instructions affected by relocation. This can take one of two forms:
* Relocations that operate on ADD (immediate) instructions where the result of the Operation is 0 may replace the ADD (immediate) instruction with a MOV (register) or NOP.
* A sequence of ordered instructions and relocations can be relaxed to an alternative sequence of ordered instructions and relocations when all of the following are true:
  * The instructions in the sequence are consecutive and in the correct order.
  * Each instruction has a relocation directive.
  * The relocation directives are in the same order as the instructions.
  * The conditions of the relaxation are met.
  
The ABI identifies the following relaxation sequences:

Relaxation of GOT-Relative to PC-Relative addressing mode.
Instructions and Relocations to be relaxed:
```
ADRP rd, :got: symbol                // R_<CLS>_ADR_GOT_PAGE 
LD      rd, [rd :got_lo12: symbol]   // R_<CLS>_LD64_GOT_LO12_NC
```
Relaxed instruction sequence:
```
ADRP rd, symbol                  // R_<CLS>_ADR_PREL_PG_HI21 
LD      rd, [rd :lo12: symbol]   // R_<CLS>_LD64_ABS_LO12_NC
```
Conditions for the relaxation:
* ``rd`` is the same register.
* ``symbol`` is a non-preemptible definition.
* The ``symbol`` does not have type ``STT_GNU_IFUNC``.
* The relocation addend is 0.
* The result of the relaxed relocations are within range.

smithp35: Other thoughts:
* I've suggested not doing the relaxation for ifuncs resolvers as this causes more problems (canonical PLT). This could be reduced to a recommendation or omitted as QoI but thought it simple enough to just say don't do it.
* (Not for this patch) The TLS relaxations should ideally be documented here, or the sysvabi64 (by reference).

One possible objection to applying the relaxation is the intent of the compiler/assembler. For example did they use a GOT generating relocation because they wanted to force the generation of a GOT entry, or did they just want to load the address of a variable via the most efficient means necessary. I can't currently think of a reason to require a GOT entry when PC-relative addressing would suffice but it may exist.

Another possibility that we could do instead of, or even in addition to is to introduce a new set of relocation that explicitly permit relaxation. These would obviously be harder to deploy as it would require a new linker to work at all, not just to be optimised. I had originally thought it would be possible to define these new relocations such that it would be possible to avoid the constraint that the ADRP and LD were consecutive, however if they weren't then there is a risk that the ADRP and LD are in different 4k pages and one being in range and the other not.

Wilco1: That's a good suggestion, I'll update the text. Note the pairs are necessary since both relocations must be relaxed in lockstep (you obviously can't relax one but not the other!). Having pairs also means the ADRP value of the first instruction cannot escape and be combined with a 2nd instruction elsewhere, ie. each pair is independent. Any new relocations would still need to define pairs - the only advantage of new relocations is that the linker could assume that all preconditions are met, which might save some checks.

aaelf64/aaelf64.rst
```diff
@@ -1249,6 +1249,10 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
 
 For "``MOVW``" type relocations it is the assembler’s responsibility to encode the hw bits (bits 21 and 22) to indicate the bits in the target value that the immediate field represents.
 
+Relocation relaxation
```
MaskRay, line 1252: It seems that the term "linker optimization" is often used when the number of bytes does not change while "linker relaxation" is used when the number of bytes decreases.

The two arches which support GOT optimization use the term optimization.

aaelf64/aaelf64.rst
MaskRay, line 1255: "pre-emptable" is uncommon. Either "preemptible" or "preemptable" is ok. "preemptible" as a word is more common (and in lld/ELF, my preference) while "preemptable" is used by the ELF spec (and in the LLVM IR "dso_preemptable").

MaskRay:
Two points. First, for a regular `adrp+ldr+ldr` code sequence loading the value of a variable through GOT indirection, either the first two instructions (adrp+ldr) can be optimized (compute the GOT address by PC-relative), or the three instructions can be optimized as a whole (load the variable directly via LDR (literal)).

Here is an aarch64 Mach-O example:
```
% cat a.s
.globl _main
.p2align 2
_main:
Lloh0:
  adrp x8, _C@GOTPAGE
Lloh1:
  ldr x8, [x8, _C@GOTPAGEOFF]
Lloh2:
  ldr w0, [x8]

  .loh AdrpLdrGotLdr Lloh0, Lloh1, Lloh2

.globl _C
.zerofill __DATA,__common,_C,4,2
.subsections_via_symbols
% clang a.s
% objdump -d a.out
...
0000000100003fa4 <_main>:
100003fa4: 1f 20 03 d5  nop
100003fa8: 1f 20 03 d5  nop
100003fac: a0 02 00 18  ldr     w0, 0x100004000 <_C>
100003fb0: c0 03 5f d6  ret
100003fb4: e9 03 09 aa  mov     x9, x9
```

Second. The Mach-O scheme allows unrelated instructions interleaved with optimized code sequences, e.g. the `mov x9, x9` below.

```asm
Lloh0:
  adrp x8, _C@GOTPAGE
  mov x9, x9     # unrelated
Lloh1:
  ldr x8, [x8, _C@GOTPAGEOFF]
Lloh2:
  ldr w0, [x8]

  .loh AdrpLdrGotLdr Lloh0, Lloh1, Lloh2
```

In PowerPC ELFv2, the following TOC (GOT) loads can be optimized simultaneously, even though the code sequences are interleaved.
```asm
addis 3, 2, .LC0@toc@ha
addis 4, 2, .LC1@toc@ha
ld 3, .LC0@toc@l(3)
ld 4, .LC1@toc@l(4)
lwa 3, 0(3)
lwz 5, 0(4)
```

aaelf64/aaelf64.rst
MaskRay: For the Mach-O directive `.loh`, it stands for Linker Optimization Hints :)

Wilco1:
> Two potential extensions to "consecutive pairs".
> 
> First, for a regular `adrp+ldr+ldr` code sequence loading the value of a variable through GOT indirection, either the first two instructions (adrp+ldr) can be optimized (compute the GOT address by PC-relative), or the three instructions can be optimized as a whole (load the variable directly via LDR (literal)).

That's interesting. My initial goal here is to enable optimization of existing relocations in a backwards compatible way, and removing the load of a GOT access is the most significant one. Doing anything more requires new relocations and/or special annotation sections like .loh. Do you have an idea what kind of gains remain after removing the GOT indirection? I'm not convinced it would be worth it.

> Second. The Mach-O scheme allows unrelated instructions interleaved with optimized code sequences, e.g. the `mov x9, x9` below.

Well there is no benefit in scheduling the ADRP and LDR of a GOT access apart - in GCC there are major performance gains just from keeping them together!

> Both Mach-O `.loh` directives and PPC relocation types encode more information and imply that optimization will be legitimate. Put it simply, the producer ensures correctness. The aarch64 ELF relocation types may not have that strong guarantee.

Yes, this means new relocations which are somehow linked together so they can be resolved as a group. I don't believe ELF has a good mechanism to support that in its relocations, so you might need something like .loh.

aaelf64/aaelf64.rst
```diff
@@ -1249,10 +1249,57 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
...
+      ADD x2, x2, 0   // or R_<CLS>_ADD_ABS_LO12_NC
+
+      // after optimization:
+
+      MOV x0, x1
+      NOP
+
+  - The relocation ``R_<CLS>_ADR_PREL_PG_HI21`` may emit a MOV with zero immediate for undefined weak symbols.
+
+- A sequence of relocations can be relaxed into more efficient relocations. All of the following conditions must be true:
```
smithp35, line 1273: "A sequence of relocations can be relaxed into more efficient relocations" reads a bit strangely, and we'll also need to be careful with ``R_<CLS>_ADD_ABS_LO12_NC`` as this will need to do more than just relocate the immediate field, it will also need to change the instruction. As we've not got a relocation code for rewrite this ``ld`` as an ``add`` then the static linker will most likely have to introduce an internal one or rewrite the instruction separately.

I suggest:
A sequence of instructions with relocations, under certain conditions, can be relaxed into more efficient instructions. For the relaxation all of the following conditions must be true:
...

smithp35 (commented):
Only one small comment. Rest looks good to me.

MaskRay:
> > Two potential extensions to "consecutive pairs".
> > First, for a regular `adrp+ldr+ldr` code sequence loading the value of a variable through GOT indirection, either the first two instructions (adrp+ldr) can be optimized (compute the GOT address by PC-relative), or the three instructions can be optimized as a whole (load the variable directly via LDR (literal)).
> 
> That's interesting. My initial goal here is to enable optimization of existing relocations in a backwards compatible way, and removing the load of a GOT access is the most significant one. Doing anything more requires new relocations and/or special annotation sections like .loh. Do you have an idea what kind of gains remain after removing the GOT indirection? I'm not convinced it would be worth it.

I don't know. It is worth measuring on larger projects like Clang/Chromium: (a) how much the current restricted scheme can optimize and (b) how much a flexible Mach-O like form can optimize,

> > Second. The Mach-O scheme allows unrelated instructions interleaved with optimized code sequences, e.g. the `mov x9, x9` below.
> 
> Well there is no benefit in scheduling the ADRP and LDR of a GOT access apart - in GCC there are major performance gains just from keeping them together!

OK:) So perhaps my following point may be moot:

I was thinking that if new relocation types will ever be introduced, specifying optimization for old relocation types can be a waste.
For GNU ld's x86-64 port (a register-memory arch), it optimizes R_X86_64_GOTPCREL for mov and R_X86_64_{REX_,}GOTPCRELX for more opcodes.
The former is mostly redundant nowadays.

> > Both Mach-O `.loh` directives and PPC relocation types encode more information and imply that optimization will be legitimate. Put it simply, the producer ensures correctness. The aarch64 ELF relocation types may not have that strong guarantee.
> 
> Yes, this means new relocations which are somehow linked together so they can be resolved as a group. I don't believe ELF has a good mechanism to support that in its relocations, so you might need something like .loh.

ld64's .loh testsuite is here: https://github.com/apple-opensource/ld64/tree/master/unit-tests/test-cases/linker-optimization-hints

If you have Apple ld64 contact, perhaps worth asking them the up/down sides for the scheme (any over-engineering?)

CC @alexander-shaposhnikov who is interested and more familiar with the .loh scheme.

aaelf64/aaelf64.rst
Wilco1: Indeed, I'm not suggesting that linkers literally change GOT relocations to PC-relative internally.

smithp35 (approved):
I'm happy with the proposal and don't have any more suggestions.

alexander-shaposhnikov:
I've done a quick experiment using LLVM/Clang and ported "linker optimization hints" (currently emitted for Apple targets) to ELF with the goal to collect some initial numbers e.g. the distribution of different kinds of hints, I think I will have them ~today/~tomorrow and can publish the numbers here. (If I understand correctly the current proposal covers some of them but not everything).

alexander-shaposhnikov:
I'll probably double check tomorrow, but it looks like enabling the pass AArch64CollectLOH (https://www.llvm.org/doxygen/AArch64CollectLOH_8cpp_source.html) for a Chromium build would yield
~ 2263779   AdrpAdd
~ 92542       AdrpLdrGot
~ 82145       AdrpLdrGotLdr
~24076        AdrpLdr
and the rest are very small.

Wilco1:
> I'll probably double check tomorrow, but it looks like enabling the pass AArch64CollectLOH (https://www.llvm.org/doxygen/AArch64CollectLOH_8cpp_source.html) for a Chromium build would yield
> ~ 2263779 AdrpAdd
> ~ 92542 AdrpLdrGot
> ~ 82145 AdrpLdrGotLdr
> ~24076 AdrpLdr
> and the rest are very small.

That must be a huge binary! How many ADRPs are there in total? It seems odd there are so many AdrpAdd (especially since LLVM doesn't enable anchors/base pointers like GCC). Since ADRP+ADD requires the binary to be less than 1MB in total, I expect none of those can be optimized. So the main benefit of LOH looks like the AdrpLdrGotLdr case.

alexander-shaposhnikov:
Those were measured on the compiler's side  (dumped all the lohs from all the object files in the build tree), 
I'm currently working on getting more accurate numbers (and how many of them can be really relaxed) from LLD's side, 
I'll try to post them here ~tomorrow.

alexander-shaposhnikov:
I've been looking into ADRP+ADDs, it turns out that approximately 43000 pairs can be optimized 
in the sense that 
    L1: adrp xA, sym@PAGE
    L2: add xB, xA, sym@PAGEOFF
can be transformed into 
    L1:  adr xB, sym
    L2 : nop
(where the correctness is ensured by the compiler, e.g. L2 is the only use of L1's xA.)
Unfortunately I was looking at the build without a sections ordering file, I would expect this number to be higher if the the layout is specified (or if the binary is smaller).
I'm interested in hearing your opinion if there is a way to add support for this.
For example, hypothetically, one could introduce a new type of relocation R_AARCH64_RELAXABLE_ADR_PREL_PG_HI21 that would be required to be paired with R_AARCH64_ADD_ABS_LO12_NC and could be optionally emitted by the compiler
in place of the usual R_AARCH64_ADR_PREL_PG_HI21 to communicate to the linker that it's safe to try to apply the aforementioned transformation.

smithp35:
> I've been looking into ADRP+ADDs, it turns out that approximately 43000 pairs can be optimized by the linker (once the final layout is known)
> in the sense that
> L1: adrp xA, sym@PAGE
> L2: add xB, xA, sym@PAGEOFF
> can be transformed into
> L1: adr xB, sym
> L2 : nop
> (where the correctness is ensured by the compiler, e.g. L2 is the only use of L1's xA.)
> Unfortunately I was looking at the build without a sections ordering file, I would expect this number to be higher if the the layout is specified (or if the binary is smaller).
> I'm interested in hearing your opinion if there is a way to add support for this.
> For example, hypothetically, one could introduce a new type of relocation R_AARCH64_RELAXABLE_ADR_PREL_PG_HI21 that would be required to be paired with R_AARCH64_ADD_ABS_LO12_NC and could be optionally emitted by the compiler
> in place of the usual R_AARCH64_ADR_PREL_PG_HI21 to communicate to the linker that it's safe to try to apply the aforementioned transformation.

For xA != xB I think that particular transformation would need another relocation or some other way (can't think of anything better right now) of communicating that xA is safe to not calculate as there's no later use of xA.

For xA == xB I think it would be possible for a linker to recognise without another relocation in a similar way to Wilco's existing relaxation proposal.

I'm guessing we're talking about the majority if not all the cases being the former here.

If we need a new relocation then it may be worth splitting it out into a separate pull request. It is also possible for small < 1 MiB code + data executables that something similar can be done with the adrp followed by load.

Wilco1:
> I've been looking into ADRP+ADDs, it turns out that approximately 43000 pairs can be optimized by the linker (once the final layout is known)

What is the total binary size and the number of ADRPs? Did you see any other cases being optimized?

> I'm interested in hearing your opinion if there is a way to add support for this.

I think this would need new relocations since R_AARCH64_ADR_PREL_PG_HI21 is used in many different ways. Small binaries would benefit the most.

Wilco1:
I did some experiments on SPEC2017 which has a mixture of binary sizes from small to large. 27% of all R_AARCH64_ADD_ABS_LO12_NC was in range of ADR. In 95% of cases it was immediately preceded by ADRP using the same destination register, so it could be changed to NOP (17% of all R_AARCH64_ADR_PREL_PG_HI21).

These are quite high percentages, so if it results in performance gains it is worth supporting.

alexander-shaposhnikov:
Hi! Many thanks! Sorry about the delay on my side, I got a little bit distracted by some other stuff.
The binary I'm looking at is indeed quite large (and as I mentioned above the linking step doesn't specify the order file).
43K is approximately ~4-5 % of the total number of ADRPs, I would expect this number to increase once the order file is specified. I took a look at distribution of the value of (L2 - L1), the case where (L2 - L1) == 4  (i.e. ADRP and ADD are adjacent) accounts for approximately 37%, (L2 - L1) == 8 - 48%, (L2 - L1) = 12 - 10% and the rest are very small. I'm a little bit surprised that on SPEC2017 ADD is preceded immediately by ADRP in 95% cases - were you using the latest version of Clang ? (I was using Clang built from the current master).
But overall even in this case it seems to be relatively significant so would be good not to leave these opportunities on the table.
Regarding the implementation details - yeah, I can start a separate pull request, I'm just a little bit worried that there is a lot of context here (on this discussion thread) that would be good not lose.
Maybe we can briefly summarize the potential plan while everyone is here. 

1/ The case where ADD is immediately preceded by ADRP and xA == xB seems to be the simplest one and doesn't require new types of relocations, it only needs to be documented.
2/ In the case where ADD is not immediately preceded by ADRP or xA != xB 
- it looks like we need a new type of relocation here that would be emitted in place of R_AARCH64_ADR_PREL_PG_HI21 and be paired with R_AARCH64_ADD_ABS_LO12_NC - is that correct or do you have some other schema on your mind?) (also maybe you have an idea about potential names for the new type ?)

Wilco1:
I used GCC with scheduling switched off so it won't schedule the ADRP and ADD apart. It is easy to teach the compiler to keep them together (this ensures the destination registers are the same and improves code quality). So we only need to consider consecutive pairs, and if we optimize to NOP+ADR (instead of ADR+NOP) then I believe it is safe and we don't need new relocations.

It looks like we can support almost all optimizations without introducing the LOH scheme or new relocations. If there are more useful cases we can consider new relocations for those. Note it is a good idea to improve other relocations at the same time - there are plenty opportunities to improve addressing.

alexander-shaposhnikov:
All right, thanks, are you planning to include a clause for this case into the current proposal or would you like me to send a new pull request ?

Wilco1:
I have a draft patch for ADR+ADD which I'll post soon. I'm currently looking into adding a few missing TLS relaxations. With those I believe we will have all beneficial relaxations covered that work on existing relocations.

pcc:
Does this proposal account for the case where the address is outside the code model? With HWASan for example, the addresses of globals may be tagged (bits 56-63 set), and therefore outside of the code model. We currently use the following instruction sequence to take the address of the global:
```
adrp x0, :pg_hi21_nc:global
movk x0, #:prel_g3:global+4294967296
add x0, x0, :lo12:global
```
However, prior to this proposal, the following sequence would work as well:
```
adrp x0, :got:global
ldr x0, [x0, :got_lo12:global]
```
With relaxation I think this would end up rendering the program unrelocatable.

I think there are two options to address this:

1. Require the linker to disable relaxation if the global address is outside the code model.
2. Implement this via an additional pair of relaxable relocations instead, like on x86.

Wilco1:
I'm not sure why you think this could fail. If an address is far away due to setting the top byte then it would be out of range and so relaxations would not apply. If you dynamically set the top byte in the GOT then the symbol would need to be marked preemptable (since its address is determined dynamically) and relaxations will not apply.

pcc:
Ah, I see that you've addressed this via this language in the specification:

> The symbol is within range of the ``ADRP``.

I don't think I have any remaining concerns, thanks.

aaelf64/aaelf64.rst
```diff
@@ -1249,10 +1249,57 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
...
+      LDR   x0, [x0 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC
+
+      // after optimization:
+
+      ADRP  x0, symbol                  // R_<CLS>_ADR_PREL_PG_HI21
+      ADD   x0, x0 :lo12: symbol        // R_<CLS>_ADD_ABS_LO12_NC
+
+    A sequence may be optimized if it meets all of the following conditions:
+
+    - ``symbol`` is a non-preemtable definition.
```
smithp35, line 1314: In the rest of the document we've used pre-empt rather than preempt. Would be good to keep that consistent. I guess non-pre-empable would work. Or we could reword to avoid the `non` with something like ``symbol`` is a definition that cannot be pre-empted.

aaelf64/aaelf64.rst
```diff
@@ -1249,10 +1249,57 @@ Finally the bit-field of X specified in the table (those bits of X picked out by
...
+    - ``symbol`` is not of type ``STT_GNU_IFUNC``.
+    - The addend of both relocations is zero.
+    - The symbol is within range of the ``ADRP``.
```
smithp35, line 1317: Possibly worth using the relocation information here for clarity?
- The result of ``R_<CLS>_ADR_PREL_PG_HI21`` to ``symbol`` passes the range check.

smithp35 (approved):
A couple of small wording suggestions, but otherwise LGTM.

smithp35 (approved):
LGTM

smithp35 (approved):
LGTM.

I guess ADRP/LDR could be replaced with LDR // R_<CLS>_GOT_LD_PREL19 ? Although not strictly necessary to put that in here.

aaelf64/aaelf64.rst
```diff
@@ -1260,16 +1260,32 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
...
 
       // after optimization:
 
-      MOV x0, x1
+      MOV   x0, x1
       NOP
 
   - The relocation ``R_<CLS>_ADR_PREL_PG_HI21`` may emit a MOV with zero immediate for undefined weak symbols.
 
+  - The following TLS relocations may be optimized if the symbol is a non-preemtable definition and the TLS offset fits in 16 bits:
```
smithp35, line 1273: Typo preemtable with no p. To be consistent with the rest of the document I think non-pre-emptable, or definition that cannot be pre-empted.

aaelf64/aaelf64.rst
```diff
@@ -1260,16 +1260,32 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
...
+      LDR   x1, [x0, :gottprel_lo12: symbol] // R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC
+      LDR   x2, :gottprel: symbol            // R_<CLS>_TLSIE_LD_GOTTPREL_PREL19
+
+      // after optimization:
+
+      NOP
+      MOV   x1, :tprel_g0: symbol            // R_<CLS>_TLSLE_MOVW_TPREL_G0
+      MOV   x2, :tprel_g0: symbol            // R_<CLS>_TLSLE_MOVW_TPREL_G0
+
+    Linkers must ensure that ``R_<CLS>_TLSIE_ADR_GOTTPREL_PAGE21`` is only optimized for a symbol if ``R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC`` is optimized too.
```
smithp35, line 1287: Could be worth emphasising the same ``symbol``. For example Linkers must ensure that ``R_<CLS>_TLSIE_ADR_GOTTPREL_PAGE21`` to ``symbol`` is only optimized if ``R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC`` to ``symbol`` is also optimized.

aaelf64/aaelf64.rst
```diff
@@ -1260,16 +1260,32 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
...
+
```
smithp35, line 1288: May be worth documenting in the sequence section the existing TLS IE to LE relaxation that BFD and LLD already implement.
ADRP x0, :gottprel: symbol            // R_<CLS>_TLSIE_ADR_GOTTPREL_PAGE21
LDR   x0, [x0, :gottprel_lo12: symbol] // R_<CLS>_TLSIE_LD64_GOTTPREL_LO12_NC
to
MOVZ x0, :tprel_g1: symbol
MOVK x0, :tprel_g0_nc: symbol

No strong opinions though.
Although there are way more possible ones which may better belong in the sysvabi.

Wilco1:
> I guess ADRP/LDR could be replaced with LDR // R__GOT_LD_PREL19 ? Although not strictly necessary to put that in here.

Yes, and you could also use ``R_<CLS>_ LD_PREL_LO19`` for general loads.

aaelf64/aaelf64.rst
Wilco1: Interestingly this means there is already code in linkers which checks for a consecutive pair of relocations using the same destination register. However that case is less general and less efficient than what I am proposing. And TLS sizes larger than 64KB should be exceedingly rare!

My goal is to document the most beneficial relaxations here and ensure all linkers support these by default. There are likely more relaxation opportunities, but there is no point in documenting/implementing relaxations on relocations that are rarely used.

aaelf64/aaelf64.rst
Wilco1: Yes, I'll make it clear it needs to fit ``R_<CLS>_ADR_PREL_PG_HI21``

aaelf64/aaelf64.rst
Wilco1: I'll fix

aaelf64/aaelf64.rst
Wilco1: I'll fix

aaelf64/aaelf64.rst
Wilco1: And it should be obvious it applies to all occurrences of these relocations. The statement is just to make it explicit there is a dependency rather than trying to describe the linker implementation in detail. I'll reword it slightly

smithp35 (approved):
Thanks for the update. I've no more comments. Happy to see this merged.

Wilco1:
@stuij Done, it's a single commit now.

stuij (approved):
Thanks!

PR #107, opened 2021-08-17
[sysvabi64] Medium code model

Wilco1:
The medium code model is an extension of the small model to allow more than 4GB of data. It is significantly more efficient than the large model and supports PIC and PIE. Symbols above a certain size use large sections which are placed after all other sections. Small objects are addressed like in the small model, large objects use a GOT indirection.

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -612,7 +612,7 @@ performance and reduce static code size. The relevant constraints are:
   typically used for function / procedure calls.
 
 * A series of MOVZ and up to 3 MOVK instructions can be used to
-  construct a 64-bit value
+  construct a 64-bit value.
 
 * Static linkers may insert a veneer (a sequence of instructions) to
   implement a relocated B or BL to a destination further away than the
@@ -636,24 +636,28 @@ along with the assumptions that the code model may make.
 
 .. table:: Code Models
 
-  +-------+----------------+----------------+------------------------+
-  | Code  | Max text       | Max combined   | Additional GOT         |
-  | Model | segment size   | span of text   | restrictions           |
-  |       |                | and data       |                        |
-  |       |                | segments       |                        |
-  +=======+================+================+========================+
-  | tiny  | 1 MiB          | 1 Mib          | none                   |
-  |       |                |                |                        |
-  +-------+----------------+----------------+------------------------+
-  | small | 2GiB           | 4 GiB          | pic: got size < 32 KiB |
-  |       |                |                +------------------------+
-  |       |                |                | PIC: none              |
-  +-------+----------------+----------------+------------------------+
-  | large | 2GiB           | no restriction | pic: got size < 32 KiB |
-  |       |                |                +------------------------+
-  |       |                |                | PIC: max distance from |
-  |       |                |                | text to GOT < 4 GiB    |
-  +-------+----------------+----------------+------------------------+
+  +--------+----------------+----------------+------------------------+
+  | Code   | Max text       | Max combined   | Additional GOT         |
+  | Model  | segment size   | span of text   | restrictions           |
+  |        |                | and data       |                        |
+  |        |                | segments       |                        |
+  +========+================+================+========================+
+  | tiny   | 1 MiB          | 1 Mib          | none                   |
+  |        |                |                |                        |
+  +--------+----------------+----------------+------------------------+
+  | small  | 2GiB           | 4 GiB          | pic: got size < 32 KiB |
+  |        |                |                +------------------------+
+  |        |                |                | PIC: none              |
+  +--------+----------------+----------------+------------------------+
+  | medium | 2GiB           | no restriction | pic: got size < 32 KiB |
+  |        |                |                +------------------------+
+  |        |                |                | PIC: none              |
+  +--------+----------------+----------------+------------------------+
+  | large  | 2GiB           | no restriction | max distance from text |
+  |        |                |                | to GOT < 4 GiB         |
+  |        |                |                |                        |
+  |        |                |                |                        |
+  +--------+----------------+----------------+------------------------+
 
 .. note::
 
@@ -706,8 +710,8 @@ Implementation of code models
 -----------------------------
 
 The convention for command-line option to select code model is
-``-mcmodel=<model>`` where code model is one of small, medium or
-large.
+``-mcmodel=<model>`` where code model is one of ``tiny``, ``small``,
+``medium`` or ``large``.
 
 The convention for command-line option to select position-independent
 code is ``-fpic`` for pic ``-fPIC`` for PIC. When compiling for an
@@ -720,22 +724,37 @@ open-source compilers.
 
 .. table:: Code Model Support
 
-  +-------+----------------+----------------+
-  | Code  | GCC 11         | Clang 12.0     |
-  | Model |                |                |
-  +=======+================+================+
-  | tiny  | yes            | yes            |
-  |       |                |                |
-  +-------+----------------+----------------+
-  | small | pic and PIC    | pic and PIC    |
-  |       | implemented    | accepted but   |
-  |       |                | PIC used       |
-  +-------+----------------+----------------+
-  | large | no-pic support | no-pic support |
-  |       |                |                |
-  |       |                |                |
-  |       |                |                |
-  +-------+----------------+----------------+
+  +--------+----------------+----------------+
+  | Code   | GCC 11         | Clang 12.0     |
+  | Model  |                |                |
+  +========+================+================+
+  | tiny   | yes            | yes            |
+  |        |                |                |
+  +--------+----------------+----------------+
+  | small  | pic and PIC    | pic and PIC    |
+  |        | implemented    | accepted but   |
+  |        |                | PIC used       |
+  +--------+----------------+----------------+
+  | medium | pic and PIC    | pic and PIC    |
+  |        | proposed       | proposed       |
+  |        |                |                |
+  +--------+----------------+----------------+
+  | large  | no-pic support | no-pic support |
+  |        |                |                |
+  |        |                |                |
+  |        |                |                |
+  +--------+----------------+----------------+
+
+
+Medium code model
+-----------------
+
+Symbols smaller than 64KiB (value TBD) are treated like the small code model.
+Symbols which are larger than 64KiB, are ``common`` or of unknown size use
+sections ``.ldata``, ``.lbss`` and ``.lrodata``. These sections are placed after
+all other sections and can be unlimited in size. Symbols in them are addressed
+via the GOT. Linkers may relax GOT indirections into PC-relative addressing for
+non pre-emptible symbols within 4GiB.
 
 Sample code sequences for code models
 -------------------------------------
@@ -759,12 +778,20 @@ PC-relative offset of +/- 1 MiB. Suitable for tiny code model.
 
     adr x0, var
 
-PC-relative offset of +/- 4 GiB. Suitable for small code model.
+PC-relative offset of +/- 4 GiB. Suitable for small/medium code model.
 
 .. code-block:: asm
 
     adrp x0, var
     add x0, x0, #:lo12: var
+
+PC-relative offset not within +/- 4 GiB. Suitable for large symbols in
+the medium code model.
+
+.. code-block:: asm
+
+    adrp x0, :got: var
+    ldr x0, [x0, :got_lo12: var]
 
 Absolute load of a literal, where the literal is within 1
 MiB. Suitable for large code model if the literal is defined within
@@ -802,7 +829,7 @@ In the general case one of the sequences above is used to get the
 address. A load instruction can then be used to obtain the value. In
 some cases the add can be folded into the load.
 
-PC-relative offset of +/- 4 GiB. Suitable for the small code model.
+PC-relative offset of +/- 4 GiB. Suitable for the small/medium code model.
 
 .. code-block:: asm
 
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -720,22 +724,37 @@ open-source compilers.
...
+
+Medium code model
+-----------------
+
+Symbols smaller than 64KiB (value TBD) are treated like the small code model.
+Symbols which are larger than 64KiB, are ``common`` or of unknown size use
+sections ``.ldata``, ``.lbss`` and ``.lrodata``. These sections are placed after
+all other sections and can be unlimited in size. Symbols in them are addressed
+via the GOT. Linkers may relax GOT indirections into PC-relative addressing for
+non pre-emptible symbols within 4GiB.
```
smithp35, line 758: If I've understood correctly we'll end up with an image layout (simplified) like:
.text
.rodata
.rwdata
.bss
.lrodata
.lrwdata
.lbss

This would mean at least a couple more page-aligned segments. It would also mean linker's needing modifications in generic places (linker scripting) to recognise the sections, we'd probably need a more distinctive name to avoid name clahes. By default they'd be placed as orphans matched to the nearest type (.lrodata after .rodata) etc.  

Is that worth it? For example could this be simplified to everything being accessed via the GOT, with the GOT relaxation to the PC relative addressing mode doing the heavy lifting. This might mean introducing a new relocation directive to make the relaxation easier to do in more cases. For example R\_\<CLS\>\_ADR_GOTPAGERELAX and R\_\<CLS\>\_LD64_GOTRELAX\_LO12\_NC which a linker only needs to check if there is a non-pre-emptable definition of the symbol before deciding to relax?

I can see a case when there are a few really gigantic sections which it would be advisable to sort later. In terms of the requirements, perhaps it would be better to just say that sections should have a prefix of .ldata, .lbss and .lrodata and let the programmer write a linker script if they felt like sorting these after. This would make this proposal easier to deploy. The prefix is important as something like -fdata-sections will name sections .data.<variable-name>.

smithp35 (commented):
I have some concerns on how easy this will be to deploy upstream due to it needing changes in architecture neutral parts of linkers. We may be able to get most of the benefit with a few small changes though.

sysvabi64/sysvabi64.rst
Wilco1: Yes, in the medium model you end up with an extra load segment if you do have some large arrays. The linker could automatically merge large sections with the standard ones if they fit, and place them at the end if they don't. An alternative may be to place small symbols in sbss/sdata and ensure both come before data/bss - but that might also add an extra load segment?

Note lbss/ldata/sbss/sdata are supported on multiple targets, it is mostly turning on the sections and implementing target hooks. Since it's not new functionality I'd expect the implementation to be fairly straightforward.

Without a way to sort symbols into small and large, we would need to assume all symbols are out of range. That means we have an overhead on every symbol access rather than only on a few huge arrays (since it applies to static variables as well, it's even worse than -fPIC!). On a prototype sorting into different sections reduces the performance overhead by 3x and codesize overhead by 10x.

Yes, adding new relaxing relocations is certainly possible. However it means changes to compilers, assemblers and linkers to support new relocations, and swap between new/old relocations. This creates backwards compatibility issues which you avoid when relaxing existing relocations as an optional optimization. Note this should ideally be discussed on my other pull request since relaxation is not required for or exclusive to the medium model.

I'm not sure I get your point about writing a linker script - the medium model is meant to do that automatically for you.

smithp35: Thanks for pointing out that existing uses of .ldata etc. Looking at AArch64 and GNU ld I can see that X86_64 has .ldata etc. in its default linker script, but AArch64 does not, presumably part of the build-time configuration. LLD while not having a built-in linker script does not treat .ldata or .lbss separately for any architecture. That just means that we'll need to have linker work to implement this, or require a linker script if using an older linker. I note that the script puts the .lbss adjacent to the .bss which makes it more likely only a single additional segment is needed.

aarch64:
```
  . = SEGMENT_START("ldata-segment", .);
  . = ALIGN(64 / 8);
  __end__ = .;
  _end = .; PROVIDE (end = .);
  . = DATA_SEGMENT_END (.);
```

x86_64:
```
  .bss            :
  {
   *(.dynbss)
   *(.bss .bss.* .gnu.linkonce.b.*)
   *(COMMON)
   /* Align here to ensure that the .bss section occupies space up to
      _end.  Align after .bss to ensure correct alignment even if the
      .bss section disappears because there are no input sections.
      FIXME: Why do we need it? When there is no .bss section, we don't
      pad the .data section.  */
   . = ALIGN(. != 0 ? 64 / 8 : 1);
  }
  .lbss   :
  {
    *(.dynlbss)
    *(.lbss .lbss.* .gnu.linkonce.lb.*)
    *(LARGE_COMMON)
  }
  . = ALIGN(64 / 8);
  . = SEGMENT_START("ldata-segment", .);
  .lrodata   ALIGN(CONSTANT (MAXPAGESIZE)) + (. & (CONSTANT (MAXPAGESIZE) - 1)) :
  {
    *(.lrodata .lrodata.* .gnu.linkonce.lr.*)
  }
  .ldata   ALIGN(CONSTANT (MAXPAGESIZE)) + (. & (CONSTANT (MAXPAGESIZE) - 1)) :
  {
    *(.ldata .ldata.* .gnu.linkonce.l.*)
    . = ALIGN(. != 0 ? 64 / 8 : 1);
  }
  . = ALIGN(64 / 8);
  _end = .; PROVIDE (end = .);
  . = DATA_SEGMENT_END (.);`
```

Some thoughts:
* Should .ldata, .lbss etc be a mandatory part of the medium code-model or an optimisation of it? For example could it be QoI? For example implementations may choose to use ...

I'll add a comment to your relaxation proposal as well.

Wilco1: If we set the threshold for .lrodata/.ldata large enough, you very rarely need them. Large arrays are typically uninitialized and thus use .lbss, so you might not need an extra segment in most cases.

The only potentially tricky part is handling of common symbols. Given common symbols are already placed at the end of .bss (as shown above), you could assume all common symbols are large and use a GOT indirection (since -fno-common is the default, only Fortan COMMON blocks will create common symbols).

I'm not sure how you could make .lbss optional since it results in binary incompatibility (eg. GCC and LLVM x86 medium models are incompatible).

smithp35: Just to check on binary compatibility? 

Do you mean that some programs may fail to link due to differing assumptions of placement. For example compiler 1 that uses GOT addressing everywhere but doesn't use .lbss when mixed with objects from compiler 2 that only uses GOT addressing for .lbss may end up with variables out of range due to large arrays from compiler 1 placed in .bss.

From looking at some descriptions of the x86_64 medium code model it does sound like it is only worth doing the medium code-model without .ldata and friends. If that is the case then fine; it is not worth trying to simplify it further.

sysvabi64/sysvabi64.rst
```diff
@@ -720,22 +724,37 @@ open-source compilers.
...
+  |        |                |                |
+  |        |                |                |
+  +--------+----------------+----------------+
+
+
+Medium code model
+-----------------
+
+Symbols smaller than 64KiB (value TBD) are treated like the small code model.
+Symbols which are larger than 64KiB, are ``common`` or of unknown size use
```
smithp35, line 753: I'd prefer Symbols with type ``STT_COMMON`` or of unknown size ... Although I'd expect most people reading the spec to know that, ``common`` might just be confused with its more common (pun intended) English language meaning of frequent.

sysvabi64/sysvabi64.rst
Wilco1: Indeed, these are incompatible ABIs so you can't mix and match them. This is why we have to clearly define a medium model for AArch64.

Which description do you mean? By using .ldata you only need a GOT indirection on large arrays, so almost all accesses are as efficient as the small code model (and since array addresses are typically lifted out of loops, the GOT indirection is fairly cheap). Without .ldata every global access would need a GOT indirection, so you have a similar overhead as PIC with -fPIE or -fno-pic. To optimize this you might need new relocations to avoid GOT indirections on every global access.

smithp35: The description I was referring to was https://eli.thegreenplace.net/2012/01/03/understanding-the-x64-code-models which has a section on Medium code model, at least the GCC variant of it.

smithp35: FWIW I'm happy with the proposal. I'll need to read through carefully to see if there are any suggestions for wording/claifications.

My understanding is that this code-model should do the majority of what the large code model does, but in a more efficient and position independent way.

Wilco1: Note the medium model allows rodata of unlimited size which is not possible in the large model (text+rodata must stay <2GB due to eh_frame relocations - unless you swap eh_frame and ro_data placement if ro_data is large). So the new medium model effectively obsoletes the large model - there are no circumstances where you would prefer using the existing large model!

sysvabi64/sysvabi64.rst
```diff
@@ -720,22 +724,37 @@ open-source compilers.
...
+  |        |                |                |
+  +--------+----------------+----------------+
+  | large  | no-pic support | no-pic support |
+  |        |                |                |
+  |        |                |                |
+  |        |                |                |
+  +--------+----------------+----------------+
+
+
+Medium code model
```
smithp35, line 749: Here's an alternative wording suggestion.

The medium code model is suitable for programs that do not have large amounts of executable code, but do contain large static data objects such as large look up tables. The medium code model separates data into ``small data`` and ``large data``. The ``small data`` is addressed in the same way as the small code model, with the same maximum size restrictions as the small code model. The ``large data`` is addressed via the GOT and if placed after the ``small data`` does not have a maximum size limit.

Data is defined as ``large data`` if it is not Relocation Read-Only (RELRO) and any of the following are true:
* The data is >= 64 KiB in size.
* The data is of unknown size.
* The data is represented as a symbol with type ``STT_COMMON``.

All other data is ``small data``.

``large data`` are placed in sections with a prefix of ``l``, for example ``.lbss``, ``.ldata`` and ``.lrodata``. At static link time the ``large data`` sections must be placed after all ``small data`` sections. There is no large data equivalent of the ``.data.rel.ro`` section as all Relocation Read-Only data is ``small data``.

sysvabi64/sysvabi64.rst
MaskRay: `SHN_COMMON` is more common (pun intended).

In LLVM's integrated assembler, `.common` never uses `STT_COMMON`. binutils defaults to `--elf-stt-common=no` for most configurations and uses `SHN_COMMON` most of the times.

sysvabi64/sysvabi64.rst
Wilco1: > There is no large data equivalent of the `.data.rel.ro` section as all Relocation Read-Only data is `small data`.

Is there a reason behind this? It's easy to create a large rel.ro symbol, and for the medium model to work correctly, it has to be treated like large data and placed at the end.

smithp35: There's no fundamental reason, although IIUC the only reason to use data.rel.ro is if data contains pointers that need static initialisation. How many large objects, or arrays of objects contains that kind of data.

Wilco1: It should be very rare, but it seems best to avoid odd limitations that make it less generic.

sysvabi64/sysvabi64.rst
```diff
@@ -749,12 +749,27 @@ open-source compilers.
...
+model separates data into small data and large data. Small data is addressed
+in the same way as the small code model, with the same maximum size
+restrictions as the small code model. Large data is addressed via the GOT and
+does not have a maximum size limit.
+
+Data is defined as large data if any of the following are true:
+
+- The data is >= 64 KiB in size.
+- The data is of unknown size.
+- The data has type ``SHN_COMMON``.
```
smithp35, line 764: Being picky the symbol type is `STT_COMMON`` but the ELF spec refers to ``SHN_COMMON`` Symbols (``STT_COMMON`` symbols in objects have ``st_index`` = ``SHN_COMMON``. I take your point that people not familiar with ELF might not know what that means. I've made a suggestion below:

The data will be represented by an unallocated common block, described by a ``SHN_COMMON`` symbol.

sysvabi64/sysvabi64.rst
MaskRay: glibc/FreeBSD/musl's ld.so implementations only support one PT_GNU_RELRO.

LLD therefore has a diagnostic `<section_name> is not contiguous with other relro sections`

sysvabi64/sysvabi64.rst
```diff
@@ -749,12 +749,27 @@ open-source compilers.
...
+
+All other data is small data.
+
+Large data is placed in sections with a ``l`` prefix, for example ``.lbss``,
```
smithp35, line 768: I've made a few suggestions making the distinction between sections in objects and at link time. I've included a suggestion for ``.data.rel.ro`` (thanks MaskRay) I had forgotten that ``ld.so`` etc. only support one RELRO segment.  

The object producer places large data in sections with an additional ``l`` prefix, for example ``.lbss``, ``.ldata``, and ``.lrodata``. Both small and large RELRO data are placed in ``.data.rel.ro`` as many loaders only support one ``PT_GNU_RELRO`` program header.

At static link time the large data sections must be placed after all small data sections. The static linker allocates storage for SHN_COMMON symbols at the end of the ``.bss`` section if it is the last small data section, otherwise it is allocated in ``.lbss``.

smithp35 (commented):
Looking close. Just a few small suggestions.

sysvabi64/sysvabi64.rst
Wilco1: That's clearer indeed.

sysvabi64/sysvabi64.rst
Wilco1: It's a shame there can only be one RELRO segment, however limiting .data.rel.ro to 4GB doesn't seem like a big restriction. The distinction between object producers/consumers looks good indeed.

smithp35 (approved):
Thanks for the update. I'm happy with the text.

As the document is Alpha and this model is marked as Proposed I think we're good to go. Will need an editor to merge.

PR #108, opened 2021-08-18
aaelf64: Fix typo in description of PLT entry code generation

acoplan-arm:
This fixes what appears to be a typo in the "Overview of PLT entry code generation" section (currently 4.1.2.2).


smithp35 (approved):
FWIW I agree it looks like a typo. One of the areas I'd like to add to the sysvabi64 is some sample PLT sequences (we can be a bit more platform specific in that doc) that will hopefully illustrate at least an example of those constraints.

stuij (approved):

PR #112, opened 2021-09-10
[Morello] Rework varargs for Morello

sbaranga-arm:
For AAPCS64 (hybrid), if an argument is a capability or contains
capabilities, pass it indirectly. This avoids having an extra register
spill area. This also allows a capability-aware caller to pass
capabilities to a capability-unaware callee that would only construct a
va_list and pass the result to some other capability-aware code.

For AAPCS64-cap (pure capability ABI), this is reworked to add
extra protection. All variadic arguments are passed in memory
in 16-byte slots (if the type pased is larger than 16 or has
larger aligment requirements, pass it indirectly). The capability
to the variadic arguments is passed in C11, and this capability
will be NULL if there are no variadic arguments.

This protects from the following issues:
 - Caller reads more arguments than were passed
 - Some type mismatches

Note: this causes mismatches between AAPCS64 and AAPCS64-cap.

### DIFF
--- a/aapcs64-morello/aapcs64-morello.rst
+++ b/aapcs64-morello/aapcs64-morello.rst
@@ -376,7 +376,9 @@ Additionally, a stack-pointer register, SP in a 64-bit context or CSP in a capab
     +------------+----------+----------------------------------------------------------------------------------------------------+
     | r16        | CIP0     | The first intra-procedure-call scratch register (can be used by call veneers and PLT code).        |
     +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r9-r15     |          | Temporary registers.                                                                               |
+    | r10-r15    |          | Temporary registers.                                                                               |
+    +------------+----------+----------------------------------------------------------------------------------------------------+
+    | r9         |          | Parameter register for variadic calls, temporary register otherwise.                               |
     +------------+----------+----------------------------------------------------------------------------------------------------+
     | r8         |          | The capability indirect result location register.                                                  |
     +------------+----------+----------------------------------------------------------------------------------------------------+
@@ -415,6 +417,8 @@ Additionally, a stack-pointer register, SP in a 64-bit context or CSP in a capab
 
 
 The first eight registers, r0-r7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).
+
+In AAPCS64-cap the r9 register is used to pass anonymous arguments in variadic calls.
 
 Registers r16 (IP0/CIP0) and r17 (IP1/CIP1) may be used by a linker as a scratch register between a routine and any subroutine it calls (for details, see `Use of CIP0 and CIP1 by the linker`_). They can also be used within a routine to hold intermediate values between subroutine calls.
 
@@ -519,6 +523,16 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
     |                              |                                                                                          |
     | A.3                          |                                                                                          |
     +------------------------------+------------------------------------------------------------------------------------------+
+    |                              | If the callee is variadic in AAPCS64-cap, the Anonymous Arguments Memory area is         |
+    |                              | allocated in memory with a size of at least 16 * (number of Anonymous arguments) bytes.  |
+    |                              | The Anonymous Arguments Memory area is 16-byte aligned. The Anonymous Arguments          |
+    |                              | Capability points to the Anonymous Arguments Memory area and is copied to C9. If there   |
+    | A.4                          | are no Anonymous arguments passed, C9 is set to NULL.                                    |
+    +------------------------------+------------------------------------------------------------------------------------------+
+    |                              | The Anonymous Arguments Index (AArgsIdx) is set to zero.                                 |
+    |                              |                                                                                          |
+    | A.5                          |                                                                                          |
+    +------------------------------+------------------------------------------------------------------------------------------+
 
 
 .. rubric:: Stage B – Pre-padding and extension of arguments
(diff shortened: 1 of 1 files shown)


aapcs64-morello/aapcs64-morello.rst
```diff
@@ -810,78 +816,9 @@ The algorithm to implement the generic ``va_arg(ap,type)`` macro is then most ea
...
-    on_stack:
-        intptr_t arg = ap.__stack;
-        if (alignof(type) > 8)
-            arg = (arg + 15) & -16;
-        ap.__stack = (void *)((arg + sizeof(type) + 7) & -8);
-    #ifdef BIG_ENDIAN
-        if (classof(type) != "aggregate" && sizeof(type) < 8)
-            arg += 8 - sizeof(type);
-    #endif
+        intptr_t arg = (intptr_t)ap.__stack;
```
jrtc27, line 819: I don't see why this can't be `void *`? No arithmetic is done on it and it's always a valid pointer.

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -810,78 +816,9 @@ The algorithm to implement the generic ``va_arg(ap,type)`` macro is then most ea
...
+        if (alignof(type) > 16 || sizeof(type) > 16)
```
jrtc27, line 820: Hm. In theory you could pass larger values inline and it's still memory-safe, but I don't know how common it is and thus if the indirection really matters, plus this does provide slightly more type "safety"

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -753,53 +782,30 @@ The definition of ``va_list`` has implications for the internal implementation i
...
 
 .. code-block:: c
 
     typedef struct  va_list {
-        void * stack; // next stack param
-        void * gr_top; // end of GP arg reg save area
-        void * vr_top; // end of FP/SIMD arg reg save area
-        int gr_offs; // offset from  gr_top to next GP register arg
-        int vr_offs; // offset from  vr_top to next FP/SIMD register arg
+        void * stack; // next param
```
jrtc27, line 799: Or just make va_list itself be a `void *`? That's what other architectures do that don't have complex varargs tracking (e.g. RISC-V, for better or worse).

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -634,25 +664,28 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
...
     +-------------------------------+----------------------------------------------------------------------------------------+
     |                               | The argument is copied to memory at the adjusted NSAA.  The NSAA is incremented by the |
     |                               | size of the argument. The argument has now been allocated.                             |
-    | C.16                          |                                                                                        |
+    | C.18                          |                                                                                        |
     +-------------------------------+----------------------------------------------------------------------------------------+
 
+.. note::
+
+    In AAPCS64-cap if the callee is variadic and there are less than 4094 Anonymous arguments, the length of C11 divided by 16 is equal to the number of Anonymous arguments. The length of C11 divided by 16 is always greater or equal to the number of Anonymous arguments.
```
jrtc27, line 688: FWIW C only requires you support up to a total of 127 arguments for function calls

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -517,6 +517,16 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
     |                              |                                                                                          |
     | A.3                          |                                                                                          |
     +------------------------------+------------------------------------------------------------------------------------------+
+    |                              | If the callee is variadic in AAPCS64-cap, the Anonymous Arguments Memory area is         |
+    |                              | allocated in memory with a size of at least 16 * (number of Anonymous arguments) bytes.  |
+    |                              | The Anonymous Arguments Memory area is 16-byte aligned. The Anonymous Arguments          |
+    |                              | Capability points to the Anonymous Arguments Memory area and is copied to C11. If there  |
+    | A.4                          | are no Anonymous arguments passed, C11 is set to NULL.                                   |
```
jrtc27, line 524: Out of interest, why C11?

Also, this specifies that it must be nulled out on every non-variadic function call, right? That shouldn't add too much overhead because you can assume it's null inside a non-variadic function and thus only need to null it out if you allocate it yourself, but a function call could clobber it with unknown data so if you call `foo()` then `bar()` you can't guarantee that it's null any more for the second call. I think if you want this to be efficient you either need to make it only apply to variadic functions, or require that C11 also be null on return, that way it just becomes "if you allocate it you must clear it" and so can make it a higher-cost register to allocate.

(Note that prototype mismatches are already generally uncaught for CHERI, so I'm not sure if the specific case of variadic vs non-variadic needs special catching as it's just that problem; the wrong number/type of variadic arguments is however different because even if your prototypes match you can't check them without knowledge of the callee's semantics)

jrtc27:
Doesn't using a temporary also break lazily-bound functions, as the resolver won't save/restore it?

(STO_AARCH64_VARIANT_PCS exists to "solve" that but I doubt you want all varargs functions to use it)

aapcs64-morello/aapcs64-morello.rst
sbaranga-arm: The reasoning was that if the caller reads the wrong type once, reading subsequent argument with the correct type would still give you the correct value back (since we're only advancing by 16 bytes at a time). Plus, we can get the number of passed arguments this way.

jrtc27: The number of passed arguments is only relevant for the Linux syscall ABI, which is designed to avoid indirect arguments anyway, so I don't think that's ever relevant?

sbaranga-arm:
I thought the resolver can only clobber c16/c17, I need to double check.

jrtc27:
> I thought the resolver can only clobber c16/c17, I need to double check.

That's true for STO_AARCH64_VARIANT_PCS (but in practice lazy binding just isn't done because the cost of saving/restoring is way too expensive, and eats too much stack). Were that true for normal calls then STO_AARCH64_VARIANT_PCS would serve no purpose. Neither glibc nor FreeBSD's libc save x11, for example.

aapcs64-morello/aapcs64-morello.rst
sbaranga-arm: C11 is the first register that's not used for anything else (c9 and c10 are used for cheri_ccall).

It's not nulled out on non-variadic function calls at the moment, but might be a good idea.

sbaranga-arm:
> > I thought the resolver can only clobber c16/c17, I need to double check.
> 
> That's true for STO_AARCH64_VARIANT_PCS (but in practice lazy binding just isn't done because the cost of saving/restoring is way too expensive, and eats too much stack). Were that true for normal calls then STO_AARCH64_VARIANT_PCS would serve no purpose. Neither glibc nor FreeBSD's libc save x11, for example.

I see, this needs more thought then. Additionally saving/restoring C11 would probably work, but there might be other solutions to this as well.

aapcs64-morello/aapcs64-morello.rst
jrtc27: FYI cheri_ccall(ee) are going to be removed because they're not good interfaces and the only user, the MIPS-specific libcheri, is already gone from CheriBSD. I wouldn't worry about them, just use C9. We never added support for it to CHERI-RISC-V.

Interestingly glibc does save/restore x9 to get a whole number of register pairs, which seems like a waste, but maybe there's a good reason for that. FreeBSD is even weirder, it saves xzr and "restores" it...

I wouldn't bother with it, as I say that's really just a special case of prototypes not matching the definition. I had missed that this whole step was guarded by "If the callee is variadic".

jrtc27:
§6.1.1 still needs updating

aapcs64-morello/aapcs64-morello.rst
kevin-brodsky-arm: I think this comment still applies in the latest version.

kevin-brodsky-arm (approved):

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -634,25 +668,28 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
...
     +-------------------------------+----------------------------------------------------------------------------------------+
     |                               | The argument is copied to memory at the adjusted NSAA.  The NSAA is incremented by the |
     |                               | size of the argument. The argument has now been allocated.                             |
-    | C.16                          |                                                                                        |
+    | C.18                          |                                                                                        |
     +-------------------------------+----------------------------------------------------------------------------------------+
 
+.. note::
+
+    In AAPCS64-cap if the callee is variadic and there are less than 4094 Anonymous arguments, the length of C9 divided by 16 is equal to the number of Anonymous arguments. The length of C9 divided by 16 is always greater or equal to the number of Anonymous arguments.
```
acoplan-arm, line 692: Nitpick, but  s/less/fewer/ for the number of arguments.

sbaranga-arm: This is now removed. The syscall interface shouldn't be relying on this in the first place, and instead it should be getting the number of arguments by doing (length-offset)/16.

kevin-brodsky-arm: OK that sounds like a good approach, but I feel that having a note is still useful, otherwise there's no guarantee that the kernel can rely on the bounds being exact.

kevin-brodsky-arm: Discussed offline, this is not actually needed - even if there is zero padding at the end, loading additional zeroes is harmless.

sbaranga-arm: Re-added this after some more offline discussions. I also did some experiments, and we always get exact bounds for fewer than 4096 arguments (also adding more arguments means that we don't) .

kevin-brodsky-arm (approved):

PR #113, opened 2021-10-05
[sysvabi64] Incorporate tech-author review comments

smithp35:
Changes:
* Clarify that document is Alpha not Beta
* Wording changes for TBD sections.
* Wording change for installation section.

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -40,8 +40,8 @@ Date of Issue: |date-of-issue|
 
    PageBreak oneColumn
 
-Preamble
-========
+Preface
+=======
 
 Abstract
 --------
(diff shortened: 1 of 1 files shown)


sallyarmneale (approved):
LGTM

sysvabi64/sysvabi64.rst
```diff
@@ -40,8 +40,8 @@ Date of Issue: |date-of-issue|
 
    PageBreak oneColumn
 
-Preamble
-========
+Preface
```
stuij, line 43: I personally would like to be a bit conservative with changing the template of how we dress these documents. We have over 20 of them, and to maintain conformity, if we change one, we need to change them all.

I'm not a native English speaker, but I've looked both terms up in a couple of online dictionaries, and it looks to me like both of them could be used. Is there a strong reason to choose the one over the other?

preamble:
> A short preliminary statement or remark, especially an explanatory introduction to a formal document or statute.

preface:
>  The beginning or introductory portion that comes before the main text of a document or book.

smithp35: They are synonymous. I think preface is the current Arm standard term. For example the Arm Compiler Documentation https://developer.arm.com/documentation/101754/0616/Preface 
Same with the Arm ARM.

I expect the ABI documents were written before this. So I guess it is question of whether we stick with the existing ABI or whether  we follow the rest of Arm documents.

PR #120, opened 2021-11-22
aapcs64-morello: Fix up rule C.8

acoplan-arm:
Prior to this change, the wording of rule C.8 didn't allow passing a capability
in C7 in the case where the NGRN is 7 (i.e. when C0 through C6 are already
allocated). This change relaxes the rule, instead of requiring that the size of
the argument be strictly less than 16 * (8 - NGRN), we only require it to be
less than or equal to this value, which I believe gives the intended behaviour.
We also take the opportunity to avoid using the "quad words" terminology which
can be considered ambiguous.


sbaranga-arm (approved):
LGTM

sallyarmneale:
Looks good to me

PR #121, opened 2021-12-03
Soft-deprecate GNU_PROPERTY_AARCH64_FEATURE_1_PAC

pbarrio:
Users should not rely on this property to enquire about the status
of the protection of an ELF file. If it is present, then yes, the
file has been protected. If it is not present, it can still have
been protected to some extent.

The comment on Custom PLT protection has also been removed. This
protection is seen as a separate feature to PAC-ret, since you could
ask the linker to generate protected PLTs regardless of the presence
of PAC-ret.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1700,10 +1700,11 @@ sections are compatible with Branch Target Identification mechanism. An
 executable or shared object with this bit set is required to generate
 `Custom PLTs`_ with BTI instruction.
 
-``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all executable
-sections have Return Address Signing enabled. An executable or shared object
-with this bit set can generate `Custom PLTs`_ with a PAC
-instruction.
+``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all
+executable sections have been protected with Return Address Signing.
+Its use is optional, meaning that an ELF file where this feature bit
+is unset can still have Return Address signing enabled in some or all
+its executable sections.
 
 Program Loading
 ---------------


pbarrio:
@smithp35 @DanielKristofKiss  FYI clarifying the meaning of the PAC program property.

aaelf64/aaelf64.rst
```diff
@@ -1700,10 +1700,12 @@ sections are compatible with Branch Target Identification mechanism. An
 executable or shared object with this bit set is required to generate
 `Custom PLTs`_ with BTI instruction.
 
-``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all executable
-sections have Return Address Signing enabled. An executable or shared object
-with this bit set can generate `Custom PLTs`_ with a PAC
-instruction.
+``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all
+executable sections have been protected with Return Address Signing.
+Its use is optional, meaning that an ELF file where this flag is unset
```
smithp35, line 1705: I think the flags are called "feature bits", abbreviated to bit or bits in the BTI description. Would be good to keep the new text consistent.

DanielKristofKiss (approved):
LGTM, thanks!

aaelf64/aaelf64.rst
```diff
@@ -1700,10 +1700,12 @@ sections are compatible with Branch Target Identification mechanism. An
...
+can still have Return Address signing enabled in some of or all its
+executable sections. If used, this flag simplifies the task of
```
smithp35, line 1707: For the final sentence starting "If used". We normally don't add justification to the ABI documents, the argument normally goes that if you do it somewhere then you've got to do it everywhere.

I suggest having a think about whether it is necessary for the sentence to exist in terms of a reference specification? If it isn't then I suggest removing it. To be clear, I've not personally got a strong opinion that it must be removed, if at least one person wants to keep it, I'm happy to keep it.

smithp35 (commented):
Thanks for the update. A couple of small suggestions.

aaelf64/aaelf64.rst
pbarrio: Thanks, I don't have a strong opinion either, happy to go either way. @DanielKristofKiss what do you think?

Diverging a bit from the topic, the Arm ARM does have a mechanism to distinguish between what is a rule and what is an explanation. I find it useful to understand the intent of the spec without restricting implementations. Might be worth considering something similar for the ABI (FYI @stuij). It would also make it easier for external partners to contribute to the spec. I don't think this imposes a requirement on us to add explanations everywhere, though. Maybe something to discuss offline.

DanielKristofKiss: > We normally don't add justification to the ABI documents, the argument normally goes that if you do it somewhere then you've got to do it everywhere.
Normally we don't change the ABI in this way. I think it is important to describe the purpose of the flag. ( otherwise it could be just deprecated and dropped in this case).

In the ACLE spec the design docs hold additional information/motives/rationals besides the spec. (https://github.com/ARM-software/acle/tree/next-release/main/design_documents)

smithp35: There's certainly a case for supporting documents to contain additional information. In this particular case I don't want to hold up the change as it is only a sentence.

pbarrio:
@smithp35 I updated a new patch with your two suggestions.

smithp35 (approved):
LGTM, thanks for the update.

DanielKristofKiss (approved):
Thanks!

aaelf64/aaelf64.rst
```diff
@@ -1700,10 +1700,11 @@ sections are compatible with Branch Target Identification mechanism. An
...
 
-``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all executable
-sections have Return Address Signing enabled. An executable or shared object
-with this bit set can generate `Custom PLTs`_ with a PAC
-instruction.
+``GNU_PROPERTY_AARCH64_FEATURE_1_PAC`` This indicates that all
+executable sections have been protected with Return Address Signing.
+Its use is optional, meaning that an ELF file where this feature bit
+is unset can still have Return Address signing enabled in some or all
+its executable sections.
```
sallyarmneale, line 1707: of its executable sections

sallyarmneale (commented):
One minor change

PR #122, opened 2021-12-16
Use “thread” rather than “process”

rsandifo-arm:
The AAPCS64 uses “process” as a cover-all term for “thread
or process”:

  The AAPCS64 applies to a single thread of execution or
  process (hereafter referred to as a process).

However, SME adds a register called TPIDR2_EL0 that forms an
important part of the SME PCS.  The “T” in TPIDR2_EL0 stands
for “thread”, so it seemed incongruous to use “process” when
describing its role in the PCS.

This preliminary patch therefore tries to avoid treating
“thread” and “process” as synonymous.  Since the exact
nature of threads and processes depends on the platform,
the patch tries to give a flexible definition.

Two later parts of the PCS refer specifically to processes:

- “Universal stack constraints”, when describing the parts of stack
  memory that can be safely accessed.  These restrictions apply to
  the program generally (wrt a given thread's stack and SP), so the
  patch uses “conforming program” instead of “thread”.  This is also
  the term used in:

    A conforming program must only execute instructions that are in
    areas of memory designated to contain code.

- The section on interworking.  These restrictions specifically apply
  at the process level rather than just the thread level, so the patch
  leaves the wording unchanged.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -926,7 +926,7 @@ Memory and the Stack
 Memory addresses
 ^^^^^^^^^^^^^^^^
 
-The address space may consist of one or more disjoint regions. Regions
+The address space consists of one or more disjoint regions. Regions
 must not span address zero (although one region may start at zero).
 
 The use of tagged addressing is platform specific and does not apply to
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -898,10 +898,32 @@ registers, or if it is a function that returns results in such registers,
 it must ensure that p4-p15 are preserved across the call. In other cases
 it need not preserve any scalable predicate register contents.
 
-Processes, Memory and the Stack
--------------------------------
+Threads and processes
+---------------------
 
-The AAPCS64 applies to a single thread of execution or process (hereafter referred to as a process). A process has a program state defined by the underlying machine registers and the contents of the memory it can access. The memory a process can access, without causing a run-time fault, may vary during the execution of the process.
+The AAPCS64 applies to a single _`thread` of execution.  Each thread is in
+turn part of a _`process`.  A process might contain one thread or several
+threads.
+
+The exact definition of “thread” and “process” depends on the platform.
+If the platform is a traditional multi-threaded operating system,
+“thread” and “process” generally have their usual meaning for that
+operating system. If the platform supports multiple processes but has
+no separate concept of threads, “thread” is synonymous with “process”.
+If a platform has no concurrency or preemption then it will generally
+have a single “thread” and “process” that executes all instructions.
+
+Each thread has its own register state, defined by the contents of the
+underlying machine registers. A process has a program state defined by
+its threads' register states and by the contents of the memory that the
+process can access. The memory that a process can access, without causing
+a run-time fault, may vary during the execution of its threads.
+
+Memory and the Stack
+--------------------
+
+Categories of memory
+^^^^^^^^^^^^^^^^^^^^
 
 The memory of a process can normally be classified into five categories:
 
@@ -913,14 +935,24 @@ The memory of a process can normally be classified into five categories:
 
 - The heap.
 
-- The stack.
+- Stacks, with one stack for each thread.
 
-Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data. Except for the stack there is no requirement for each class of memory to occupy a single contiguous region of memory. A process must always have some code and a stack, but need not have any of the other categories of memory.
+Each category of memory can contain multiple individual regions.
+These individual regions do not need to be contiguous and regions of one
+memory class can be interspersed with regions of another memory class.
+
+Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data.
 
 The heap is an area (or areas) of memory that are managed by the process itself (for example, with the C malloc function). It is typically used for the creation of dynamic data objects.
 
-A conforming program must only execute instructions that are in areas of memory designated to contain code.
+Each individual stack must occupy a single, contiguous region of memory.
+However, as noted above, multiple stacks do not need to be organized
+contiguously.
 
+A process must always have access to code and stacks, but need not have
+access to any of the other categories of memory.
+
+A conforming program must only execute instructions that are in areas of memory designated to contain code.
 
 Memory addresses
 ^^^^^^^^^^^^^^^^
@@ -928,7 +960,6 @@ Memory addresses
 The address space may consist of one or more disjoint regions. No region may span address zero (although one region may start at zero).
 
 The use of tagged addressing is platform specific and does not apply to 32-bit pointers. When tagged addressing is disabled all 64 bits of an address are passed to the translation system. When tagged addressing is enabled, the top eight bits of an address are ignored for the purposes of address translation. See also `Pointers`_, above.
-
 
 The Stack
 ^^^^^^^^^
@@ -948,7 +979,7 @@ At all times the following basic constraints must hold:
 
 - Stack-limit ≤ SP ≤ stack-base. The stack pointer must lie within the extent of the stack.
 
-- A process may only access (for reading or writing) the closed interval of the entire stack delimited by [SP, stack-base – 1].
+- A conforming program may only access (for reading or writing) the closed interval of the entire stack delimited by [SP, stack-base – 1].
 
 - If MTE is enabled, then the tag stored in the stack pointer must match the tag set on the range SP - Stack-limit (i.e. the unallocated portion of the stack).
 
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -929,7 +929,7 @@ The address space may consist of one or more disjoint regions. No region
 may span address zero (although one region may start at zero).
 
 The use of tagged addressing is platform specific and does not apply to
-32-bit pointers. When tagged addressing is disabled all 64 bits of an
+32-bit pointers. When tagged addressing is disabled, all 64 bits of an
 address are passed to the translation system. When tagged addressing is
 enabled, the top eight bits of an address are ignored for the purposes
 of address translation. See also `Pointers`_, above.
@@ -953,16 +953,16 @@ Each category of memory can contain multiple individual regions.
 These individual regions do not need to be contiguous and regions of one
 memory class can be interspersed with regions of another memory class.
 
-Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data.
+Writable static data may be further sub-divided into initialized, zero-initialized, and uninitialized data.
 
-The heap is an area (or areas) of memory that are managed by the process itself (for example, with the C malloc function). It is typically used for the creation of dynamic data objects.
+The heap is an area (or areas) of memory that the process manages itself (for example, with the C malloc function). It is typically to create dynamic data objects.
 
 Each individual stack must occupy a single, contiguous region of memory.
 However, as noted above, multiple stacks do not need to be organized
 contiguously.
 
-A process must always have access to code and stacks, but need not have
-access to any of the other categories of memory.
+A process must always have access to code and stacks, but does not need
+to have access to any of the other categories of memory.
 
 A conforming program must only execute instructions that are in areas of memory designated to contain code.
 
@@ -982,12 +982,12 @@ The stack is defined in terms of three values:
 
 * the special-purpose register SP
 
-The SP moves from the base to the limit as the stack grows and from the
+The SP moves from the base to the limit as the stack grows, and from the
 limit to the base as the stack shrinks. In practice, an application might
 not be able to determine the value of either the base or the limit.
 
 In the description below, the base, limit and SP for a thread T are
-denoted T.base, T.limit and T.SP respectively.
+denoted T.base, T.limit, and T.SP respectively.
 
 The stack implementation is full-descending, so that for each thread T:
 
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -998,6 +998,10 @@ The stack implementation is full-descending, so that for each thread T:
   by the half-open interval [T.SP, T.base). The active region is empty
   when T.SP is equal to T.base.
 
+* The inactive region of T's stack is the area of memory denoted
+  by the half-open interval [T.limit, T.SP). The inactive region is empty
+  when T.SP is equal to T.limit.
+
 The stack may have a fixed size or be dynamically extendable (by adjusting the stack-limit downwards).
 
 The rules for maintenance of the stack are divided into two parts: a set of constraints that must be observed at all times, and an additional constraint that must be observed at a public interface.
@@ -1008,16 +1012,15 @@ Universal stack constraints
 At all times during the execution of a thread T, the following basic
 constraints must hold for its stack S:
 
-- T.limit ≤ T.SP ≤ T.base. T's stack pointer must lie within the memory
-  occupied by S.
+- T.limit ≤ T.SP ≤ T.base. T's stack pointer must lie within the extent
+  of the memory occupied by S.
 
 - A conforming program must not access (for reading or writing) the
-  inactive region of S, denoted by the half-open interval [T.limit, T.SP).
-  This is true regardless of whether the access is performed by T or by
-  some other thread.
+  inactive region of S. This is true regardless of whether the access
+  is performed by T or by some other thread.
 
 - If MTE is enabled, then the tag stored in T.SP must match the tag set
-  on the inactive region of S, defined in the same way as above.
+  on the inactive region of S.
 
 Additionally, at any point at which memory is accessed via SP, the hardware requires that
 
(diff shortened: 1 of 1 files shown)


rearnsha:
Whilst there is room for improvement in the wording here, I don't entirely agree with this change.  Conceptually, at least (it may be different in environments where memory is flat-mapped with a single page mapping shared by all processes), a process consists of one or more threads of execution.   Threads may either execute concurrently (on multiple compute elements in the system) or serially (being switched under the control of an operating system).

With the exception of the stack and thread-local storage the memory is managed and shared across all threads of the process.  I don't think it makes sense to say that the heap is managed by a thread - it's managed across the entire process.

Stack and thread-local data are /normally/ private to the thread, but a thread may export the addresses of certain objects to other threads, provided that it can guarantee that the contents of those objects exists and is stable at the times that other threads might access it.

rearnsha:
Of course, there is also 'shared memory': memory that is shared by two or more processes.  How such sharing is done is, I think, beyond the scope of the AAPCS and into the realms of platform-specific behaviour.

rsandifo-arm:
> Whilst there is room for improvement in the wording here, I don't entirely agree with this change. Conceptually, at least (it may be different in environments where memory is flat-mapped with a single page mapping shared by all processes), a process consists of one or more threads of execution. Threads may either execute concurrently (on multiple compute elements in the system) or serially (being switched under the control of an operating system).
> 
> With the exception of the stack and thread-local storage the memory is managed and shared across all threads of the process. I don't think it makes sense to say that the heap is managed by a thread - it's managed across the entire process.
>
> Stack and thread-local data are /normally/ private to the thread, but a thread may export the addresses of certain objects to other threads, provided that it can guarantee that the contents of those objects exists and is stable at the times that other threads might access it.

Yeah, the new wording was trying to capture that (but clearly failed :-)).  The claim wasn't that the thread “owns” these areas of memory in any sense, or that they're private to the thread.  It was just that a thread (potentially) has access to these areas of memory.  That's why it says:

> Each thread must have its own stack, but it can share other categories of memory with other threads in a process.
> (A platform might allow a thread to access the stacks of other threads, perhaps by treating them as part of the heap.)

The point with the second sentence is that, if thread T1 shares a stack object S with thread T2, S is not part of stack memory from T2's point of view.  The distinction is important because of the later requirements about which stack addresses a thread/process is allowed to access.  If we said that S was in “stack memory” even from T2's point of view, an access to S would be out of bounds (and so invalid) for T2.

So it feels like we might be saying the same thing, but in different ways.

Like you say, there are a few possible variations, and I don't think it's really up to the PCS to define exactly which memory is shared between threads in a process or between processes in a running system, beyond the requirement that the area of memory designated to be stack memory is specific to each thread.  E.g. there's no reason in principle to exclude FDPIC-like models, where code is shared between processes.

rsandifo-arm:
Hi @rearnsha.  I've (very belatedly) tried to address your concerns in the updated version of the patch.  How does this version look?

aapcs64/aapcs64.rst
```diff
@@ -948,7 +979,7 @@ At all times the following basic constraints must hold:
 
 - Stack-limit < SP ≤ stack-base. The stack pointer must lie within the extent of the stack.
 
-- A process may only access (for reading or writing) the closed interval of the entire stack delimited by [SP, stack-base – 1].
+- A conforming program may only access (for reading or writing) the closed interval of the entire stack delimited by [SP, stack-base – 1].
```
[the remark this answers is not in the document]
rsandifo-arm: Good point.  I've tried to adjust the description of the stack accordingly.  Does the new version look OK?

stuij:
@rearnsha, are you now happy with this pull request, after the latest changes of @rsandifo-arm?

aapcs64/aapcs64.rst
```diff
@@ -898,10 +898,44 @@ registers, or if it is a function that returns results in such registers,
...
+turn part of a _`process`.  A process might contain one thread or several
+threads.
+
+The exact definition of “thread” and “process” depends on the platform.
+If the platform is a traditional multi-threaded operating system,
+“thread” and “process” generally have their usual meaning for that
+operating system. If the platform supports multiple processes but has
+no separate concept of threads, “thread” is synonymous with “process”.
+If a platform has no concurrency or preemption then it will generally
+have a single “thread” and “process” that executes all instructions.
```
sallyarmneale, line 914: Avoid using quotation marks for emphasis - see Writing Guide https://confluence.arm.com/display/TCOM/Avoid+formatting+for+emphasis

aapcs64/aapcs64.rst
```diff
@@ -898,10 +898,44 @@ registers, or if it is a function that returns results in such registers,
...
+
+Each thread has its own register state, defined by the contents of the
+underlying machine registers. A process has a program state defined by
+its threads' register states and by the contents of the memory that the
+process can access. The memory that a process can access, without causing
+a run-time fault, may vary during the execution of its threads.
```
sallyarmneale, line 920:
```suggestion
a run-time fault, might vary during the execution of its threads.
```

aapcs64/aapcs64.rst
```diff
@@ -898,10 +898,44 @@ registers, or if it is a function that returns results in such registers,
...
+
+Memory and the Stack
+--------------------
+
+Memory addresses
+^^^^^^^^^^^^^^^^
+
+The address space may consist of one or more disjoint regions. No region
```
sallyarmneale, line 928:
```suggestion
The address space might consist of one or more disjoint regions. No region
```

aapcs64/aapcs64.rst
```diff
@@ -898,10 +898,44 @@ registers, or if it is a function that returns results in such registers,
...
+may span address zero (although one region may start at zero).
```
sallyarmneale, line 929:
```suggestion
might span address zero (although one region may start at zero).
```

aapcs64/aapcs64.rst
```diff
@@ -898,10 +898,44 @@ registers, or if it is a function that returns results in such registers,
...
 
-The AAPCS64 applies to a single thread of execution or process (hereafter referred to as a process). A process has a program state defined by the underlying machine registers and the contents of the memory it can access. The memory a process can access, without causing a run-time fault, may vary during the execution of the process.
+The use of tagged addressing is platform specific and does not apply to
+32-bit pointers. When tagged addressing is disabled all 64 bits of an
```
sallyarmneale, line 932:
```suggestion
32-bit pointers. When tagged addressing is disabled, all 64 bits of an
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
+- Stacks, with one stack for each thread.
+
+Each category of memory can contain multiple individual regions.
+These individual regions do not need to be contiguous and regions of one
+memory class can be interspersed with regions of another memory class.
 
-Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data. Except for the stack there is no requirement for each class of memory to occupy a single contiguous region of memory. A process must always have some code and a stack, but need not have any of the other categories of memory.
+Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data.
 
 The heap is an area (or areas) of memory that are managed by the process itself (for example, with the C malloc function). It is typically used for the creation of dynamic data objects.
```
sallyarmneale, line 958:
```suggestion
The heap is an area (or areas) of memory that the process manages (for example, with the C malloc function). It is typically used to create dynamic data objects.
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
 
+Each individual stack must occupy a single, contiguous region of memory.
+However, as noted above, multiple stacks do not need to be organized
+contiguously.
+
+A process must always have access to code and stacks, but need not have
```
sallyarmneale, line 964:
```suggestion
A process must always have access to code and stacks, but do not have
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
+
 A conforming program must only execute instructions that are in areas of memory designated to contain code.
 
+The Stack
+^^^^^^^^^
 
-Memory addresses
-^^^^^^^^^^^^^^^^
+Each thread has a stack. This stack is a contiguous area of memory that the
+thread may use for storage of local variables and for passing additional
```
sallyarmneale, line 973:
```suggestion
thread might use for storage of local variables and for passing additional
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
-The use of tagged addressing is platform specific and does not apply to 32-bit pointers. When tagged addressing is disabled all 64 bits of an address are passed to the translation system. When tagged addressing is enabled, the top eight bits of an address are ignored for the purposes of address translation. See also `Pointers`_, above.
+* a base
 
+* a limit
 
-The Stack
-^^^^^^^^^
+* the special-purpose register SP
+
+The SP moves from the base to the limit as the stack grows and from the
```
sallyarmneale, line 985:
```suggestion
The SP moves from the base to the limit as the stack grows, and from the
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
+limit to the base as the stack shrinks. In practice, an application might
+not be able to determine the value of either the base or the limit.
+
+In the description below, the base, limit and SP for a thread T are
+denoted T.base, T.limit and T.SP respectively.
```
sallyarmneale, line 990:
```suggestion
denoted T.base, T.limit, and T.SP respectively.
```

aapcs64/aapcs64.rst
```diff
@@ -913,29 +947,60 @@ The memory of a process can normally be classified into five categories:
...
+* The active region of T's stack is the area of memory delimited
+  by the half-open interval [T.SP, T.base). The active region is empty
+  when T.SP is equal to T.base.
 
-The stack implementation is full-descending, with the current extent of the stack held in the special-purpose register SP. The stack will, in general, have both a base and a limit though in practice an application may not be able to determine the value of either.
+* The inactive region of T's stack is the area of memory denoted
+  by the half-open interval [T.limit, T.SP). The inactive region is empty
+  when T.SP is equal to T.limit.
 
 The stack may have a fixed size or be dynamically extendable (by adjusting the stack-limit downwards).
```
sallyarmneale, line 1005:
```suggestion
The stack might have a fixed size or be dynamically extendable (by adjusting the stack-limit downwards).
```

aapcs64/aapcs64.rst
```diff
@@ -944,13 +1009,18 @@ The rules for maintenance of the stack are divided into two parts: a set of cons
 Universal stack constraints
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-At all times the following basic constraints must hold:
+At all times during the execution of a thread T, the following basic
```
sallyarmneale, line 1012:
```suggestion
At all times when a thread T executes, the following basic
```

sallyarmneale (commented):
A few minor changes. Always use _might_ instead of _may_. It's more easily understood by non-native speakers.

aapcs64/aapcs64.rst
rsandifo-arm: I wasn't really intending to use quotes for emphasis here.  This paragraph is trying to define what the ABI means by the terms “thread” and “process” (sometimes in circumstances where the concept of thread and process might seem unnatural).

aapcs64/aapcs64.rst
rsandifo-arm: I think “during the execution of X” is the term I'd usually expect to see here.  For example, it occurs quite often in the C++ standard.

rsandifo-arm:
> A few minor changes. Always use _might_ instead of _may_. It's more easily understood by non-native speakers.

Thanks for the reviews!

I was originally trying to keep the existing wording as much as possible, because I was worried about over-editorialising.  But a lot of the suggestions seem uncontroversial, so I've pushed an update with them included.

The “may” vs. “might” thing is trickier though.  I think we need to take an action to go through the doc and make a coordinated attempt to remove uses of “may”.  It isn't always obvious what the right replacement would be.  I get the impression that the doc is being deliberately vague about whether it's describing a state of affairs that holds naturally/inherently, or whether it's prescribing what platforms can and can't do.

For example, I think this:

> The address space may consist of one or more disjoint regions

was probably intended to be a concession: platforms _are allowed_ to create disjoint memory regions.  Similarly, I think:

> No region may span address zero

is supposed to prohibit any attempt to create regions that span address zero.  I'm not sure “might” would be a suitable replacement.

rearnsha:
> 

"May" is the correct term here.  In standards documents like these it always has the specific meaning of "permitted but not required".  Using another word or term would just be distracting.

aapcs64/aapcs64.rst
```diff
@@ -953,16 +953,16 @@ Each category of memory can contain multiple individual regions.
 These individual regions do not need to be contiguous and regions of one
 memory class can be interspersed with regions of another memory class.
 
-Writable static data may be further sub-divided into initialized, zero-initialized and uninitialized data.
+Writable static data may be further sub-divided into initialized, zero-initialized, and uninitialized data.
 
-The heap is an area (or areas) of memory that are managed by the process itself (for example, with the C malloc function). It is typically used for the creation of dynamic data objects.
+The heap is an area (or areas) of memory that the process manages itself (for example, with the C malloc function). It is typically to create dynamic data objects.
```
rearnsha, line 958: typically used to create
or
typical to create
(I think I prefer the former).

aapcs64/aapcs64.rst
rearnsha: No, we use may consistently with the specific meaning of "permitted, but not required".  This is standard practice in standards documents.

aapcs64/aapcs64.rst
sallyarmneale: In this case - I would just use the quotes in the first instance of the term in in the paragraph.

aapcs64/aapcs64.rst
sallyarmneale: could we use 'can' instead of 'may'. May is ambiguous for not-native speakers.

Can start at zero? I

aapcs64/aapcs64.rst
sallyarmneale: I'll leave this to your discretion. In Tech Comms style, we're encourage to write in the active voice where possible.

aapcs64/aapcs64.rst
statham-arm: "Can" expresses _capability_. "May" expresses _permission_. In an authoritative standard, the two are not the same – it's completely possible for there to be a thing you _can_ do, which is not also a thing that you _may_ do!

aapcs64/aapcs64.rst
sallyarmneale: The stack 'can' have a fixed size or be....?

aapcs64/aapcs64.rst
rearnsha: See my most recent comment.  'May' is the term used in equivalent standards documents for exactly the situation we have here.  In the context of standards it has a well defined and well understood meaning.
These documents are not end-user documentation. They are expected to be read by experienced developers who already understand such terms.  Trying to find alternative words is just going to cause confusion for the target audience.

aapcs64/aapcs64.rst
sallyarmneale: Would 'can' work in this instance

aapcs64/aapcs64.rst
sallyarmneale: Ok - I'll leave you to make the decision!

aapcs64/aapcs64.rst
sallyarmneale: :)

aapcs64/aapcs64.rst
```diff
@@ -944,13 +1009,18 @@ The rules for maintenance of the stack are divided into two parts: a set of cons
...
+constraints must hold for its stack S:
 
-- Stack-limit ≤ SP ≤ stack-base. The stack pointer must lie within the extent of the stack.
+- T.limit ≤ T.SP ≤ T.base. T's stack pointer must lie within the extent
+  of the memory occupied by S.
 
-- A process may only access (for reading or writing) the closed interval of the entire stack delimited by [SP, stack-base – 1].
+- A conforming program must not access (for reading or writing) the
+  inactive region of S. This is true regardless of whether the access
+  is performed by T or by some other thread.
```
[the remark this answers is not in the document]
rsandifo-arm: Yeah, but I think the idea was to tackle that as a separate PR.

aapcs64/aapcs64.rst
rsandifo-arm: I'd rather keep the paragraph here.  But I've tried to reword it to avoid excessive quoting.

PR #123, opened 2021-12-17
Add alpha-level support for SME

rsandifo-arm:
This patch adds support for SME (Scalable Matrix Extension).
At this stage it is still alpha quality.

The request also includes https://github.com/ARM-software/abi-aa/pull/122

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -1528,7 +1528,7 @@ If a call to a subroutine S `returns normally`_, the call is said to
 “commit a lazy save” *for that particular return* if all the following
 conditions are true:
 
-* ZA is dormant on entry ot S, which implies that `TPIDR2_EL0`_ has
+* ZA is dormant on entry to S, which implies that `TPIDR2_EL0`_ has
   a nonzero value (“BLK”) on entry to S.
 
 * On return from S, all the following conditions are true:
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -695,6 +695,8 @@ The definitions are recursive; that is, each of the types may contain a Composit
    type i.e. before any alignment adjustment of the entire composite is
    applied
 
+.. _`aggregate`:
+
 Aggregates
 ^^^^^^^^^^
 
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -1006,6 +1006,9 @@ The thread “has access” or “does not have access” to TPIDR2_EL0
 
 If the thread has access to SME then it must also have access to TPIDR2_EL0.
 
+The |__arm_sme_state|_ function provides a simple way of determining whether
+the current thread has access to SME or TPIDR2_EL0.
+
 TPIDR2_EL0
 ^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


sallyarmneale (approved):

aapcs64/aapcs64.rst
```diff
@@ -1023,6 +1158,604 @@ The A64 branch instructions are unable to reach every destination in the address
...
+above.  They could instead choose to save ZA before the call to S2
+and restore ZA after the call.
+
+ZA save buffers
+^^^^^^^^^^^^^^^
+
+**(Alpha)**
+
+A “\ _`ZA save buffer`\ “ is an area of memory of size W×16×\ `SVL.B`_
+bytes for some cardinal W.  The value of W is a property of the save
```
sdesmalen-arm, line 1262: The size of ZA is defined as `SVL.B x SVL.B bytes`.

What is the reason for defining the ZA save buffer as `W x 16 x SVL.B bytes` instead of:
```  W x SVL.B bytes```
where
``` num_za_save_slices <= W <= SVL.B```

Comment (rsandifo-arm) [reply aapcs64/aapcs64.rst:1262]: The current definition allows the save and restore code to be a loop that contains 16 loads and stores.  16 is special because it's maximum value for a 128-bit SME implementation (and also because LDR and STR have an immediate range of [0, 15]).

File: aapcs64/aapcs64.rst [hunk 2]
```diff
@@ -996,6 +1158,604 @@ The A64 branch instructions are unable to reach every destination in the address
...
+for `reserved bytes in the TPIDR2 block`_ is to call ``__arm_tpidr2_save``;
+see `SME support routines`_ for details.
+
+.. _`committed the lazy save`:
+
+If a call to a subroutine S `returns normally`_, the call is said to
+“commit a lazy save” *for that particular return* if all the following
+conditions are true:
+
+* ZA is dormant on entry ot S, which implies that `TPIDR2_EL0`_ has
```
paulwalker-arm, line 1490: s/entry ot/entry to/

rsandifo-arm:
@paulwalker-arm pointed out that the `__arm_tpidr2_save` spec was contradictory: it said that the function must clear TPIDR2_EL0 after saving ZA, but it also said that the function left ZA dormant.

It would have been possible to fix that by saying that the function leaves ZA active if ZA was previously dormant.  This would make the function's interface incompatible with both the private-ZA and shared-ZA specs.  However, it didn't seem worth adding a third type of ZA interface for this one function.

The new version therefore makes it the caller's responsibility to clear TPIDR2_EL0 instead.  `__arm_tpidr2_save` is then a normal private-ZA function.

rsandifo-arm:
I've updated the definition of streaming-compatible function interfaces so that they no longer have the extra PSTATE argument.  There is instead a new `__arm_sme_state` that returns whether the function has access to SME and (if so) what the current values of TPIDR2_EL0, PSTATE.SM and PSTATE.ZA are.

The man advantage of this is that an object file could define a streaming-compatible function symbol to be an alias of a non-streaming function symbol, if the non-streaming function is known to be naturally streaming-compatible.  The main disadvantage is that it's now more expensive to read PSTATE.SM in a “safe” way (that is, in a way that works even on non-SME targets).  However, it seems relatively unlikely that performance-critical paths of a streaming-compatible function would call (say) a non-streaming function; such a function should probably not be streaming-compatible.

PR #131, opened 2022-03-14
set white background on SVGs

stuij:
Previously the background of the SVGs was transparent, which makes the diagrams
hard to read on Github when the dark theme is enabled.


statham-arm:
Yes, I noticed this last week as well. It's frustrating to have to fix it this way, because you'd _like_ the images to adapt to light vs dark mode in the same way that the text does! If only SVG had some kind of colour specification along the lines of 'pick whichever of white and black is more visible against the current background'.

I suppose the obvious thing to do would be to reuse the CSS colour specifications that apply to the text, for the appropriate elements of the SVG diagrams. But that is surely more trouble than it's worth, for at least two reasons.

Firstly, it's a generality problem. The CSS in question is specific to viewing these SVGs _on Github_, and would be useless in any of the other contexts that the SVGs in this git repo will be used – perhaps even harmful, if the context happened to define the same CSS class name for some other purpose.

Secondly, it makes editing the SVGs even more painful, because surely Inkscape would not be well set up to work effectively with strange abstract colour references like that...

Issue #126, opened 2022-01-25
Please mention that C/C++ language binding is in the same document


Kristine1975:
[Section 5.3.4 of the AAPCS 32 bit](https://github.com/ARM-software/abi-aa/blob/d6e9abbc5e9cdcaa0467d8187eec0049b44044c4/aapcs32/aapcs32.rst#bit-fields) ends with:

> The layout of bit-fields within an aggregate is defined by the appropriate language binding.

Please add a note that for C/C++ this language binding is in [section 8 of the very same document](https://github.com/ARM-software/abi-aa/blob/d6e9abbc5e9cdcaa0467d8187eec0049b44044c4/aapcs32/aapcs32.rst#arm-c-and-c-language-mappings), so it's easier to find.

stuij:
closing. fixed by https://github.com/ARM-software/abi-aa/pull/132

PR #132, opened 2022-03-15
for aapcs32 and aapcs64, link mention of language bindings

stuij:
In the bitfield section of these documents there is a mention of language
bindings, which is expounded upon in another section of those documents. To make
this clear and for quick referencing, links were added to the language binding
section.

This should address https://github.com/ARM-software/abi-aa/issues/126


sallyarmneale:
Looks good to me

PR #138, opened 2022-04-04
Add aarch64_sve_vector_pcs rule to sections Scalable vector registers and Scalable Predicate Registers

MDevereau:
Functions can only gain the aarch64_sve_vector_pcs attribute implicitly by having an SVE type as the return type or as a parameter. This can cause problems when a function signature which cannot satisfy those rules requires the caller to save registers z8-z23 and p4-p15. Being able to specify the aarch64_sve_vector_pcs attribute manually enables the callee to save only the live registers, saving many cycles when calling debugging or printing functions such as printf that do not support SVE types.

Rule RWKYLB from the ARMARM SVE supplement can also be violated when the aarch64_sve_vector_pcs attribute cannot be inferred from the function signature. When SVE vector bits are greater than 128, calls from an SVE attributed caller which call a non-SVE attributed callee where the callee uses NEON code would result in the caller's SVE register high-bits being clobbered. Being able to manually specify the attribute aarch64_sve_vector_pcs resolves this.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -878,7 +878,6 @@ z0-z7 are used to pass scalable vector arguments to a subroutine, and to
 return scalable vector results from a function. If a subroutine takes
 at least one argument in scalable vector registers or scalable predicate
 registers, or if it is a function that returns results in such registers,
-or the function attribute aarch64_sve_vector_pcs is specified,
 it must ensure that the entire contents of z8-z23 are preserved across
 the call. In other cases it need only preserve the low 64 bits of z8-z15,
 as described in `SIMD and Floating-Point registers`_.
@@ -896,9 +895,17 @@ p0-p3 are used to pass scalable predicate arguments to a subroutine and
 to return scalable predicate results from a function. If a subroutine takes
 at least one argument in scalable vector registers or scalable predicate
 registers, or if it is a function that returns results in such registers,
-or the function attribute aarch64_sve_vector_pcs is specified,
 it must ensure that p4-p15 are preserved across the call. In other cases
 it need not preserve any scalable predicate register contents.
+
+aarch64_sve_vector_pcs
+^^^^^^^^^^^^^^^^^^^^^^
+
+The attribute `aarch64_sve_vector_pcs` can be specified to ensure a subroutine
+preserves the contents of Scalable Vector Registers z8-z23 and Scalable 
+Predicate Registers p4-15. This attribute can be used when a subroutine can
+benefit from preserving scalable vector and scalable predicate registers but
+has no scalable vector or scalable predicate arguments or return results.
 
 Processes, Memory and the Stack
 -------------------------------
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -878,6 +878,7 @@ z0-z7 are used to pass scalable vector arguments to a subroutine, and to
 return scalable vector results from a function. If a subroutine takes
 at least one argument in scalable vector registers or scalable predicate
 registers, or if it is a function that returns results in such registers,
+or the function attribute aarch64_sve_vector_pcs is specified,
 it must ensure that the entire contents of z8-z23 are preserved across
 the call. In other cases it need only preserve the low 64 bits of z8-z15,
 as described in `SIMD and Floating-Point registers`_.
@@ -895,6 +896,7 @@ p0-p3 are used to pass scalable predicate arguments to a subroutine and
 to return scalable predicate results from a function. If a subroutine takes
 at least one argument in scalable vector registers or scalable predicate
 registers, or if it is a function that returns results in such registers,
+or the function attribute aarch64_sve_vector_pcs is specified,
 it must ensure that p4-p15 are preserved across the call. In other cases
 it need not preserve any scalable predicate register contents.
 


fpetrogalli (commented):
Hi @MDevereau  - thank you for upstreaming this!

I have a couple of requests, please let me know what you think.

1. Given the impact on the way we change the "standard" behavior, I think that the attribute deserve its own section, as 6.1.5:

Screenshot 2022-04-04 at 17 33 50


2. The explanation in the PR is very useful - worth storing in the commit message.
3. My understanding on the intent of this attribute is "the function you are calling doesn't change of the SVE state, no need to preserve the status of the registers across the call". Couldn't that be an useful feature for NEON and SME too? In that case, how about a more generic `aarch64_no_vector_pcs` or `aarch64_no_extension_pcs` that can be used to mark all the function that do not use such extensions? It could even be added in the form where you could specify the extension you want to guarantee is not used by the function:

```
void my_function(...) __attribute__((aarch64_no_vector_pcs("sve")))
void my_function(...) __attribute__((aarch64_no_vector_pcs("advsim")))
void my_function(...) __attribute__((aarch64_no_vector_pcs("sve")),(aarch64_no_vector_pcs("sme")))
```
The meaning could be that none of the registers used by the extension needs to be preserved across the call? 
4. Finally: we discussed to add this into the aapcs... but maybe if we end up having a separate paragraph we might want to write it directly in the vector function ABI?

Thank you!

Francesco

MDevereau:
@fpetrogalli 
1. This sounds reasonable however I'm wondering how a whole section dedicated to the function attribute would look given how seamlessly it fits into 6.1.3 and 6.1.4. Would it just be as simple as "6.1.5. aarch64_sve_vector_pcs attribute" with a short explaination explaining that it also enforces the content of registers z8-23?
2. Agreed
3. I'll forward this to the team as I know little about SME and not much about NEON calling conventions. However its probably worth mentioning that the original scope for this work was to enable clang to parse the aarch64_sve_vector_pcs attribute at the c/c++ level (it already exists functionally at the LLVM level). Going down this route would likely require some heavier changes and approval for LLVM
4. I'm happy to do either, and will go with whichever gets the most votes

Thanks, Matt

paulwalker-arm:
In answer to point 3, as Matthew said, we're not looking to create a new calling standard.  Here we're only proposing a mechanism to allow users to explicitly request an existing calling standard.

peterwaller-arm:
>     3. My understanding on the intent of this attribute is "the function you are calling doesn't change of the SVE state, no need to preserve the status of the registers across the call".  

I don’t think this interpretation is right? With the `aarch64_sve_vector_pcs` calling convention in effect, the callee is taking on the responsibility to preserve the SVE register state. These functions may still have to preserve this state if they use NEON/Float instructions, or call into another function which doesn't use the SVE calling convention.

rsandifo-arm:
Since the attribute is a C/C++ feature, it would normally go in ACLE rather than the ABI.  However, this is closely related to the arm_vector_pcs attribute, so I think it should be documented in vfabi64 instead.

aapcs64/aapcs64.rst
```diff
@@ -898,6 +898,15 @@ registers, or if it is a function that returns results in such registers,
...
 it need not preserve any scalable predicate register contents.
 
+aarch64_sve_vector_pcs
+^^^^^^^^^^^^^^^^^^^^^^
+
+The attribute `aarch64_sve_vector_pcs` can be specified to ensure a subroutine
+preserves the contents of Scalable Vector Registers z8-z23 and Scalable 
+Predicate Registers p4-15. This attribute can be used when a subroutine can
+benefit from preserving scalable vector and scalable predicate registers but
+has no scalable vector or scalable predicate arguments or return results.
```
fpetrogalli, line 908: HI @MDevereau, apologies for the delay in my feedback.

This to me is more clear than the original version, thank you.

I think there is a risk that the name `aarch64_sve_vector_pcs` will get mixed up with the one defined in the Vector Function ABI (VFABI) `aarch64_vector_pcs`. The two are kind of related, but given that `aarch64_vector_pcs` is also about SVE, I think we should try to rename the new attribute in a way that avoids mixing it up with the VFABI one. I think we should reserve the `aarch64_vector*` prefix for the VFABI. Also, the `SVE PCS` clearly says that the PCS we want to use is attached to functions that use SVE registers in their interface. We are enforcing the SVE PCS into a function that would not qualify as a function that would have to obey to the rules we want to enforce. 

Here is what I suggest:

The SVE-specific PCS is attached to functions that (lines 878-880)

>  If a subroutine takes
at least one argument in scalable vector registers or scalable predicate
registers, or if it is a function that returns results in such registers,

We could extract this text into the definition of `SVE function`, and rename the attribute introduced in this patch as `aarch64_force_sve_function`. The PCS for SVE would read as:

```
SVE function
^^^^^^^^^^^^

An _SVE function_ is  a subroutine that takes
at least one argument in scalable vector registers or scalable predicate
registers, or a function that returns results in such registers.

Scalable vector registers
^^^^^^^^^^^^^^^^^^^^^^^^^
The Arm 64-bit architecture also defines [...]
z0-z7 are used to pass scalable vector arguments to a subroutine, and to
return scalable vector results from a function. An SVE function (link to the definition) must 
ensure that the entire contents of z8-z23 are preserved across
the call. In other cases [...]

Scalable Predicate Registers
^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[similar changes]

aarch64_force_sve_function
^^^^^^^^^^^^^^^^^^^^^^^^^^

The attribute `aarch64_force_sve_function` can be specified to ensure a subroutine is interpreted as an SVE function (link to definition) even if it does not match the definition of it.  [etc etc.. ].
```

Thanks,

Francesco

fpetrogalli:
> 



> In answer to point 3, as Matthew said, we're not looking to create a new calling standard. Here we're only proposing a mechanism to allow users to explicitly request an existing calling standard.

Hi @paulwalker-arm - yeah, I think I diverted too much from the original intent of this patch with my suggestions.

Francesco

fpetrogalli:
> Since the attribute is a C/C++ feature, it would normally go in ACLE rather than the ABI. However, this is closely related to the arm_vector_pcs attribute, so I think it should be documented in vfabi64 instead.

I saw this comment after submitting mine about the name change, and how `aarch64_sve_vector_pcs` and `aarch64_vector_pcs` should be kept separate :)

The `vector PCS` attribute redirects to the SVE bits of the PCS, and it boils down to the rule: _*if* it is an SVE function, *then* preserve this and that_.

The `SVE vector PCS` we are introducing is just removing the *if* condition of the SVE bits of the PCS: it says that the function _ *is* an SVE function_, not matter what, and for this reason it must follow the rules of the SVE bits of the PCS.

So the attribute is marking as `SVE function` a function that is not an `SVE function`.

I think that the name `aarch64_force_sve_function` is better suited for this operation.

Francesco

MDevereau:
Abandoning this pull request. The addition of this will instead be added to the ACLE repository here https://github.com/ARM-software/acle/pull/194

PR #141, opened 2022-04-11
add Armv8-AR value to the AArch32 build attibute Tag_CPU_arch tag

stuij:


smithp35 (approved):
LGTM. Unlike v7 where there is a reasonable intersection across all profiles. There isn't a particularly useful v8. The use of Tag_CPU_arch_profile is deprecated for v8 and above so this is not a viable alternative.

rearnsha:
Adding bespoke pseudo architectures for specific compiler vendors does not seem the right way to address this.  It simply doesn't scale as there are many possible permutations and adding them in an ad-hoc manner is unprincipled.
If this were to be sanctioned in the ABI, the right way would be to extend Tag_also_compatible_with, but historically that was frowned upon because of the difficulties in defining compatibility.  This proposed change in no way tries to address any of those difficulties.
I think the right thing to do is for the ABI attributes to pick one of the two architectures as the 'reference' architecture and then use a private 'vendor' attribute to describe the additional compatibility.  Then any additional complication and difficulties of interpretation is passed back entirely to the vendor toolchain that intends to support this, and isn't exposed to other implementers of the ABI for them to worry about.

stuij:
Abandoning this as we're taking [rearnsha](https://github.com/rearnsha)'s advice, and we're currently working on a solution involving `Tag_also_compatible_with`.

PR #148, opened 2022-04-25
Document new PT_AARCH64_MEMTAG_MTE segment

luis-machado-arm:
Add PT_AARCH64_MEMTAG_MTE to the table of AARCH64 segment types and document
its purpose.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1725,15 +1725,15 @@ p\_type
 
 .. table:: Processor-specific segment types
 
-    +-------------------------+-------------+------------------------------------------------------+
-    | Name                    | p\_type     | Meaning                                              |
-    +=========================+=============+======================================================+
-    | PT\_AARCH64\_ARCHEXT    | 0x70000000  | Reserved for architecture compatibility information  |
-    +-------------------------+-------------+------------------------------------------------------+
-    | PT\_AARCH64\_UNWIND     | 0x70000001  | Reserved for exception unwinding tables              |
-    +-------------------------+-------------+------------------------------------------------------+
-    | PT\_AARCH64\_MEMTAG_MTE | 0x70000002  | Reserved for MTE memory tag data dumps in core files |
-    +-------------------------+-------------+------------------------------------------------------+
+    +--------------------------+-------------+------------------------------------------------------+
+    | Name                     | p\_type     | Meaning                                              |
+    +==========================+=============+======================================================+
+    | PT\_AARCH64\_ARCHEXT     | 0x70000000  | Reserved for architecture compatibility information  |
+    +--------------------------+-------------+------------------------------------------------------+
+    | PT\_AARCH64\_UNWIND      | 0x70000001  | Reserved for exception unwinding tables              |
+    +--------------------------+-------------+------------------------------------------------------+
+    | PT\_AARCH64\_MEMTAG\_MTE | 0x70000002  | Reserved for MTE memory tag data dumps in core files |
+    +--------------------------+-------------+------------------------------------------------------+
 
 A segment of type ``PT_AARCH64_ARCHEXT`` (if present) contains information describing the architecture capabilities required by the executable file. Not all platform ABIs require this segment; the Linux ABI does not. If the segment is present it must appear before segment of type ``PT_LOAD``.
 
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -262,27 +262,29 @@ This document refers to, or is referred to by, the following documents.
 
 .. table::
 
-  +------------------------------+-------------------------------------------------------------+-------------------------------------------------------------+
-  | Ref                          | External reference or URL                                   | Title                                                       |
-  +==============================+=============================================================+=============================================================+
-  | AAELF64                      | Source for this document                                    | ELF for the Arm 64-bit Architecture (AArch64).              |
-  +------------------------------+-------------------------------------------------------------+-------------------------------------------------------------+
-  | AAPCS64_                     | IHI 0055                                                    | Procedure Call Standard for the Arm 64-bit Architecture     |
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -1722,18 +1722,22 @@ p\_type
...
+    | PT\_AARCH64\_UNWIND     | 0x70000001  | Reserved for exception unwinding tables              |
+    +-------------------------+-------------+------------------------------------------------------+
+    | PT\_AARCH64\_MEMTAG_MTE | 0x70000002  | Reserved for MTE memory tag data dumps in core files |
+    +-------------------------+-------------+------------------------------------------------------+
 
 A segment of type ``PT_AARCH64_ARCHEXT`` (if present) contains information describing the architecture capabilities required by the executable file. Not all platform ABIs require this segment; the Linux ABI does not. If the segment is present it must appear before segment of type ``PT_LOAD``.
 
 ``PT_AARCH64_UNWIND`` (if present) describes the location of a program’s exception unwind tables.
 
+``PT_AARCH64_MEMTAG_MTE`` segments (if present) hold MTE memory tags for a particular memory range. The data is packed as 2 MTE tags per byte. There can be multiple ``PT_AARCH64_MEMTAG_MTE`` segments in a core file.
```
smithp35, line 1739: Just to check my understanding this Program Header is only expected to be present in a Core File? Which is why it would be still be appropriate to write "Reserved for" in the table above.

If that is the case, I think it would be worth mentioning that the segment type can only appear in Core files. Is there a canonical reference to the Core file format we can add as a reference, I found that the ELF type is ET_CORE but it wasn't a very authoratitive reference.

I think it may be better to describe the format of the program header in a different document. For example if Core files are Linux/Android specific then maybe the sysvabi64 (https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst) is a better place for the details, or perhaps even the GDB documentation? You'd then only need to reference that from here.

One thing I can't work out is how the memory range is deduced. Is it [p_vaddr, p_vaddr + p_memsz) and the location of the tags is [p_offset, p_offset + p_filesz)? May be worth being explicit. Could also be useful to describe the order within the byte and whether it is always the same or different for big-endian.

luis-machado-arm: > Just to check my understanding this Program Header is only expected to be present in a Core File? Which is why it would be still be appropriate to write "Reserved for" in the table above.
> 

That's correct, right now we only want to use it for core files. There has been some ideas about having memory tag data that would be used to initialize memory on program load, but that's is only an idea. I didn't want to commit to locking this to core files. What do you think?

> If that is the case, I think it would be worth mentioning that the segment type can only appear in Core files. Is there a canonical reference to the Core file format we can add as a reference, I found that the ELF type is ET_CORE but it wasn't a very authoratitive reference.
> 

Unfortunately core file formats are not very well documented. It seems like nobody wants to own it.

> I think it may be better to describe the format of the program header in a different document. For example if Core files are Linux/Android specific then maybe the sysvabi64 (https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst) is a better place for the details, or perhaps even the GDB documentation? You'd then only need to reference that from here.
> 

This particular format is already documented in the Linux Kernel (Documentation/arm64/memory-tagging-extension.rst - Core Dump Support).

> One thing I can't work out is how the memory range is deduced. Is it [p_vaddr, p_vaddr + p_memsz) and the location of the tags is [p_offset, p_offset + p_filesz)? May be worth being explicit. Could also be useful to describe the order within the byte and whether it is always the same or different for big-endian.

The memory range is [p_vaddr, p_vaddr + p_memsz), and the tags are within the [p_offset, p_offset + p_filesz) range. So your understanding is correct.

The order within the byte is always the same, and it goes from the first memory tag granule from the memory range to the last.

smithp35: > That's correct, right now we only want to use it for core files. There has been some ideas about having memory tag data that would be used to initialize memory on program load, but that's is only an idea. I didn't want to commit to locking this to core files. What do you think?

I think the benefit of saying only for core files, or at least "at present, only for core files" is that it means that readers not concerned with core files can look away now.  When I first looked at it I put my static linker glasses on and thought how would the linker generate these program headers, for example would the compiler generate sections that the linker would aggregate, or would the linker generate the sections from static relocations?

My guess is that we'd be much more likely to use dynamic relocations (even when static-linking) to initialise tags, much the same way that capabilities are initialized in Morello.

My preference would be to just reserve the Program Header in this document and reference the Linux kernel documentation for the details. For example:

``PT_AARCH64_MEMTAG_MTE`` segments (if present) hold MTE memory tags for a particular memory range. At present they are defined for core dump files of type ``ET_CORE``. A description of the program header and contents can be found in [`MTEEXTENSIONS`_].

With MTEEXTENSIONS defined in the references section, pointing to https://www.kernel.org/doc/html/latest/arm64/memory-tagging-extension.html#core-dump-support

luis-machado-arm: Thanks for the input Peter. Let me rework it.

smithp35 (approved):
Thanks for the update. I'm happy with the changes.

luis-machado-arm:
Sorry. I missed the bit Richard mentioned above. Fixed now.

luis-machado-arm:
Hi Peter/Richard. Could either of you approve this change again, after the typo fix?

MaskRay:
There is a glibc patch adding `PT_AARCH64_MEMTAG_MTE` (https://sourceware.org/pipermail/libc-alpha/2022-June/139314.html). The Linux kernel added the value earlier. ISTM it'd be better for the ABI change to be pushed earlier than kernel/glibc/etc.

stuij:
Sorry for the lag on this @luis-machado-arm. Just noticed this one was still lingering in the system.

PR #149, opened 2022-04-28
Include the ABS64 and ABS32 relocations in Dynamic relocations

mmalcomson:
These relocations are special since they are both dynamic and static
relocations.  At the start of the "Relocation" section (currently 5.7)
we state that "A well-formed executable file or dynamic shared object
has no static relocations after static linking".  At the moment there is
nothing in this document which mentions that the above relocations are
both dynamic and static, and hence it could be confusing to see such
relocations in an executable if just going based on the ABI
documentation.

This commit adds both relocations to the "Dynamic Relocations" table in
order to make the document self-consistent.  We also include a note
explaining that these relocations are special.

--- Explanation of some decisions:
In the dynamic relocations table I do not mention the ELF64 code for
ABS32 even though there is such a code.  This is because for ELF64 the
ABS32 relocation is not dynamic.

I include a comment about how these relocations are dynamic and static
for extra clarity, even though that is supposed to be implicit in the
fact that they occur in both tables.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1609,6 +1609,10 @@ The dynamic relocations for those execution environments that support only a lim
   +------------+------------+-------------------------+------------------------+-------------------------------------------+
   | ELF64 Code | ELF32 Code | Name                    | Operation              | Comment                                   |
   +============+============+=========================+========================+===========================================+
+  | 257        | \-         | R\_<CLS>\_ABS64         | S + A                  | See note below.                           |
+  +------------+------------+-------------------------+------------------------+-------------------------------------------+
+  | \-         | 1          | R\_<CLS>\_ABS32         | S + A                  | See note below.                           |
+  +------------+------------+-------------------------+------------------------+-------------------------------------------+
   | 1024       | 180        | R\_<CLS>\_COPY          |                        | See note below.                           |
   +------------+------------+-------------------------+------------------------+-------------------------------------------+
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -1635,6 +1639,8 @@ The dynamic relocations for those execution environments that support only a lim
 
 With the exception of ``R_<CLS>_COPY`` all dynamic relocations require that the place being relocated is an 8-byte aligned 64-bit data location in ELF64 or a 4-byte aligned 32-bit data location in ELF32.
 
+``R_<CLS>_ABS64`` and ``R_<CLS>_ABS32`` may only appear in a well-formed executable or dynamic shared object in ELF64 or ELF32 respectively.  Note that for their respective file format these relocations are both static and dynamic relocations.
```
smithp35, line 1642: Looks like there are two spaces between "respectively.  Note" not sure if this will show up in the rendering, but probably worth fixing if non-intentional.

smithp35 (commented):
One small nit to do with spacing, but otherwise looks good to me.

aaelf64/aaelf64.rst
mmalcomson: Heh -- GNU comment style creeping in, will fix ;-)

smithp35 (approved):
Thanks for the update. LGTM.

stuij (approved):

PR #154, opened 2022-06-24
Fix typo: .ehframe => .eh_frame

MaskRay:


smithp35 (approved):
Thanks for the correction. Hopefully can get this merged soon.

mmalcomson:
I have no idea how easy it would be to fix this, but it seems that the `.eh_frame` text in the diagram that ends up in the final PDF has some strange spacing so that it looks like `.eh_fr a me`.
It's a minor thing, but I figured I'd mention it in case anyone knows how to fix it.

MaskRay:
Ran `tools/inkscape/export-svg.sh sysvabi64/sysvabi64-elf-layout-source.svg sysvabi64/sysvabi64-elf-layout.svg`. Looks like `sysvabi64/sysvabi64-elf-layout.svg` has many changes because it was not generated by inkscape?

smithp35:
That's possible, I know I created the diagram outside of inkscape, however I can't remember what post-processing was done to get the two svg files.

Looks like the new diagram has fixed the rendering, at least in github's markup. Thanks for the update!

mmalcomson:
Also looks like the change has propagated to the PDF's that the CI generates.
Thanks!

PR #155, opened 2022-06-24
[aaelf64] ADRP + LDR GOT relaxation symbol should not be absolute

smithp35:
When position independent code is used (linking -fpic or -fpie)
it is not safe to relax ADRP + LDR to ADRP + ADD when the destination
symbol is absolute (st_shndx = SHN_ABS) as the symbol address will
always be the same and ADRP + ADD is relative. The relaxation can
still be applied when all addresses are fixed.

https://reviews.llvm.org/D128492 for LLD patch.


mmalcomson:
Looks good to me, tricky little detail.

PR #156, opened 2022-06-24
Update rstcheck options for github CI

mmalcomson:
The current version of rstcheck (i.e. the version that the github CI
uses) does not have an option `--ignore-language` but instead has
`--ignore-languages`.

Without this update to our rst syntax checking script the github CI
fails on every PR.

N.b. The `rstcheck` change was here https://github.com/rstcheck/rstcheck/pull/100
N.b. I don't really know how to run the github actions CI on a PR change.
Hoping that this PR triggers it ;-)
E.g. https://github.com/ARM-software/abi-aa/pull/155

### DIFF
--- a/tools/common/check-rst-syntax.sh
+++ b/tools/common/check-rst-syntax.sh
@@ -38,5 +38,5 @@ declare -a docs=(
 )
 
 for doc in "${docs[@]}"; do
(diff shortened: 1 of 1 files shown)


smithp35 (approved):
Looks good to me. Thanks for the fix. We'll need to find someone that can merge fixes to apply this as soon as possible.

PR #159, opened 2022-07-14
[addenda32] Fix typo in NTBS

vhscampos:
Changing from NBTS to NTBS.


stuij (approved):
Approved. This should clearly read NTBS (Null Terminated Byte String), as it's an idiom. NTBS with regards to ULEB128 is also mentioned elsewhere in the document.

Issue #152, opened 2022-06-23
[SYSVABI64] Complete program linking and dynamic linking section


smithp35:
Document AArch64 PLT sequences, include BTI and PAC options

Move the AArch64 Feature bits and dynamic tags as they are more suited to the SYSVABI than AAELF64.

PR #160, opened 2022-07-14
[sysvabi64] Add chapter on Program Loading and Dynamic Linking.

smithp35:
Replace the current placeholder chapter with information on:
* ELF
* GOT and PLT
* GNU Indirect Functions
* Relocation Read-Only

A lot of the information in the chapter is general information. Not
specific to AArch64. However it can be useful to put the AArch64
information in context in a single place.

Resolves https://github.com/ARM-software/abi-aa/issues/152

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1111,8 +1111,7 @@ consider a symbol not pre-emptable if:
    ``--dynamic-list`` is used.
 
 Shared objects must indirect all function calls to pre-emptable
-symbols must be indirected through the static linker created Procedure
-Linkage Table.
+symbols through the static linker created Procedure Linkage Table.
 
 Procedure Linkage Table
 -----------------------
@@ -1168,16 +1167,16 @@ procedure linkage table and the global offset table.
    dynamic section entry. The designated relocation entry must have
    type ``R_AARCH64_JUMP_SLOT``, and its ``r_offset`` field will
    specify the ``.got.plt`` table entry used by the PLT entry in
-   ``PLT[N]``, and also a symbol table index that references the
-   appropriate symbol, i.e. ``name1`` in this example.
+   ``PLT[N]``, the ``r_info`` encodes a symbol table index that
+   references the appropriate symbol, i.e. ``name1`` in this example.
 
  * The ``PLT[0]`` entry then loads the adress of the third
-   ``.plt.got`` entry into ``ip0``, and then loads and jumps to the
+   ``.got.plt`` entry into ``ip0``, and then loads and jumps to the
    address in the third table entry, which transfers control to the
    dynamic linker's lazy resolver function.
 
  * The value in ``ip0`` is used by the dynamic linker's entrypoint to
-   load the second ``.plt.got`` entry using ``[ip0,#-8]``, which gives
+   load the second ``.got.plt`` entry using ``[ip0,#-8]``, which gives
    it one 64-bit word of private information with which to identify
    the calling module.
 
@@ -1226,7 +1225,7 @@ Sample PLT sequences
 ^^^^^^^^^^^^^^^^^^^^
 
 The following section shows some examples of PLT sequences. The
-sequences assume that the ``.plt.got`` section is within +- 4 GiB from
+sequences assume that the ``.got.plt`` section is within +- 4 GiB from
 the ``.plt`` section.
 
 In the code samples below we assume that the entries in the
@@ -1246,17 +1245,17 @@ PLT header ``PLT[0]``
 
     stp    x16, x30, [sp,#-16]!
     adrp   x16, :page: &.got.plt[2]
-    ldr    x17, [x16, (&.got.plt[2] - &.got.plt[0])]
-    add    x16, x16, (&got.plt[2] - &got.plt[0])
+    ldr    x17, [x16, :lo12: &.got.plt[2]]
+    add    x16, x16, :lo12: &got.plt[2]
     br     x17
 
 Nth PLT entry ``PLT[N]``
 
 .. code-block:: asm
 
-    adrp x16, :page: &.plt.got[N + 3]
-    ldr  x17, [x16, (&.got.plt[N + 3] - &.got.plt[0])
-    add  x16, x16, (&.got.plt[N + 3] - &.got.plt[0])
+    adrp x16, :page: &.got.plt[N + 3]
+    ldr  x17, [x16, :lo12: &.got.plt[N + 3]]
+    add  x16, x16, :lo12: &.got.plt[N + 3]]
     br   x17
 
 BTI
@@ -1277,8 +1276,8 @@ PLT header ``PLT[0]``
     bti  c
     stp  x16, x30, [sp,#-16]!
     adrp   x16, :page: &.got.plt[2]
-    ldr    x17, [x16, (&.got.plt[2] - &.got.plt[0])]
-    add    x16, x16, (&got.plt[2] - &got.plt[0])
+    ldr    x17, [x16, :lo12: &.got.plt[2]]
+    add    x16, x16, :lo12: &got.plt[2]
     br   x17
 
 Nth PLT entry ``PLT[N]``
@@ -1286,9 +1285,9 @@ Nth PLT entry ``PLT[N]``
 .. code-block:: asm
 
     bti  c
-    adrp x16, :page: &.plt.got[N + 3]
-    ldr  x17, [x16, (&.got.plt[N + 3] - &.got.plt[0])
-    add  x16, x16, (&.got.plt[N + 3] - &.got.plt[0])
+    adrp x16, :page: &.got.plt[N + 3]
+    ldr  x17, [x16, :lo12: &.got.plt[N + 3]]
+    add  x16, x16, :lo12: &.got.plt[N + 3]
     br   x17
 
 PAC
@@ -1311,9 +1310,9 @@ Nth PLT entry ``PLT[N]``
 
 .. code-block:: asm
 
-    adrp x16, :page: &.plt.got[N + 3]
-    ldr  x17, [x16, (&.got.plt[N + 3] - &.got.plt[0])
-    add  x16, x16, (&.got.plt[N + 3] - &.got.plt[0])
+    adrp x16, :page: &.got.plt[N + 3]
+    ldr  x17, [x16, :lo12: &.got.plt[N + 3]]
+    add  x16, x16, :lo12: &.got.plt[N + 3]
     autia1716
     br   x17
 
@@ -1335,8 +1334,8 @@ PLT header ``PLT[0]``
     bti  c
     stp  x16, x30, [sp,#-16]!
     adrp   x16, :page: &.got.plt[2]
-    ldr    x17, [x16, (&.got.plt[2] - &.got.plt[0])]
-    add    x16, x16, (&got.plt[2] - &got.plt[0])
+    ldr    x17, [x16, :lo12: &.got.plt[2]]
+    add    x16, x16, :lo12: &got.plt[2]
     br   x17
 
 Nth PLT entry ``PLT[N]``
@@ -1344,9 +1343,9 @@ Nth PLT entry ``PLT[N]``
 .. code-block:: asm
 
     bti  c
-    adrp x16, :page: &.plt.got[N + 3]
-    ldr  x17, [x16, (&.got.plt[N + 3] - &.got.plt[0])
-    add  x16, x16, (&.got.plt[N + 3] - &.got.plt[0])
+    adrp x16, :page: &.got.plt[N + 3]
+    ldr  x17, [x16, :lo12: &.got.plt[N + 3]]
+    add  x16, x16, :lo12: &.got.plt[N + 3]
     autia1716
     br   x17
 
@@ -1358,7 +1357,7 @@ implementation of a function to be chosen from multiple candidates,
 with the choice taken by an IFUNC resolver function.
 
 GNU Indirect Functions require static and dynamic linker support. They
-are known to be supported on GNU/Linux, Android and many of the BSD
+are known to be supported on GNU/Linux, Android, and many of the BSD
 operating systems.
 
 GNU Indirect Functions are called via a PLT entry that loads the
@@ -1414,10 +1413,10 @@ the GCC and Clang compilers an attribute can be used to achieve this.
   /* make symbol ifunc type STT_GNU_IFUNC using resolver as the IFUNC resolver. */
   int ifunc(void) __attribute__((ifunc("resolver")));
 
-The first ``uint64`` parameter is ``AT_HWCAP``. If the
+The first ``uint64_t`` parameter is ``AT_HWCAP``. If the
 ``_IFUNC_ARG_HWCAP`` flag is set the second parameter is present. The
 second parameter is a structure ``__ifunc_arg_t`` defined in
-``sys/ifunc.h`` it provides access to all of the flags in HWCAP_ via
+``sys/ifunc.h``, it provides access to all of the flags in HWCAP_ via
 fields of the structure.
 
 The IFUNC resolver function returns the address of the function
@@ -1433,10 +1432,6 @@ additional restrictions on what they can contain.
 
  * IFUNC resolvers must not have a symbol binding of ``STB_WEAK``.
 
- * IFUNC resolvers that may be called via lazy-binding must be
-   asynchronous-signal safe and thread-safe as the resolver may be
-   called from a signal handler or multi-threaded process.
-
 The order of dynamic relocation resolution across an executable and
 all its shared libraries is platform specific. The following
 recommendations for writing IFUNC resolvers apply to the GNU glibc
@@ -1449,7 +1444,7 @@ dynamic loader. Other dynamic linkers may have fewer requirements.
  * In position-independent code an IFUNC resolver functions must not
    call a function that requires a PLT entry. If the IFUNC resolver
    runs as a result of a relocation in ``.rela.dyn`` then the
-   relocations in ``rela.plt`` will not have been resolved. This means
+   relocations in ``.rela.plt`` will not have been resolved. This means
    that addresses in the ``.got.plt`` will be unchanged from their
    static link time value.
 
@@ -1481,7 +1476,7 @@ resolved to a PLT entry that loads the value of:
 Due to the ordering requirements on IFUNC resolvers the PLT entry and
 associated ``.got.plt`` entry are by often implemented in a separate
 ``.iplt`` and ``.iplt.got`` sections so that they placed after the
-``.plt`` and ``.plt.got`` sections respectively.
+``.plt`` and ``.got.plt`` sections respectively.
 
 Like ``R_AARCH64_RELATIVE`` the ``R_AARCH64_IRELATIVE`` relocation
 does not require a symbol. For ``RELA`` type relocations the addend of
@@ -1539,7 +1534,7 @@ specified by ``DT_FINI`` and ``DT_FINI_ARRAY``.
 Relocation Read Only (RELRO)
 ----------------------------
 
-Several sections in an exectuable or shared-library are logically
+Several sections in an executable or shared-library are logically
 read-only but they require dynamic relocation which forces them to be
 writeable. Relocation Read Only (RELRO) is a GNU extension to ELF that
 permits a dynamic linker to remap the pages described by the RELRO
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -395,10 +395,10 @@ Absolute addressing is suitable for most bare-metal code, including
 the Linux kernel, as well as for normal GNU/Linux executables which –
 while dynamically linked – are loaded at a fixed address.
 
-Position independent addressing
+Position-independent addressing
 -------------------------------
 
-In position independent code (PIC) the virtual addresses of
+In position-independent code (PIC) the virtual addresses of
 instructions and static data are not known until dynamic link
 time.
 
@@ -416,8 +416,8 @@ linker to create the GOT entry. The address of symbol definitions that
 cannot be pre-empted at dynamic link time can have their address
 taken so no GOT generating relocation is required.
 
-PIC can also be used to build position independent executables. A
-variant of PIC called PIE (position independent executable) can be
+PIC can also be used to build position-independent executables. A
+variant of PIC called PIE (position-independent executable) can be
 used to build an executable. PIE assumes that global symbols cannot be
 pre-empted, which means that an indirection via the GOT is not needed.
 
@@ -462,7 +462,7 @@ syntax is of the form ``#:<operator>:<symbol name>``
   | ``abs_g3``            | ``mov[nz]`` | R_AARCH64_MOVW_UABS_G3     |
   +-----------------------+-------------+----------------------------+
 
-.. table:: Position independent operators
+.. table:: Position-independent operators
 
   +-----------------------+-------------+-------------------------------+
   | Operator              | Instruction | Relocation                    |
@@ -737,7 +737,7 @@ The convention for command-line option to select code model is
 ``-mcmodel=<model>`` where code model is one of ``tiny``, ``small``,
 ``medium`` or ``large``.
 
-The convention for command-line option to select position independent
+The convention for command-line option to select position-independent
 code is ``-fpic`` for pic ``-fPIC`` for PIC. When compiling for an
 executable ``-fpie`` and ``-fPIE`` have the same GOT size limitations
 as ``-fpic`` and ``-fPIC`` respectively.
@@ -807,8 +807,8 @@ effects on code-generation of the code-models.
 Get the address of a symbol defined in the same ELF file
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Code that is not position independent may use the absolute address of
-the symbol. Code that is position independent may use a pc-relative
+Code that is not position-independent may use the absolute address of
+the symbol. Code that is position-independent may use a pc-relative
 offset to the symbol if the definition of the symbol is not
 pre-emptible. If the symbol is pre-emptible the address must be loaded
 from the GOT.
@@ -933,8 +933,8 @@ page is mapped into physical memory depends on the execution behaviour
 of the program and the working set that it requires for its correct
 execution. To minimize the number of such reads from the physical
 medium containing the program, executables and shared objects must
-have program segments whose file offsets and virtual addresses are
-congruent modulo the page size.
+have loadable program segments whose file offsets and virtual
+addresses are congruent modulo the page size.
 
 Since the architecture supports a maximum page size of 64KiB the page
 size should be set to 64KiB for compatibility with systems that use
@@ -961,17 +961,19 @@ the other for mapping as data.
 Shared objects are always built with a base virtual address of 0,
 while executables with an ELF ``e_type`` of ``ET_EXEC`` are built with
 a fixed virtual address. The underlying operating system maps shared
-objects in at different virtual addresses, utilizing the fact that the
-code and data addressing within such a shared object is position
-independent. The position independence relies on the principle that
-the distance between any two points within the image is identical at
-static link time and at run time.
+objects at different virtual addresses, utilizing the fact that the
+code and data addressing within such a shared object is
+position-independent. The position independence relies on the
+principle that the offset between loadable program segments is fixed
+at static link time. This permits code to access data in another
+loadable program segment using a pc-relative offset that will not
+change at run time.
 
 Program segments for executables built with a fixed virtual address
 have to be mapped identically to the virtual addresses used to create
 the executable file.
 
-Position independent executables are mapped in the same way as shared
+Position-independent executables are mapped in the same way as shared
 objects. They have an ELF ``e_type`` of ``ET_DYN``.
 
 Dynamic Section
@@ -988,13 +990,13 @@ Section.
 Global Offset Table (GOT)
 -------------------------
 
-Position independent code cannot, in general, contain absolute (fixed)
+Position-independent code cannot, in general, contain absolute (fixed)
 virtual addresses. Global offset tables hold absolute addresses in
 private data, thus making the addresses available without compromising
-the position independence and shareability of a program's text
+the position-independence and shareability of a program's text
 segment. A program references its global offset table using
-position independent addressing and extracts the absolute values from
-it, thereby redirecting position independent references to their
+position-independent addressing and extracts the absolute values from
+it, thereby redirecting position-independent references to their
 actual locations.
 
 The global offset table (GOT) is created by the static linker in
@@ -1011,20 +1013,26 @@ AArch64 splits the global offset table (GOT) into two sections:
 Both the ``.got`` and ``.got.plt`` sections are aligned to a minimum
 of 8 bytes. They both contain 8-byte sized entries.
 
-If a program requires direct access to the absolute address of a
-symbol, that symbol will have a GOT entry. Because the executable and
-shared objects have their own independent GOTs, a symbol's address may
-appear in several tables. The dynamic linker processes all of the
-relocations in the ``.got`` section before giving control to any code
-in the process image, thus ensuring the absolute addresses are
-available during execution.
+If a position-independent program requires direct access to the
+absolute address of a symbol, that symbol will have a GOT
+entry. Because the executable and shared objects have their own
+independent GOTs, a symbol's address may appear in several tables. The
+dynamic linker processes all of the relocations in the ``.got``
+section before giving control to any code in the process image, thus
+ensuring the absolute addresses are available during execution. Code
+in the process that runs as part of a GNU indirect function resolver
+is an exception to this rule. See `GNU Indirect Functions`_ for the
+restrictions on what addresses can be accessed from a resolver
+function.
 
 By convention the ``.got`` section's first entry (number zero) is
 reserved to hold the address of the dynamic structure, referenced with
-the linker defined symbol ``_DYNAMIC``. Some dynamic linker's use this
-address or another reserved symbol to find its own dynamic structure
-without without having yet processed their own relocation entries. For
-AArch64 entries one and two in the ``.got`` are also reserved.
+the linker defined symbol ``_DYNAMIC``. Versions of glibc prior to
+version 2.35 use this address to find its own dynamic structure
+without without having yet processed their own relocation entries.
+
+AArch64 entries one and two in the ``.got.plt`` are reserved for the
+dynamic linker.
 
 For AArch64 the linker defined ``_GLOBAL_OFFSET_TABLE_`` symbol should
 be the address of the first global offset table entry in the ``.got``
@@ -1037,14 +1045,15 @@ Function Addresses
 ^^^^^^^^^^^^^^^^^^
 
 Direct function calls are those where the name of the called function
-is known at compile time. The PC-relative BL instruction may be used
-for all direct function calls, whether absolute or position
-independent.
+is known at compile time. The PC-relative direct branch instructions
+may be used for all direct function calls, whether absolute or
+position-independent.
 
 Indirect function calls are those where the address of the function is
 in a pointer. Appropriate code is used to load the value of the
-pointer into a register, as for other data, and then a BLR instruction
-is used.
+pointer into a register, as for other data, and then an indirect
+branch instruction is used with the register as an operand to the
+instruction.
 
 In statically linked code the address of a function is always its real
 address. However in dynamically linked environments references to the
@@ -1055,23 +1064,24 @@ References from within shared objects to a function address will
 normally be resolved by the dynamic linker to the virtual address of
 the function itself.
 
-References from within a non position independent executable file to a
+References from within a non-position-independent executable file to a
 function address defined in a shared object will normally be resolved
 by the static linker to the address of the `Procedure Linkage Table`_
-(PLT) for that function within the executable file. If the address is
-in a writeable location the linker may use a dynamic relocation
-instead.
+(PLT) entry for that function within the executable file. If the
+address is in a writeable location the linker may use a dynamic
+relocation instead.
 
-When both shared objects and a non position independent executable use
+When both shared objects and a non-position-independent executable use
 the address of a function both the shared library and the application
-must use the same address for the functions for address comparisons to
-work as expected. If the static linker uses a PLT entry for a function
-as the address for a function then it must place the address of the
-PLT for the function in the dynamic symbols table entry for the
-function. This will result in a dynamic symbol table entry with a
-section index of ``SHN_UNDEF``, a type ``STT_FUNC`` and a non-zero
-``st_value``. A reference to this symbol will be resolved by the
-dynamic linker to the address PLT for the function in the executable.
+must use the same address for the functions in order for address
+comparisons to work as expected. If the static linker uses a PLT entry
+for a function as the address for a function then it must place the
+address of the PLT for the function in the dynamic symbols table entry
+for the function. This will result in a dynamic symbol table entry
+with a section index of ``SHN_UNDEF``, a type ``STT_FUNC`` and a
+non-zero ``st_value``. A reference to this symbol will be resolved by
+the dynamic linker to the address PLT for the function in the
+executable.
 
 The ``R_<CLS>_JUMP_SLOT`` relocations defined in AAELF64_ are
 associated with PLT entries. These entries are used for direct
@@ -1079,28 +1089,37 @@ function calls rather than for references to function addresses. These
 relocations do not use the special symbol value described
 above. Otherwise a very tight endless loop would result.
 
-In a position independent executable all function addresses have
-dynamic relocations so no indirection via the PLT entry is necessary.
+In a position-independent executable all non-local function addresses
+are accessed via the GOT so no indirection via the PLT entry is
+necessary.
 
 Shared objects may implement direct function calls to non-pre-emptable
-symbols using the BL instruction. A symbol is not pre-emptable if:
+symbols using a direct branch instruction. A static linker will
+consider a symbol not pre-emptable if:
 
- * It has ``STB_LOCAL`` binding.
+ * The symbol has ``STB_LOCAL`` binding.
 
- * It has a symbol visibility that is not ``STV_DEFAULT``.
+ * The symbol has a symbol visibility that is not ``STV_DEFAULT``.
 
- * It has been given the ``local`` symbol version, defined in
+ * The symbol has been given the ``local`` symbol version, defined in
    SYM-VER_.
 
-All function calls to pre-emptable symbols must be indirected through
-the static linker created Procedure Linkage Table.
+ * The static linker adds the dynamic tag ``DT_FLAGS`` with the
+   ``DF_SYMBOLIC`` flag set. This makes all symbols non-pre-emptable.
+
+ * Some other implementation defined linker feature such as
+   ``--dynamic-list`` is used.
+
+Shared objects must indirect all function calls to pre-emptable
+symbols must be indirected through the static linker created Procedure
+Linkage Table.
 
 Procedure Linkage Table
 -----------------------
 
-Much as the global offset table (GOT) redirects position independent
+Much as the global offset table (GOT) redirects position-independent
 address calculations to absolute locations, the Procedure Linkage
-Table (PLT) redirects position independent function calls to absolute
+Table (PLT) redirects position-independent function calls to absolute
 locations. The static linker cannot resolve execution transfers (such
 as function calls) from one executable or shared object to
 another. Consequently, the static linker arranges to have the program
@@ -1109,18 +1128,18 @@ PLTs reside in the shared text segment, but they in turn use addresses
 in the per-process global offset table section ``.got.plt``. The
 dynamic linker determines the destinations' absolute addresses and
 modifies the global offset table entry accordingly. The dynamic linker
-can thus redirect the entries without compromising the position
-independence and shareability of the program's text. Executable files
-and shared objects have their own separate procedure linkage
-tables. The same procedure linkage table format is used for both
-executables and shared objects.
+can thus redirect the entries without compromising the
+position-independence and shareability of the program's
+text. Executable files and shared objects have their own separate
+procedure linkage tables. The same procedure linkage table format is
+used for both executables and shared objects.
 
 Following the steps below, the dynamic linker and the program
 cooperate to resolve symbolic references to functions through the
 procedure linkage table and the global offset table.
 
  * When first creating the memory image of the program, the dynamic
-   linker sets the second and the third entries in the global offset
+   linker sets the second and the third entries in the ``.got.plt``
    table to special values. The steps below explain more about these
    values.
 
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
-Program Linking and Dynamic Linking
+Program Loading and Dynamic Linking
 ===================================
 
-Not available for this Alpha release.
+The system works on creating a program/process image by logically
+copying or mapping the program image from storage into a virtual
+memory segment. In most systems with demand paging, when the actual
+page is mapped into physical memory depends on the execution behaviour
+of the program and the working set that it requires for it's correct
```
acoplan-arm, line 933: Very minor nit, but I think this should be "its correct execution" rather than "it's correct execution".

smithp35: Thanks, at least it is easy to fix!

mmalcomson (commented):
This looks pretty awesome to have in!

Most of the comments I've made are just because it read wrong to me when I was going through it first time -- so I suggested an alternate wording that I think is easier to understand.  Feel free to ignore them if you disagree.

There are a few more substantial questions.  I wouldn't be surprised if I were wrong in those.  But I figure it's better to ask just in case.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+64KiB pages. There are a number of optimizations that can be
+applied to save space.
+
+ * The first page of text contains the ELF header, program header
+   information and other information from the ELF file.
+
+ * The last page of text contains parts of the first page of data.
+
+ * The first data page has a copy of the last page of text.
+
```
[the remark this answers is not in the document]
smithp35: This bit always makes my head hurt!

The ELF file needs to satisfy p_vaddr is congruent to `p_offset` (modulo `p_align`). The linker can round both the `p_offset` and `p_vaddr` up to the next `p_align` boundary, putting zeroes in the file or it can add `p_align` bytes to `p_addr` with no padding needed in the file.
  
The best explanation I've read is the commit message for ld.bfd PR28824  https://sourceware.org/bugzilla/show_bug.cgi?id=28824#c3

In summary:
There will be a minimum of 2 pages required, one for the RO and one for the RW. In many cases mapping the same part of the file twice results in a smaller file, which could be significant for a really high page size such as 1 MiB on X86_64. In some cases where the RO almost fills up a full page there isn't much saving to be made by mapping the same page of the file twice. The `DATA_SEGMENT_ALIGN` linker script command has a calculation that pads the end of the RO in the file.

There is then the difference between max page size and common page size, that also makes my head hurt. The best explanation of this is in https://reviews.llvm.org/D33630?id=102280#767301

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
-Not available for this Alpha release.
+The system works on creating a program/process image by logically
+copying or mapping the program image from storage into a virtual
+memory segment. In most systems with demand paging, when the actual
+page is mapped into physical memory depends on the execution behaviour
+of the program and the working set that it requires for its correct
+execution. To minimize the number of such reads from the physical
+medium containing the program, executables and shared objects must
+have program segments whose file offsets and virtual addresses are
+congruent modulo the page size.
```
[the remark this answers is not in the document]
smithp35: In ELF the relationship is `p_vaddr` is congruent to `p_offset` (modulo `p_align`). It gets a bit complicated with the relationship between `p_align` and the page size as we have max-page-size which is set to the maximum possible supported page size (64 KiB) and common-page-size which is often set to (4 KiB) I had originally wrote that `p_align` should be set to the max-page-size but it turns out (https://sourceware.org/bugzilla/show_bug.cgi?id=28689) that this is overly strict as some dynamic linkers can adjust based on what platform they are running on to avoid overalignment.

I'll try and find a better way of wording it.

I'll clarify the wording. A loadable segment is used in http://www.sco.com/developers/gabi/latest/ch5.pheader.html to refer to a segment described by a program header with type `PT_LOAD` 
Checking ELF the requirements for non `PT_LOAD` program headers are stricter, they should be `p_vaddr` = `p_offset` (modulo `p_align`) although typically the `p_align` is only 4 or 8 for non loadable segmentes.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+the other for mapping as data.
+
+Shared objects are always built with a base virtual address of 0,
+while executables with an ELF ``e_type`` of ``ET_EXEC`` are built with
+a fixed virtual address. The underlying operating system maps shared
+objects in at different virtual addresses, utilizing the fact that the
+code and data addressing within such a shared object is position
+independent. The position independence relies on the principle that
+the distance between any two points within the image is identical at
+static link time and at run time.
```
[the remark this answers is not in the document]
smithp35: I'll see if I can word it better. The static linker hard-codes the PC-relative offsets from the Code segment to the Data Segment(s) so a dynamic linker can't change the offsets at runtime, even if the static linker can calculate them.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+Both the ``.got`` and ``.got.plt`` sections are aligned to a minimum
+of 8 bytes. They both contain 8-byte sized entries.
+
+If a program requires direct access to the absolute address of a
+symbol, that symbol will have a GOT entry. Because the executable and
+shared objects have their own independent GOTs, a symbol's address may
+appear in several tables. The dynamic linker processes all of the
+relocations in the ``.got`` section before giving control to any code
+in the process image, thus ensuring the absolute addresses are
+available during execution.
```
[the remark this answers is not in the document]
smithp35: I'll have a think to see if I can come up with anything.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+By convention the ``.got`` section's first entry (number zero) is
+reserved to hold the address of the dynamic structure, referenced with
+the linker defined symbol ``_DYNAMIC``. Some dynamic linker's use this
```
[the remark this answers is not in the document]
smithp35: I think it isn't required now, but it would be for earlier versions of glibc. I wanted to find some words to express that, but not sure if I've succeeded.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+Direct function calls are those where the name of the called function
+is known at compile time. The PC-relative BL instruction may be used
+for all direct function calls, whether absolute or position
+independent.
+
+Indirect function calls are those where the address of the function is
+in a pointer. Appropriate code is used to load the value of the
+pointer into a register, as for other data, and then a BLR instruction
+is used.
```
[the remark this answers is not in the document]
smithp35: I think I need to take out the BL, and BLR as a direct function call can be made with a B (tail call) and as you say BR can be used for indirect.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+function address defined in a shared object will normally be resolved
+by the static linker to the address of the `Procedure Linkage Table`_
+(PLT) for that function within the executable file. If the address is
+in a writeable location the linker may use a dynamic relocation
+instead.
+
+When both shared objects and a non position independent executable use
+the address of a function both the shared library and the application
+must use the same address for the functions for address comparisons to
+work as expected. If the static linker uses a PLT entry for a function
```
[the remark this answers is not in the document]
smithp35: Thanks for the suggestion, sounds better than what is there already.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+as the address for a function then it must place the address of the
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+PLT for the function in the dynamic symbols table entry for the
+function. This will result in a dynamic symbol table entry with a
+section index of ``SHN_UNDEF``, a type ``STT_FUNC`` and a non-zero
+``st_value``. A reference to this symbol will be resolved by the
+dynamic linker to the address PLT for the function in the executable.
```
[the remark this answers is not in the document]
smithp35: Ack

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+The ``R_<CLS>_JUMP_SLOT`` relocations defined in AAELF64_ are
+associated with PLT entries. These entries are used for direct
+function calls rather than for references to function addresses. These
+relocations do not use the special symbol value described
+above. Otherwise a very tight endless loop would result.
+
+In a position independent executable all function addresses have
+dynamic relocations so no indirection via the PLT entry is necessary.
```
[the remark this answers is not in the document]
smithp35: I'll see what I can do.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * For illustration, assume the program calls ``name1``, which transfers
+   control to the Nth PLT entry in ``PLT[N]``.
+
+ * The ``PLT[N]`` entry loads the address of the GOT entry for
+   ``name1`` (``.got.plt[N]``) into register ``ip0``, and then loads
+   and jumps to the address held in that table entry using a ``BR``
+   instruction. Initially the GOT entry holds the address of the
+   ``PLT[0] entry``, the first entry in the PLT, not the real address
+   of ``name1``.
```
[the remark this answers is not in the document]
smithp35: There isn't a requirement, more of a convention as every entry in the PLT needs one `.got.plt` entry so they are allocated together in the implementation. It doesn't make a difference at runtime although some disassembly tools do assume that this convention holds when pretty printing the PLT.

I think I need to clarify, especially as I've forgot to be clear about the reserved entries at the front of the `.got.plt` that are not associated with a `.plt` entry.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * Now the ``PLT[0]`` entry pushes the ``ip0`` and ``lr`` registers
+   onto the stack.  The dynamic linker will later use the stacked
+   value of ``ip0`` to compute a relocation index by subtracting the
+   base of the ``GOT`` The index will be used to select an entry from
+   the table of relocation entries addressed by the ``DT_JMPREL``
+   dynamic section entry. The designated relocation entry must have
```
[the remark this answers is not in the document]
smithp35: I'll need to check, the original text used ``GOT`` without making the distinction between `.got` and `.got.plt` so I could have missed this one. I'm pretty sure you're right and it will be ``PLTGOT``.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+   type ``R_AARCH64_JUMP_SLOT``, and its offset will specify the
+   global offset table entry used by the branch instruction in
+   ``PLT[N]``, and also a symbol table index that references the
+   appropriate symbol, i.e. ``name1`` in this example.
+
```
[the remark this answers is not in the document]
smithp35: Will have a go at rewording.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+ * The ``PLT[0]`` entry then loads the adress of the third GOT entry
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+   (``.got.plt[2]``) into ``ip0``, and the loads and jumps to the
+   address in the third table entry, which transfers control to the
+   dynamic linker.
+
+ * The value in ``ip0`` is used by the dynamic linker's entrypoint to
+   load the second GOT entry (``.got.plt[2]``) using ``[ip0,#-8]``,
+   which gives it one word of private information with which to
+   identify the calling module.
```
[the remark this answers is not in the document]
smithp35: Thanks, I'll need to triple check my indexes and slots to make sure they are consistent.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+   will transfer to ``name1``, instead of ``PLT[0]``.
+
+The ``LD_BIND_NOW`` environment variable can change the dynamic linking
+behavior. If its value is non-null, the dynamic linker evaluates all
+procedure linkage table entries before transferring control to the
+program. That is, the dynamic linker processes relocation entries of
+type ``R_AARCH64_JUMP_SLOT`` during process initialization. Otherwise, the
+dynamic linker evaluates procedure linkage table entries lazily,
+delaying symbol resolution and relocation until the first call to a
+table entry.
```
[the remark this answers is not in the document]
smithp35: I think I should just say all relocations in the `.got.plt` and not be specific about their type.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
++---------------+-------------------------+
+| ``[sp, #8]``  | ``lr``                  |
++---------------+-------------------------+
+
+Sample PLT sequences
+^^^^^^^^^^^^^^^^^^^^
+
+The following section shows some examples of PLT sequences suitable
+for the small code model. There is an assumption that the ``.plt.got``
+section is within +- 4 GiB from the ``.plt`` section.
```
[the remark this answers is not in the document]
smithp35: Thanks for the suggestion, I think it can be used without making the text too complicated.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+   expression expr.
+
+ - ``GOTPLT`` is the address of ``.got.plt`` section.
+
+ - ``.got.plt[N]`` is the Nth entry of the ``.got.plt`` section.
+
+In combination:
+
+ - ``G(.plt.got[N]`` is the address of the Nth entry of the
+   ``.got.plt`` section.
```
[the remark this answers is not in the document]
smithp35: Ack. It should be `.got.plt` will fix that.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ - ``G(.plt.got[N]`` - ``GOTPLT`` is the offset from the base of the
+   .got.plt section to the entry in ``.plt.got[N]``.
```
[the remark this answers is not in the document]
smithp35: Ack

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+    br     x17
+
+PLT entry ``.got.plt[N]``
+
+.. code-block:: asm
+
+    adrp x16, Page(G(.plt.got[N]))
+    ldr  x17, [x16, G(.plt.got[N]) - GOTPLT]
+    add  x16, x16, G(.plt.got[N]) - GOTPLT
+    br   x17
```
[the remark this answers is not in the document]
smithp35: I'll double check these as I think I've forgotten about the reserved entries so I need to be more precise here.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+PAC
+~~~
+
+The ``.got.plt`` entries can be signed by the dynamic linker and
+authenticated by the code in the PLT entry. Note that when these PLT
+sequences are used the dynamic linker must sign the entries in
+``.plt.got[N]``. On platforms that support `Relocation Read Only
```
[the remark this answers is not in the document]
smithp35: Ack

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+ * IFUNC resolvers must not call a function that requires a PLT
+   entry. If the address of the IFUNC is taken the dynamic relocation
+   will be in the ``.got`` so the resolver may be called prior to
+   relocation of the ``.got.plt`` entries. In a shared library or
+   position independent executable, transferring to control to the
+   address stored in an unrelocated ``.got.plt`` entry will result in
+   undefined behavior.
+
+ * IFUNC resolvers for a given function must be defined in the same
+   translation unit as the implementations of the function.
```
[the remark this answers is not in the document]
smithp35: I can put it in the recommendations for maximum portability. I've separated out recommendations from requirements as some platforms like freebsd have much more complex ifunc ordering logic that doesn't suffer from many of these problems at the expense of a slower startup.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+IFUNC requirements for static linkers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Relocations to pre-emptable symbols of type ``STT_GNU_IFUNC`` are
+handled in the same way as symbols of type ``STT_FUNC``.
+
+Non pre-empatable symbols of type ``STT_GNU_FUNC`` require the
+creation of PLT entry that loads the value of:
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * A ``.got`` entry for a non-branch relocation such as
+   ``R_AARCH64_ABS64``. The entry has an ``R_AARCH64_IRELATIVE``
+   dynamic relocation in ``.rela.dyn``.
+
```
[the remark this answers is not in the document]
smithp35: I'll try and clarify. I think that there needs to be a `.got` with an associated `.rela.dyn` entry when the address of an ifunc is taken, but that doesn't prevent a PLT entry and `.got.plt` entry from being created as well for a direct call. I expect that this means that dynamic loaders must make sure that it only attempts to resolve one of the relocations for each address.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+ * Either a ``.got`` or ``.got.plt`` entry or a for a branch
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+does not require a symbol.
+
+The static linker sets the initial contents of a ``.got`` or
+``.got.plt`` entry for a non pre-empatable IFUNC resolver to the
+address of the IFUNC resolver function.
+
+To make address equivalence of functions with IFUNC resolvers work, if
+the address of a non-preemptable ``STT_GNU_FUNC`` symbol is taken in a
+non-position independent executable. The static linker must use the
+address of the PLT for the address of the symbol. If the symbol is
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+Initialization and Termination Functions
+----------------------------------------
+
+The implementation is responsible for executing the initialization
+functions specified by ``DT_INIT`` , ``DT_INIT_ARRAY`` , and
+``DT_PREINIT_ARRAY`` entries in the executable file and shared object
+files for a process, and the termination (or finalization) functions
+specified by ``DT_FINI`` and ``DT_FINI_ARRAY``. The user program plays
+no further part in executing the initialization and termination
+functions specified by these dynamic tags
```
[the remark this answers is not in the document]
smithp35: I'll clarify. It means that the user program does not directly call the functions, the dynamic loader is responsible for doing that.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
MaskRay: glibc dropped the `.got[0] == _DYNAMIC` requirement as a resolution to https://sourceware.org/bugzilla/show_bug.cgi?id=28203

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * ``.got.plt`` for code addresses accessed only from the Procedure Linkage
+   Table (PLT).
+
+ * ``.got`` all other addresses and offsets.
+
+Both the ``.got`` and ``.got.plt`` sections are aligned to a minimum
+of 8 bytes. They both contain 8-byte sized entries.
+
+If a program requires direct access to the absolute address of a
```
MaskRay, line 1014: > If a program requires direct access to the absolute address of a symbol, that symbol will have a GOT entry.

This sentence is correct for a PIE or a shared object, but not for a PDE. The context is probably not clear enough that this does not apply to PDE.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+In statically linked code the address of a function is always its real
+address. However in dynamically linked environments references to the
+address of a function from an executable file and its shared objects
+might not resolve to the same value.
+
+References from within shared objects to a function address will
+normally be resolved by the dynamic linker to the virtual address of
+the function itself.
+
+References from within a non position independent executable file to a
```
MaskRay, line 1058: ISTM `non` is always used as `non-foo` or `nonfoo`, but not `non foo`.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
MaskRay: > all function addresses have dynamic relocations

When the function address is local, PC-relative instructions can be used to avoid dynamic relocations.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+The ``R_<CLS>_JUMP_SLOT`` relocations defined in AAELF64_ are
+associated with PLT entries. These entries are used for direct
+function calls rather than for references to function addresses. These
+relocations do not use the special symbol value described
+above. Otherwise a very tight endless loop would result.
+
+In a position independent executable all function addresses have
+dynamic relocations so no indirection via the PLT entry is necessary.
+
+Shared objects may implement direct function calls to non-pre-emptable
```
MaskRay, line 1085: Calls to preemptable symbols use the BL instruction as well, even if going through PLT indirection. The sentence needs to clarify the BL instruction use.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+symbols using the BL instruction. A symbol is not pre-emptable if:
+
+ * It has ``STB_LOCAL`` binding.
+
+ * It has a symbol visibility that is not ``STV_DEFAULT``.
+
+ * It has been given the ``local`` symbol version, defined in
+   SYM-VER_.
+
```
MaskRay, line 1094: Does the list need to comprehensive? If it does, it omits some linker extensions (`-Bsymbolic` familiy options and `--dynamic-list`).

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+and shared objects have their own separate procedure linkage
+tables. The same procedure linkage table format is used for both
+executables and shared objects.
+
+Following the steps below, the dynamic linker and the program
+cooperate to resolve symbolic references to functions through the
+procedure linkage table and the global offset table.
+
+ * When first creating the memory image of the program, the dynamic
+   linker sets the second and the third entries in the global offset
```
MaskRay, line 1123: Here the "global offset table" refers to `.got.plt`.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * The dynamic linker unwinds the stack, looks at the designated
+   relocation entry, finds the symbol's value, stores the actual
+   address of ``name1`` in ``.got.plt[N]`` , then transfers control to
+   the desired destination.  Subsequent executions of ``PLT[N]`` will
+   transfer control directly to ``name1``, without calling the dynamic
+   linker a second time. That is the ``BR`` instruction in ``PLT[N]``
+   will transfer to ``name1``, instead of ``PLT[0]``.
+
+The ``LD_BIND_NOW`` environment variable can change the dynamic linking
```
MaskRay, line 1170: And `DT_BIND_NOW`.

Full RELRO has been enabled for all packages in Fedora 23 and later.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
MaskRay: .plt.got => .got.plt

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
MaskRay: `G` indicates an address. The ldr immediate is a page offset, so the `G(...)` use in `ldr` may not be that appropriate.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+GNU Indirect Functions
+----------------------
+
+A GNU Indirect Function (IFUNC) is a feature that permits a single
+implementation of a function to be chosen from multiple candidates,
+with the choice taken by an IFUNC resolver function.
+
+GNU Indirect Functions require static and dynamic linker support. They
+are known to be supported on GNU/Linux, Android and many of the BSD
```
MaskRay, line 1346: Add comma: `Android, and many ...`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+  /* Implementation alternative 2 */
+  static int implementation2(void) { ... }
+
+  /* Resolver function */
+  static void* resolver(uint64_t, const __ifunc_arg_t *) { ... }
+
+  /* make symbol ifunc type STT_GNU_IFUNC using resolver as the IFUNC resolver. */
+  int ifunc(void) __attribute__((ifunc("resolver")));
+
+The first ``uint64`` parameter is ``AT_HWCAP``. If the
```
MaskRay, line 1402: uint64 => uint64_t

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+``_IFUNC_ARG_HWCAP`` flag is set the second parameter is present. The
+second parameter is a structure ``__ifunc_arg_t`` defined in
+``sys/ifunc.h`` it provides access to all of the flags in HWCAP_ via
```
MaskRay, line 1405: Omitted punctuation before `it provides ...`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+additional restrictions on what they can contain.
+
+ * IFUNC resolvers must not be compiled with security features like
+   stack-protection, which requires a guard variable to be
+   initialized. Or instrumentation like ASAN that requires a shadow
+   map to be set up.
+
+ * IFUNC resolvers must not have a symbol binding of ``STB_WEAK``.
+
+ * IFUNC resolvers that may be called via lazy-binding must be
```
MaskRay, line 1421: > IFUNC resolvers that may be called via lazy-binding must be asynchronous-signal safe and thread-safe as the resolver may be called from a signal handler or multi-threaded process.

The "async-signal safe and thread-safe" part on https://sourceware.org/glibc/wiki/GNU_IFUNC does not match the implementation. IFUNC resolvers are eagerly called even in the lazy binding PLT mode. Perhaps drop the description here.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+
+To make address equivalence of functions with IFUNC resolvers work, if
+the address of a non-preemptable ``STT_GNU_FUNC`` symbol is taken in a
+non-position independent executable. The static linker must use the
+address of the PLT for the address of the symbol. If the symbol is
+exported to the dynamic symbol table the ``st_value`` of the symbol
+must be set to the address of the PLT entry and the dynamic symbol
+table type must be set to ``STT_FUNC``.
+
+The ``.rela.dyn`` and ``rela.plt`` relocation entries for IFUNCs must
```
MaskRay, line 1479: typo in `rela.plt`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+be sorted after all non-IFUNCs. This means that for a given executable
+or shared-library the following ordering can be relied on:
+
+ * All non IFUNC ``.rela.dyn`` relocation entries will be resolved
```
MaskRay, line 1483: The order matches the glibc behavior, but FreeBSD rtld does not use this order. I think we probably should not make FreeBSD's implementation non-conforming.

See https://maskray.me/blog/2021-01-18-gnu-indirect-function#relocation-resolving-order for some notes about FreeBSD rtld's behavior.

smithp35:
Thank you very much for the reviews. I'm attempting to get a revised draft out by the end of the week.

sysvabi64/sysvabi64.rst
smithp35: I've tried to word it as recommendations for maximum portability, rather than a requirement. For example a program written against FreeBSD `rtld` may not run if loaded by `ld.so`. That doesn't make `rtld` non compliant though.

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: Thanks, I'll drop it. I'd thought that a `R_AARCH64_JUMP_SLOT` reloc would not be eagerly evaluated but it looks like, in glibc at least, that if there is a Relocation with respect to a symbol of type STT_GNU_IFUNC the resolver is called prior to processing the relocation.

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: Ack, will add.

sysvabi64/sysvabi64.rst
smithp35: Ack, will clarify.

sysvabi64/sysvabi64.rst
smithp35: I'll add `-BSymbolic` via the ELF flag `DF_SYMBOLIC` as that is part of ELF. I'll add another entry for miscellaneous static linker options.

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: I'd probably need to make it `non-position-dependent` rather than `non-position dependent`, will need to update all `position independent` to make it `position-independent`. Sadly the ABI as a whole is a little bit inconsitent in which form it has used.

sysvabi64/sysvabi64.rst
smithp35: Will try and clarify. The opening paragraph is about position independence, but position dependent executables do have GOT entries for imported symbols and TLS.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,639 @@ These follow the AAELF64_ base definition from Arm.
...
+to these regions *as if* each segment were complete and
+separate. Typically the regions of the file holding the last page of
+the text segment and the first page of the data segment will be mapped
+twice at two different virtual addresses, one for mapping as code and
+the other for mapping as data.
+
+Shared objects are always built with a base virtual address of 0,
+while executables with an ELF ``e_type`` of ``ET_EXEC`` are built with
+a fixed virtual address. The underlying operating system maps shared
+objects in at different virtual addresses, utilizing the fact that the
```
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
smithp35: Turns out that on AArch64 the `R_AARCH64_IRELATIVE` uses the relocation addend for the ifunc resolver address so it is safe to call the resolver multiple times from `R_AARCH64_IRELATIVE`, assuming the resolution code isn't stateful.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
smithp35: I think I can fix this by adding a `LO12` operator so it becomes `LO12(G(.plt.got[N]) - GOTPLT)`.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
smithp35: Ack. Thanks.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
smithp35: Ack.

sysvabi64/sysvabi64.rst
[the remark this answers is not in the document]
Wilco1: The 4GB limit applies to **all** code models, so it's best not to mention `suitable for the small code model` because the sequences apply to all code models.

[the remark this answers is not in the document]
smithp35: OK. Will remove that part.

sysvabi64/sysvabi64.rst
MaskRay: LGTM

smithp35:
Made an attempt to address the review comments. Quite a lot of edits, so still likely I've made some mistakes somewhere.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,657 @@ These follow the AAELF64_ base definition from Arm.
...
+ * The static linker adds the dynamic tag ``DT_FLAGS`` with the
+   ``DF_SYMBOLIC`` flag set. This makes all symbols non-pre-emptable.
+
+ * Some other implementation defined linker feature such as
+   ``--dynamic-list`` is used.
+
+Shared objects must indirect all function calls to pre-emptable
+symbols must be indirected through the static linker created Procedure
+Linkage Table.
+
```
mmalcomson, line 1116: Phrasing -- "must indirect all function calls to pre-emptable symbols must be indirected".

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,657 @@ These follow the AAELF64_ base definition from Arm.
...
+PLT header ``PLT[0]``
+
+.. code-block:: asm
+
+    stp    x16, x30, [sp,#-16]!
+    adrp   x16, :page: &.got.plt[2]
+    ldr    x17, [x16, (&.got.plt[2] - &.got.plt[0])]
+    add    x16, x16, (&got.plt[2] - &got.plt[0])
+    br     x17
```
mmalcomson, line 1251: I think typo -- I believe `&.got.plt[2] - &.got.plt[0]` should be `:lo12: &.got.plt[2]` (same in the Nth entry below).

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,657 @@ These follow the AAELF64_ base definition from Arm.
...
+functions specified by ``DT_INIT`` , ``DT_INIT_ARRAY`` , and
+``DT_PREINIT_ARRAY`` entries in the executable file and shared object
+files for a process, and the termination (or finalization) functions
+specified by ``DT_FINI`` and ``DT_FINI_ARRAY``.
+
+Relocation Read Only (RELRO)
+----------------------------
+
+Several sections in an exectuable or shared-library are logically
+read-only but they require dynamic relocation which forces them to be
```
mmalcomson, line 1543: Typo `exectuable` -> `executable`.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   dynamic relocation in ``.rela.dyn``.
+
+ * A ``.got.plt`` entry for a branch relocation such as
+   ``R_AARCH64_CALL``. A ``R_AARCH64_IRELATIVE`` dynamic relocation is
+   added to ``.rela.dyn`` or ``rela.plt``.
+
+Due to the ordering requirements on IFUNC resolvers the PLT entry and
+associated ``.got.plt`` entry are by often implemented in a separate
+``.iplt`` and ``.iplt.got`` sections so that they placed after the
+``.plt`` and ``.got.plt`` sections respectively.
```
mmalcomson, line 1479: typo  -- `by often` -> `often`
            `so that they placed` -> `so that they are placed`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+function. If the symbol is exported to the dynamic symbol table the
+``st_value`` of the symbol must be set to the address of the PLT entry
+and the dynamic symbol table type must be set to ``STT_FUNC``.
+
+Relocations of type ``R_AARCH64_IRELATIVE`` relocation must be sorted
+after all other relocation types. This means that for a given
+executable or shared-library the following ordering can be relied on:
+
+ * All non ``R_AARCH64_IRELATIVE`` relocations in ``.rela.dyn`` will be resolved
+   before any ``R_AARCH_IRELATIVE`` relocations.
```
mmalcomson, line 1499: typo -- `R_AARCH_IRELATIVE`  -> `R_AARCH64_IRELATIVE`.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+The order of dynamic relocation resolution across an executable and
+all its shared libraries is platform specific. The following
+recommendations for writing IFUNC resolvers apply to the GNU glibc
+dynamic loader. Other dynamic linkers may have fewer requirements.
+
+ * IFUNC resolver functions must not call a function that may require
+   IFUNC initialization. If the IFUNC initialization for the called
+   function has not occured then undefined behavior results.
+
+ * In position-independent code an IFUNC resolver functions must not
```
mmalcomson, line 1444: Tenses `an` IFUNC resolver `functions`   --  I think should either be `an` & `function` or just `functions` without the `an`.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   call a function that requires a PLT entry. If the IFUNC resolver
+   runs as a result of a relocation in ``.rela.dyn`` then the
+   relocations in ``.rela.plt`` will not have been resolved. This means
+   that addresses in the ``.got.plt`` will be unchanged from their
+   static link time value.
```
mmalcomson, line 1449: I'm a little confused about this.
My thought is that relocations in `.rela.plt` have a lazy mechanism and so why should this matter if the relocations are not yet resolved.
Not sure if that makes sense in the case of `BIND_NOW` -- not sure if there are any other assumptions in that case -- is that the reason why this matters?
Or is it something to do with some other setup of the PLT mechanism in the dynamic linker?

sysvabi64/sysvabi64.rst
smithp35: Ack have fixed.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,657 @@ These follow the AAELF64_ base definition from Arm.
...
+ * Now the ``PLT[0]`` entry pushes the ``ip0`` and ``lr`` registers
+   onto the stack.  The dynamic linker will later use the stacked
+   value of ``ip0`` to compute a relocation index by subtracting the
+   base of the ``.got.plt`` The index will be used to select an entry
+   from the table of relocation entries addressed by the ``DT_JMPREL``
+   dynamic section entry. The designated relocation entry must have
+   type ``R_AARCH64_JUMP_SLOT``, and its ``r_offset`` field will
+   specify the ``.got.plt`` table entry used by the PLT entry in
+   ``PLT[N]``, and also a symbol table index that references the
+   appropriate symbol, i.e. ``name1`` in this example.
```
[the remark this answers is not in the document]
smithp35: Ack have fixed.

sysvabi64/sysvabi64.rst
smithp35: Thanks will fix.

sysvabi64/sysvabi64.rst
smithp35: Thanks; I've used An IFUNC resolver function.

sysvabi64/sysvabi64.rst
smithp35: In a position independent executable or shared library all the addresses in the .got.plt need to have the load bias added to them at initialisation time. This needs to be done even when lazy resolving as the address of the PLT entry for the lazy resolver in the .got.plt needs to be adjusted by the load bias.

FWIW this confused me too until I tried to write a program to see what happened.

smithp35:
I think all the comments so far have been resolved. From my side I think that this is close to being merged. Will leave open for another week for further comments, otherwise I'll ask for this to be merged.

sysvabi64/sysvabi64.rst
```diff
@@ -191,6 +193,9 @@ Change History
  +============+==============================+=======================================================+
  | 00Alpha    | 1\ :sup:`st` November 2021   | Alpha release containing Code Model only              |
  +------------+------------------------------+-------------------------------------------------------+
+ | 01Alpha    | 14\ :sup:`th` July 2022      | Add Program Loading and Dynamic Linking section       |
```
MaskRay, line 196: Update the date?

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+independent GOTs, a symbol's address may appear in several tables. The
+dynamic linker processes all of the relocations in the ``.got``
+section before giving control to any code in the process image, thus
+ensuring the absolute addresses are available during execution. Code
+in the process that runs as part of a GNU indirect function resolver
+is an exception to this rule. See `GNU Indirect Functions`_ for the
+restrictions on what addresses can be accessed from a resolver
+function.
+
+By convention the ``.got`` section's first entry (number zero) is
```
MaskRay, line 1028: The  working about the first entry can probably be non-nominal (i.e. not required). lld does not set the entry to `_DYNAMIC`.

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+reserved to hold the address of the dynamic structure, referenced with
+the linker defined symbol ``_DYNAMIC``. Versions of glibc prior to
+version 2.35 use this address to find its own dynamic structure
+without without having yet processed their own relocation entries.
```
MaskRay, line 1032: remove one `without`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+
+References from within shared objects to a function address will
+normally be resolved by the dynamic linker to the virtual address of
+the function itself.
+
+References from within a non-position-independent executable file to a
+function address defined in a shared object will normally be resolved
+by the static linker to the address of the `Procedure Linkage Table`_
+(PLT) entry for that function within the executable file. If the
+address is in a writeable location the linker may use a dynamic
```
MaskRay, line 1071: s/writeable/writable/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   value of ``ip0`` to compute a relocation index by subtracting the
+   base of the ``.got.plt`` The index will be used to select an entry
+   from the table of relocation entries addressed by the ``DT_JMPREL``
+   dynamic section entry. The designated relocation entry must have
+   type ``R_AARCH64_JUMP_SLOT``, and its ``r_offset`` field will
+   specify the ``.got.plt`` table entry used by the PLT entry in
+   ``PLT[N]``, the ``r_info`` encodes a symbol table index that
+   references the appropriate symbol, i.e. ``name1`` in this example.
+
+ * The ``PLT[0]`` entry then loads the adress of the third
```
MaskRay, line 1173: s/adress/address/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+security features enabled.
+
+PLT header ``PLT[0]``
+
+.. code-block:: asm
+
+    stp    x16, x30, [sp,#-16]!
+    adrp   x16, :page: &.got.plt[2]
+    ldr    x17, [x16, :lo12: &.got.plt[2]]
+    add    x16, x16, :lo12: &got.plt[2]
```
MaskRay, line 1249: s/got.plt/.got.plt/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+
+PLT header ``PLT[0]``
+
+.. code-block:: asm
+
+    bti  c
+    stp  x16, x30, [sp,#-16]!
+    adrp   x16, :page: &.got.plt[2]
+    ldr    x17, [x16, :lo12: &.got.plt[2]]
+    add    x16, x16, :lo12: &got.plt[2]
```
MaskRay, line 1280: s/got.plt/.got.plt/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+  static int implementation2(void) { ... }
+
+  /* Resolver function */
+  static void* resolver(uint64_t, const __ifunc_arg_t *) { ... }
+
+  /* make symbol ifunc type STT_GNU_IFUNC using resolver as the IFUNC resolver. */
+  int ifunc(void) __attribute__((ifunc("resolver")));
+
+The first ``uint64_t`` parameter is ``AT_HWCAP``. If the
+``_IFUNC_ARG_HWCAP`` flag is set the second parameter is present. The
```
MaskRay, line 1417: `is set, the second ...`

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+ * IFUNC resolvers must not have a symbol binding of ``STB_WEAK``.
+
+The order of dynamic relocation resolution across an executable and
+all its shared libraries is platform specific. The following
+recommendations for writing IFUNC resolvers apply to the GNU glibc
+dynamic loader. Other dynamic linkers may have fewer requirements.
+
+ * An IFUNC resolver function must not call a function that may itself
+   require IFUNC initialization. If the IFUNC initialization for the
+   called function has not occured then undefined behavior results.
```
MaskRay, line 1442: s/occured/occurred/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+Several sections in an executable or shared-library are logically
+read-only but they require dynamic relocation which forces them to be
+writeable. Relocation Read Only (RELRO) is a GNU extension to ELF that
+permits a dynamic linker to remap the pages described by the RELRO
+program segment as read-only after relocation.
+
+The RELRO program segment is described by a program header with type
+``PT_GNU_RELRO``, defined in LSB_. The number of supported RELRO
+segments is a contract between the static and dynamic linker. The GNU
+C library only supports one RELRO segement per ELF file.
```
MaskRay, line 1546: s/segement/segment/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+
+ * The linker created ``.got.plt`` section if the dynamic section
+   ``DT_FLAGS`` table entry contains ``DF_BIND_NOW``.
+
+ * The section in the output file has a name that matches one of
+   ``.data.rel.ro``, ``.bss.rel.ro``, ``.ctors``, ``.dtors``,
+   ``.jcr``, ``.eh_frame``, ``.fini_array``, ``.init_array``,
+   ``.preinit_array``.
+
+The size of the RELRO segment should be extended so that is a multiple of the page size.
```
MaskRay, line 1571: so that it is

MaskRay (approved):

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+  /* Implementation alternative 1 */
+  static int implementation1(void) { ... }
+
+  /* Implementation alternative 2 */
+  static int implementation2(void) { ... }
+
+  /* Resolver function */
+  static void* resolver(uint64_t, const __ifunc_arg_t *) { ... }
+
+  /* make symbol ifunc type STT_GNU_IFUNC using resolver as the IFUNC resolver. */
```
MaskRay, line 1413: s/make/Make/

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   ``R_AARCH64_CALL``. A ``R_AARCH64_IRELATIVE`` dynamic relocation is
+   added to ``.rela.dyn`` or ``rela.plt``.
+
+Due to the ordering requirements on IFUNC resolvers the PLT entry and
+associated ``.got.plt`` entry are often implemented in a separate
+``.iplt`` and ``.iplt.got`` sections so that they are placed after the
+``.plt`` and ``.got.plt`` sections respectively.
+
+Like ``R_AARCH64_RELATIVE`` the ``R_AARCH64_IRELATIVE`` relocation
+does not require a symbol. For ``RELA`` type relocations the addend of
```
MaskRay, line 1482: Q: Does `does not require` indicate the intent that the symbol index must be zero?

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   before any ``R_AARCH64_IRELATIVE`` relocations.
+
+ * ALL IFUNC resolvers with ``R_AARCH64_IRELATIVE`` relocations in
+   ``.rela.dyn`` will be run before the ``.rela.plt`` relocations are
+   resolved.
+
+ * All non ``R_AARCH64_IRELATIVE`` relocations in ``.rela.plt`` will be resolved
+   before any ``R_AARCH64_IRELATIVE`` relocations.
+
+ * All IFUNC resolvers with ``R_AARCH64_IRELATIVE`` relocations in
```
MaskRay, line 1508: glibc resolves `R_AARCH64_IRELATIVE` in `.rela.plt` eagerly. Does "will be run last" express the intent that IRELATIVE relocations still run before lazy binding JUMP_SLOT?

sysvabi64/sysvabi64.rst
```diff
@@ -916,10 +923,652 @@ These follow the AAELF64_ base definition from Arm.
...
+   ``.rela.plt`` will be run last.
+
+When static linking, the dynamic relocation section containing
+``R_AARCH64_IRELATIVE`` relocations are output as if dynamic
+linking. The static linker must define the symbols
```
MaskRay, line 1513: `__rela_iplt_start` is only defined for `-no-pie`, not for `-pie`.

See https://github.com/llvm/llvm-project/commit/f8cb78e99aae9aa3f89f7bfe667db2c5b767f21f (`[ELF] Don't define __rela_iplt_start for -pie/-shared`)

MaskRay (approved):

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: I've tried a rewrite that essentially says that if you want to link an older version of glibc then you'll need to reserve the first entry for `_DYNAMIC`.

sysvabi64/sysvabi64.rst
smithp35: Ack

sysvabi64/sysvabi64.rst
smithp35: Ack. I think writeable is an alternative spelling in British English, but writable is more appropriate here, I've also changed the other occurrences in the doc.

sysvabi64/sysvabi64.rst
smithp35: Ack

sysvabi64/sysvabi64.rst
smithp35: Ack

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: Ack

sysvabi64/sysvabi64.rst
smithp35: Ack

sysvabi64/sysvabi64.rst
smithp35: Ack.

sysvabi64/sysvabi64.rst
smithp35: It is meant to represent that the symbol index can be 0. I'll try and clarify. In practice I've not seen any with a non-0 index but the case is well defined in the ABI. Quotiing from  Delta(S) in https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5733relocation-operations .

> Delta(S) if S is a normal symbol, resolves to the difference between the static link address of S and the execution address of S. If S is the null symbol (ELF symbol index 0), resolves to the difference between the static link address of P and the execution address of P.

sysvabi64/sysvabi64.rst
smithp35: I was trying to capture that even when lazy binding is in use, the load bias (for position independent code) is added to the .got.plt entries. This means that in position independent code an ifunc resolver can call a lazy bound function and not crash.

Attempted to clean up the wording in a bit.

sysvabi64/sysvabi64.rst
smithp35: Thanks for the link. I'll update to account for that.

PR #161, opened 2022-07-15
Describe changes to VG in aadwarf64

rsandifo-arm:
SME adds the concept of “streaming mode”, which can be entered
using SMSTART and exited using SMSTOP.  The SME ABI requires
these changes in streaming mode to be naturally nested:
a function entered in streaming mode must return in streaming
mode, and a function entered in non-streaming mode must return
in non-streaming mode.

The vector length in streaming mode can be different from the vector
length in non-streaming mode.  As a consequence, it's possible for the
value of the DWARF VG pseudo register to change at an SMSTART and at
an SMSTOP.  Each such change in value could invalidate VG-based DWARF
expressions for functions further up the call stack.  For example,
if a function F1 spills an SVE register to the stack, the size of
its stack frame might depend on VG.  If F1 then calls a function F2
that does an SMSTART, the SMSTART could change the value of VG.
Trying to calculate the size of F1's stack frame using this new
VG could give the wrong result.

This patch describes a mechanism for dealing with this.  Although
SME was the motivating use case, it's in principle possible for
programs to change the SVE vector length independently of SME.
We don't yet have a defined ABI for doing that, but the mechanism
described here was always our back-up plan for handling this in DWARF,
if it ever became necessary.

### DIFF
--- a/aadwarf64/aadwarf64.rst
+++ b/aadwarf64/aadwarf64.rst
@@ -226,6 +226,9 @@ changes to the content of the document for that release.
   |        |                             |   document a limitation of             |
   |        |                             |   DW_CFA_AARCH64_negate_ra_state.      |
   +--------+-----------------------------+----------------------------------------+
+  | TBD    | TBD                         | - Added `Changes in vector length`_ at |
+  |        |                             |   **Alpha** quality.                   |
+  +--------+-----------------------------+----------------------------------------+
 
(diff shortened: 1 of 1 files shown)


sallyarmneale (commented):
Looks good to me

walkerkd (approved):

PR #165, opened 2022-08-15
add legacy documents to the repo

stuij:
Legacy documents are versions of the specs that were released before we did
releases here on Github. They have a proprietary license, and were released on
developer.arm.com. We're adding them in this repo so we have all versions in the
same place.


legacy-documents/aaelf32/ihi0044_H/IHI0044G_aaelf.html
```diff
@@ -0,0 +1,5669 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
statham-arm, line 1: This file has the wrong name! It's in the subdirectory for version `H` of the document, but its name has `G` in it.

The version number inside the file is `H`, so I think it's only the file name that's wrong.

statham-arm (commented):
As well as the specific nit of the misnamed AAELF32 HTML file:

When I try to view the HTML translations of these documents locally in Firefox using `file://` URIs, I see runaway hyperlinks covering a lot of the text. It looks to me as if that's because Firefox is trying to parse them as trad HTML, which means it's not recognising the self-closing effect of tags like `<a href="thing"/>`.

Renaming the file to have a `.xhtml` file extension causes Firefox to handle it correctly. I expect I could also make it work by putting the file on an actual HTTP server and configuring it to send an appropriate Content-Type header, but that would be a much bigger headache to set up.

Since I think the intended use of these HTML translations is to view them locally rather than putting them on a web server, perhaps we should rename them all to `.xhtml`? That would make life easier for at least Firefox users. (I don't know whether other browsers use the same heuristic.)

legacy-documents/aaelf32/ihi0044_H/IHI0044G_aaelf.html
stuij: Thanks, done, and renamed the PDF as well.

stuij: also changed `.html` to `.xhtml`

sallyarmneale (commented):
Looks good to me

statham-arm (approved):
LGTM this time, thanks!

PR #166, opened 2022-08-17
Add initial draft of MemtagABI.

hctim:
MemtagABI is an AArch64 ELF ABI extension that allows for tagging of
stack, heap, and global variables. The ABI includes:

 1. Dynamic array entries for instructing the dynamic loader to control
    the MTE mode, heap, stack, and global tagging.
 2. A compressed table of metadata that's used to tag global variables
    at runtime.
 3. Extensions to existing relocations to allow tagging of global
    variables at runtime, while still allowing correct behaviour when
    run on non-MTE hardware or by non-MemtagABI aware loaders.


hctim:
@smithp35 @MaskRay @pcc @eugenis @fmayer @enh

This is related to https://reviews.llvm.org/D128958.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+   PageBreak
+
+Run-time dynamic linking
+========================
+
+Process initialization using dynamic table entries
+--------------------------------------------------
+
+During process intialization, the dyanmic loader should use the MemtagABI
```
fmayer, line 658: nit: dynamic

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+2. Creating a new MTE-protected main stack and calling a function that
+   transitions onto the new stack, preserving all existing frames on the
+   untagged stack.
+
+Global tagging
+--------------
+
+Because memory tags can only be applied to tag-capable memory, the dynamic
+loader should ensure that any segments containing global variables are mapped as
+tag-capable. On linux, this is done by ensuring that ensuring that the mappings
```
fmayer, line 688: nit: Linux (upcase)

fmayer (commented):
Made a quick pass, need to give more thought, but some nits

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  semantics description in this document.
+
+.. raw:: pdf
+
+   PageBreak
+
+Dynamic Section
+=============
+
+MemtagABI adds the follwing processor-specific dynamic array tags:
```
kbeyls, line 322: s/follwing/following/

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+loader at load time. In order to accomplish this, the loader needs to know the
+set of regions that need tagging (exactly which symbol corresponds to each
+region is not necessary). For each global variable, we emit a compressed
+"descriptor" to the ``SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`` section.
+
+The descriptor is comprised of two parts:
+
+1. The unrelocated virtual address of the global variable, and
+
+2. The size of the global variable.
```
DavidSpickett, line 481: This is the granule boundary aligned/extended size, correct? Would be good to make that 100% clear, lest someone think this is sizeof(the variable).

Actually I see below you talk about the size being granule aligned. Though I think it would still help to note that here.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+      last_symbol_end = symbol.address + symbol.size
+
+      if size < (1 << distance_reserved_bits):
+         write_to_section(uleb128(distance | size))
+      else:
+         write_to_section(uleb128(distance))
+         write_to_section(uleb128(size - 1))
+
+.. raw:: pdf
+
```
DavidSpickett, line 544: I would really like to see pseudocode for the decode as well. The encode is easy enough to follow but there's some ordering that would benefit from being shown in a decode example.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+- ``LDG(value)`` is an instruction for the run-time environment to materialize a
+  tagged pointer from the provided address, using the ``ldg`` instruction. This
+  operation should also align the pointer down to the closest tag granule before
+  executing the ``ldg`` instruction.
+
+- ``TAG(pointer, offset)`` is an instruction for the run-time environment to
+  fetch the memory tag at ``(pointer + offset) & ~0xf``, which can be done using
+  the ``ldg`` instruction. Then, this tag should be materialized to ``pointer``,
+  overwriting the existing address tag in ``pointer`` if it exists.
```
DavidSpickett, line 566: Just checking, here we are only talking about changing values in a pointer value, right?

ldg the existing tag from some location, then set that as `pointer`'s logical tag.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+In order to support materialization of the address tag into the relevant GOT
+entry for a tagged global, some AArch64 relocations have semantics that extend
+the AAELF64_ definition. Because global variable support in the MemtagABI is
+designed to be backwards-compatible, the result of the relocation on a
+non-MemtagABI binary or non-MTE device is the same as the non-extended variant.
+
+The relocations reference the following mnemonics:
+
+- ``LDG(value)`` is an instruction for the run-time environment to materialize a
```
DavidSpickett, line 558: What is the "value" here? If it's a parameter it is unused.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+``R_AARCH64_ABS64`` and ``R_AARCH64_GLOB_DAT`` are thus extended to materialize
+the memory tag of the symbol ``S`` for relocation purposes.
+
+Extended semantics of ``R_AARCH64_RELATIVE``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``R_AARCH64_RELATIVE`` is extended with special care to preserve backwards
+compatibility. This relocation extension allows for the out-of-bounds derived
+symbols to be resolved with the correct memory tags, as shown in the example
+below:
```
DavidSpickett, line 593: Might be good to include the specific term "tag derivation" in this first paragraph because later on you use it and it's not very clear that these examples are examples of "tag derivation". You use almost the same words in this intro but not quite.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+frames as taggable is dangerous, and so dynamic loader developers may find it
+easiest to implement this functionality by:
+
+1. Ensuring the implementation completes before any threads have spawned, and
+
+2. Creating a new MTE-protected main stack and calling a function that
+   transitions onto the new stack, preserving all existing frames on the
+   untagged stack.
+
+Global tagging
```
DavidSpickett, line 683: Should this be "Globals tagging" (plural)? Or "Global variable tagging", which is a bit less awkward.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+``DT_AARCH64_MEMTAG_GLOBALSSZ`` dynamic table entry. When loading a shared
+object (including the main executable), prior to processing relocations, the
+dynamic loader should decompress the descriptors, and for each descriptor,
+perform the following operation:
+
+1. Generate a random tag (probably using the ``irg`` instruction). Ideally, to
+   ensure linear global buffer overflows are caught determinstally, the tag
+   should be different from adjacent global variables.
+
+2. Apply the random tag, using the ``stg`` family of instructions to the number
```
DavidSpickett, line 709:
```
2. Apply the random tag, using the ``stg`` family of instructions, to the number
   of granules specified in the descriptor.
```
Extra comma before "to the number". Or you could put "using the stg family of instructions" in brackets.
```
2. Apply the random tag (using the ``stg`` family of instructions) to the number
   of granules specified in the descriptor.
```

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+   of granules specified in the descriptor.
+
+dlopen/dlsym
+------------
+
+On many platforms, programs can load shared libraries at run-time via ``dlopen``
+and access symbols in that library via ``dlsym`` or ``dlvsym``. Tagged globals
+sohuld be assigned memory tags at ``dlopen``-time, and symbol addresses returned
```
DavidSpickett, line 717: sohuld -> should

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+         write_to_section(uleb128(distance | size))
+      else:
+         write_to_section(uleb128(distance))
+         write_to_section(uleb128(size - 1))
+
+.. raw:: pdf
+
+   PageBreak
+
+Relocation Operations
```
smithp35, line 547: I've done a quick check to see if any of this conflicts with the PAuthABI extension. As that uses new relocation codes https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#9auth-variant-dynamic-relocations it doesn't directly conflict.

It may be worth having a read of the PAuthABI to see if there is a scope for a signed pointer to a tagged location. While I don't think it is necessary to document all the details, it would be good to document what the expectation of the two extensions working together are. For example it could be that it is PAuthABI or Memtag and not both.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  +------------+--------------------+----------------+-------------------------------------+
+
+``R_AARCH64_ABS64`` and ``R_AARCH64_GLOB_DAT`` are thus extended to materialize
+the memory tag of the symbol ``S`` for relocation purposes.
+
+Extended semantics of ``R_AARCH64_RELATIVE``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``R_AARCH64_RELATIVE`` is extended with special care to preserve backwards
+compatibility. This relocation extension allows for the out-of-bounds derived
```
smithp35, line 591: Would be good to have a definition of out-of-bounds symbol is, unless it is defined in another referenced document. I'm guessing the bounds of a symbol are the addresses covered by  `[st_value, st_value + st_size)`

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+   // GOT_ENTRY_FOR_foo populated with R_AARCH64_RELATIVE, as usual, with
+   // addend zero and tag-offset zero. This can be compressed to a RELR
+   // relocation.
+   int* foo[32];
+
+   // GOT_ENTRY_FOR_foo_middle populated with R_AARCH64_RELATIVE, with addend
+   // 16 and tag-offset -16. Because the linker knows that the tag-offset is
+   // in-bounds of the symbol, the linker can omit the tag-offset, and
+   // compress this to a RELR relocation.
```
smithp35, line 608: Would be good to add a reference to RELR in the references section. Best that I know of is https://groups.google.com/g/generic-abi/c/bX460iggiKg

smithp35 (commented):
Thanks very much for writing this up. My comments are more as a person familiar with static linkers rather than MTE.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+.. raw:: pdf
+
+   PageBreak
+
+Introduction
+============
+
+The Armv8.5-A architecture introduced the Memory Tagging Extensions (MTE)
+feature, which allows each 16-byte region (a "tag granule") of taggable memory
```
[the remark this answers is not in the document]
hctim: Thanks, done.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+to have a 4-bit piece of metadata (a "memory tag") associated with it.
+Additionally, the 2nd top nibble of a pointer may contain a similar piece of
+metadata (a "pointer tag" or "allocation tag"). During load and store
```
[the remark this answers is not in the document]
hctim: Thanks, made both changes (and /s/pointer tag/logical tag and /s/memory tag/allocation tag throughout)

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+- Provide primitives that can be used to support protection of the heap, stack,
+  and global variables with memory tags.
+
+- Provide rationale for design choices.
+
+- Describe a scheme for tagging of the heap and global variables that is
+  backwards-compatible. In particular, a binary built with MTE heap or globals
+  protection will function correctly when linked with a non-MemtagABI aware
+  linker, loaded with a non-MemtagABI aware dynamic loader, or run on non-MTE
+  capable hardware. Note, it is expected that stack instrumentation, because MTE
```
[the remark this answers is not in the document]
hctim: Sure. Borrowed it from aaelf32, which looks like github's rst processor just spits out as text, but I'm assuming the PDF looks great :)

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  instructions are not part of the no-op space, may crash on non-MTE hardware or
+  non-MemtagABI linkers/loaders.
```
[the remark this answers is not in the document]
hctim: > For the non-MemTagABI loaders, is it that the loader wouldn't setup the stack in tagged memory and then when the store tag instructions run it will crash?

Almost. The stack won't be tagged, and so the inline `stg`-family instructions generated by the compiler instrumentation will be no-ops. Basically, on MTE HW with a non-MemtagABI-aware loader, stack instrumentation doesn't crash but doesn't do anything interesting.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+Resolving Multiple Definitions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Ideally, all translation units linked into a program are all built with or
+without tagged globals. This may not always be possible however, and so the
+linker must take special care when resolving the definition of a global variable
+if there are duplicate definitions that are not all tagged or untagged. In these
+instances, the linker must:
+
+1. Ensure the resolved definition is untagged, and
```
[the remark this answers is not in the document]
hctim: I don't have a super concrete example where things could go wrong if we choose the other option, but it's just the conservative approach.

Changed the text a little in the list element to try and convey the reasoning inline.

memtagabielf64/memtagabielf64.rst
hctim: Added in the `Run-time dynamic linking` -> `Global tagging` section.

memtagabielf64/memtagabielf64.rst
hctim: Yep. Cleaned up the sentence to try and make it more clear.

memtagabielf64/memtagabielf64.rst
hctim: Cleared it up. `value` here is a pointer.

memtagabielf64/memtagabielf64.rst
hctim: thanks. hopefully i've clearly defined it below (in the comment about `GOT_ENTRY_FOR_foo_end`).

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+=============
+
+MemtagABI adds the follwing processor-specific dynamic array tags:
+
+.. table:: MemtagABI specific dynamic array tags
+
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | Name                            | Value      | d\_un  | Executable        | Shared Object     |
+  +=================================+============+========+===================+===================+
+  | ``DT_AARCH64_MEMTAG_MODE``      | 0x70000006 | d\_val | Platform specific | Platform Specific |
```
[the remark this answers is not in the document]
hctim: Thanks for the note, have adjusted.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+protected with memory tagging. The value is the unrelocated virtual address
+which points to the start of the ``SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`` section.
+Each dynamically loaded object (including the main executable) should have its
+own ``DT_AARCH64_MEMTAG_GLOBALS`` dynamic table entry, if said object contains
+tagged global variables.
+
+``DT_AARCH64_MEMTAG_GLOBALSSZ`` is the size of the
+``SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`` section, in bytes.
+
+These values are in the AArch64 Processor-specific range. The values are subject
```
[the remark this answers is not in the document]
hctim: This comment was cargo-culted from pauthabielf, let me know if you'd rather it gone :).

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+If the translation unit contains tagged global variables, the compiler should
+produce an ``SHT_AARCH64_MEMTAG_GLOBALS_STATIC`` section, as well as a
+relocation section with ``sh_link`` pointing to the
+``SHT_AARCH64_MEMTAG_GLOBALS_STATIC`` section.
+
+For each tagged global variable, the compiler should produce an ``R_NONE``
+relocation placed in the aforementioned relocation section. This relocation
+should have its ``ELF64_R_SYM`` bits of the ``r_info`` field point to the global
+variable that needs to be tagged. This communicates to the linker that the
+global variable in question is marked for tagging.
```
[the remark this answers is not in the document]
hctim: Admittedly, I don't know comdats that well. But, my understanding is that there should be one `SHT_AARCH64_MEMTAG_GLOBALS_STATIC` per comdat to allow linker GC. Updated the text to reflect this.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  |                                        |                |
+  +----------------------------------------+----------------+
+  | ``SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`` | ``0x70000008`` |
+  |                                        |                |
+  |                                        |                |
+  +----------------------------------------+----------------+
+
+``SHT_AARCH64_MEMTAG_GLOBALS_STATIC`` is a section type used during static
+linking. This section is of size ``0``. For each tagged global variable in the
+object file, there exists an ``R_NONE`` relocation that references both the
```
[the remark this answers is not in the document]
hctim: Updated to R_AARCH64_NONE: MTE is only for AArch64 :)

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  | 257	      | R_AARCH64_ABS64    | S + A          | LDG(S) + A                          |
+  +------------+--------------------+----------------+-------------------------------------+
+  | 1025	      | R_AARCH64_GLOB_DAT | S + A          | LDG(S) + A                          |
+  +------------+--------------------+----------------+-------------------------------------+
+  | 1027	      | R_AARCH64_RELATIVE | Delta(S) + A   | TAG(Delta(S) + A, \*(Delta(S) + A)) |
+  +------------+--------------------+----------------+-------------------------------------+
+
+``R_AARCH64_ABS64`` and ``R_AARCH64_GLOB_DAT`` are thus extended to materialize
+the memory tag of the symbol ``S`` for relocation purposes.
+
```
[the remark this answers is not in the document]
hctim: It is extended to all relocations of this type, but for `ABS64/GLOB_DAT` relocations of untagged symbols, the materialized tag is zero (meaning it's a no-op). At relocation time, the loader has no knowledge whether the relocation is being applied to a tagged symbol or not :).

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  | ``DT_AARCH64_MEMTAG_MODE``      | 0x70000006 | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_HEAP``      | 0x70000007 | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_STACK``     | 0x70000008 | d\_ptr | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_GLOBALS``   | 0x70000009 | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_GLOBALSSZ`` | 0x7000000a | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
```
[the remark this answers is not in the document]
hctim: Thanks, updated (along with your above comment about odd/even).

memtagabielf64/memtagabielf64.rst
hctim: Thanks for the hint, I'll keep this thread open as I investigate further.

memtagabielf64/memtagabielf64.rst
hctim: > I'm guessing the bounds of a symbol are the addresses covered by [st_value, st_value + st_size)

Correct. I've tried to clean up the wording around the out-of-bounds handling. Hope it helps.

hctim:
Not hugely familiar with Github's review process - whether inline comments are published on-demand or only on this comment is unclear to me.

I've addressed the outstanding reviewer comments, leaving open anywhere that I'm still investigating or where it would be great to get an ack that I've done what you asked.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
smithp35: It's not doing any harm. Happy to leave it as is.

memtagabielf64/memtagabielf64.rst
smithp35: Ack

smithp35 (commented):
Thanks for the updates. I'm still trying to navigate my way through github's review interface as well. Hope I haven't missed anything it seems to have a habit of hiding comments.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
DavidSpickett: Ok right, I just made the assumption it would fault but the pseudo code backs you up. Makes sense now.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+``DT_AARCH64_MEMTAG_MODE`` indicates the initial MTE mode that should be set. It
+has two possible values: ``0``, indicating that the desired MTE mode is
+Synchronous, and ``1``, indicating that the desired mode is Asynchonous. This
+entry is only valid on the main executable, usage in dynamically loaded objects
+is ignored.
```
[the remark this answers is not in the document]
DavidSpickett: > Asymmetric mode (i.e. FEAT_MTE3) is implemented in the Linux kernel

I am behind the times! All sounds good to me.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,757 @@
...
+
+- All tags are randomly assigned at run-time. This prevents an attacker with
+  access to a compiled binary from knowing the tag values for a given pointer
+  ahead of time.
+
+General Restrictions
+--------------------
+
+- Global variables may only be tagged in dynamic executables, as materialization
+  of logical tags happens during relocation.
```
DavidSpickett, line 314: Should this be "materialization of tags" instead? If I'm correct in thinking that both logical and allocation tags would be set at this stage.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,757 @@
...
+   of granules specified in the descriptor.
+
+The decompression and tagging of the section is described in the pseudocode
+below:
+
+.. code-block::
+
+   distance_reserved_bits = 3
+
+   i = 0, virtual_address = load_bias
```
DavidSpickett, line 729: Is `load_bias` going to be the address at which you are loading the globals section?
(which makes sense to me given that these records are storing offsets)

Unless load_bias is referenced elsewhere it would be good to note what it means.

DavidSpickett (changes requested):

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,757 @@
...
+   It is expected that stack instrumentation, because MTE instructions are not
+   part of the no-op space, may crash on non-MTE hardware or non-MemtagABI
+   linkers/loaders.
+
+General Principles
+------------------
+
+- All tags are randomly assigned at run-time. This prevents an attacker with
+  access to a compiled binary from knowing the tag values for a given pointer
+  ahead of time.
```
mmalcomson, line 308: Phrasing nit (feel free to ignore if you disagree):
I believe there is some non-randomness suggested for global objects (to allow linear out-of-bounds accesses to be caught deterministically) and IIRC objects in the same frame on the stack have correlated tags to avoid too much of an increase in register pressure.
Given that I think it would be a bit clearer if this first sentence were a little less absolute.
The best I can think of right now is something like "All tags include a random component generated at runtime".

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,757 @@
...
+
+  +------------+--------------------+----------------+-------------------------------------+
+  | ELF64 Code | Name               | Base Operation | MemtagABI Extended Operation        |
+  +============+====================+================+=====================================+
+  | 257	      | R_AARCH64_ABS64    | S + A          | LDG(S) + A                          |
+  +------------+--------------------+----------------+-------------------------------------+
+  | 1025	      | R_AARCH64_GLOB_DAT | S + A          | LDG(S) + A                          |
+  +------------+--------------------+----------------+-------------------------------------+
+  | 1027	      | R_AARCH64_RELATIVE | Delta(S) + A   | TAG(Delta(S) + A, \*(Delta(S) + A)) |
+  +------------+--------------------+----------------+-------------------------------------+
```
mmalcomson, line 589: This strikes me as odd -- can I double-check that my understanding is correct?
If we want a relocation pointing to `foo + 32` where `foo` is only 32 bytes, does this mean that the offset of `-32` must be located at `foo + 32`?
That seems like it would require much more complexity around layout than we have already.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+=============
+
+MemtagABI adds the following additional Processor-specific section types:
+
+.. table:: MemtagABI Section Types
+
+  +----------------------------------------+----------------+
+  | Name                                   | Value          |
+  +========================================+================+
+  | ``SHT_AARCH64_MEMTAG_GLOBALS_STATIC``  | ``0x70000007`` |
```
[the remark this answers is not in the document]
hctim: `SHT_AARCH64_MEMTAG_GLOBALS_STATIC` shouldn't have `SHF_ALLOC` and should be discarded by the linker.

`SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC` needs `SHF_ALLOC` as it's used by the loader (although the loader could unmap the section after the globals are tagged if it's possible and desired).

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+variable that needs to be tagged. This communicates to the linker that the
+global variable in question is marked for tagging.
+
+In addition, the linker should ensure that all references to tagged global
+variable are via. the GOT (including for internal symbols). The relocations
+described later in this document ensure that the address tag for a global
+variable is materialized into the GOT, resulting in correctly tagged global
+variable accesses.
+
+Dynamic Linking
```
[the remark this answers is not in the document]
hctim: The dynamic loader is responsible for tag materialization, using the descriptors in `SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`. Have updated the text to reflect this.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+
+The dynamic linker should parse the ``SHT_AARCH64_MEMTAG_GLOBALS_STATIC``
+section to identify all tagged global variables.
+
+Resolving Multiple Definitions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Ideally, all translation units linked into a program are all built with or
+without tagged globals. This may not always be possible however, and so the
+linker must take special care when resolving the definition of a global variable
```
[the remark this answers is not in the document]
hctim: Sorry for the confusion, this section describes symbol resolution for multiple mismatching declarations, not multiple mismatching definitions.

If the definition is tagged, and one of the uses is untagged (i.e. the declaration is also untagged), then things can go wrong. One example is that a tagged symbol is generated, but there is a reference in hand-written assembly that's untagged, and the assembly uses a PC-relative reference rather than a GOT-relative reference to the global. An example from HWASan (a software implementation of MTE) is here: https://boringssl-review.googlesource.com/c/boringssl/+/36924

Hence, when we have mismatched declarations, we do the conservative thing and demote the global to be untagged. Updated the text to try and be more clear, please LMK.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+  fetch the memory tag at ``(pointer + offset) & ~0xf``, which can be done using
+  the ``ldg`` instruction. Then, this tag should be materialized to ``pointer``,
+  overwriting the existing address tag in ``pointer`` if it exists.
+
+- For all the relocation types listed below, the loader should use the ``ldg``
+  instruction on the target address before writing into the target field, as the
+  target field may be inside of a tagged region.
+
+.. table:: Relocations with extended semantics
+
```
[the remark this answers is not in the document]
hctim: Thanks, should be fixed.

memtagabielf64/memtagabielf64.rst
hctim: Thanks, clarified here. I use "materialization" specifically to describe the operation of the `ldg` instruction (loading the allocation tag and assigning it as a logical tag of a poionter).

memtagabielf64/memtagabielf64.rst
hctim: `load_bias` here is the virtual address that the DSO is loaded at. I wonder if there's some other terminology for this behaviour, it's very common parlance in our team, but I shockingly can't find it referenced in an authoritative relevant document upstream (`man elf` for example doesn't mention it).

Interesting, I'll define it in a comment in the pseudocode.

memtagabielf64/memtagabielf64.rst
hctim: Yeah, thanks, forgot here that stack tagging is semideterministic (frame is random, within a frame is incremental).

Added the text.

memtagabielf64/memtagabielf64.rst
hctim: Almost, a relocation pointing to `foo + 32` where `foo` is 32 bytes, the `-32` offset is put wherever the relocation result would be written.

More concretely:

```
char a[32]; // &a[0] = 0x1000 (.bss)
char* b = a[32]; // &b = 0x2000 (.data)
```

To relocate `b`, we generate a `R_AARCH64_RELATIVE` relocation for `0x1020 (0x1000 + 32)`. Then, we store the "tag derivation offset" in the place being relocated, i.e. `*0x2000 == -32`. Thus, the tag is loaded from `0x1020 + *0x2000 -> 0x1020 + (-32) -> 0x1000`.

mmalcomson: That makes a lot more sense, thanks!

IIUC the use of `*(Delta(S) + A)` for the offset argument to `TAG` doesn't describe that.
Since `Delta(S) + A` represents the value that we calculate with the `R_AARCH64_RELATIVE` operation -- which is `foo + 32`.

Would `Delta(0) + <some new marker to represent r_offset>` represent that value better?
(I'm suggesting `Delta(0)` since I believe that makes it explicitly calculate the offset of the place.
Since IIUC even though RELATIVE relocations are usually given the NULL symbol they could be given some other symbol and end up using an offset that may not work for the place we want to store the relocation.)

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
hctim: > IIUC the SHT_AARCH64_MEMTAG_GLOBALS_STATIC will not have the SHF_ALLOC flag.

Yep, that's correct. `SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC` will have `SHF_ALLOC`, but that's synthesized by the linker.

Ack, thanks for the info.

memtagabielf64/memtagabielf64.rst
hctim: Hmm, seems like `*P` is maybe a better way to describe this (i.e. the offset stored in the place). Given the operation is much more succinct with this, removed the whole `TAG()` operation and inlined it.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
smithp35: Ack, thanks for the update.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
smithp35: Ack, thanks for the update.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
smithp35: Ack. Thanks for the update. It is clear that there are declarations involved and not multiple definitions.

A related thought to an earlier comment about a different approach using a set of GOT generating relocations. We'd need something like :tag: and :tag_lo12: that an assembly file would need to use these to get the tagging behaviour.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,723 @@
...
+On many platforms, programs can load shared libraries at run-time via ``dlopen``
+and access symbols in that library via ``dlsym`` or ``dlvsym``. Tagged globals
+sohuld be assigned memory tags at ``dlopen``-time, and symbol addresses returned
+by ``dlsym`` should have the correct address tag materialized using the ``ldg``
+instruction.
+
+.. raw:: pdf
+
+   PageBreak
```
smithp35, line 766: The trailing PageBreak looks like causes GitHub to render #PageBreak# , could be worth taking it out. Looking at a sample of some of the other ABI docs I can't see another example with a trailing PageBreak.

stuij:
@hctim could you add this document to 
- tools/common/check-rst-syntax.sh
- tools/rst2pdf/generate-pdfs.sh
- toplevel README.md

Adding the document to `generate-pdfs.sh` will have the added benefit that it will be picked up by CI, which will include the PDF in the PDFs.zip zipfile of the CI run, which you can access if you click on the Details link for the CI run on this page, and scroll to the bottom part of the summary page: https://github.com/ARM-software/abi-aa/actions/runs/2959474522

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,766 @@
+..
+   Copyright (c) 2022, Arm Limited and its affiliates.  All rights reserved.
```
stuij, line 2: Could you add yourself (I guess it'd be Google) as a copyright owner where appropriate as per the updated contribution guidelines.

memtagabielf64/memtagabielf64.rst
hctim: Thanks, done.

memtagabielf64/memtagabielf64.rst
hctim: Done, thanks.

hctim:
> @hctim could you add this document to
> 
> * tools/common/check-rst-syntax.sh
> * tools/rst2pdf/generate-pdfs.sh
> * toplevel README.md
> 
> Adding the document to `generate-pdfs.sh` will have the added benefit that it will be picked up by CI, which will include the PDF in the PDFs.zip zipfile of the CI run, which you can access if you click on the Details link for the CI run on this page, and scroll to the bottom part of the summary page: https://github.com/ARM-software/abi-aa/actions/runs/2959474522

Thanks, done all.

Folks who have left comments that are still open (and that I've responded to) - would you mind taking a quick look and making sure that your comments were sufficiently addressed?

I think the only outstanding thing here is to address the pauth+memtag pointers.

Who'd be the right person to approve/merge this document? Have they had an opportunity to review it?

memtagabielf64/memtagabielf64.rst
DavidSpickett: For whatever reason, I can't resolve my own comment, but this looks good to me.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,773 @@
...
+``R_AARCH64_NONE`` relocation placed in the aforementioned relocation section.
+This relocation should have its ``ELF64_R_SYM`` bits of the ``r_info`` field
+point to the global variable that needs to be tagged. This communicates to the
+linker that the global variable in question is marked for tagging.
+
+In addition, the linker should ensure that all references to tagged global
+variables are via. the GOT (including for internal symbols). The relocations
+described later in this document ensure that the dynamic loader is able to
+materialized logical tags for global variables into the GOT, resulting in
+correctly tagged global variable accesses.
```
mmalcomson, line 463: Typo:  `is able to materialized logical tags`  : materialized -> materialize

smithp35:
> > @hctim could you add this document to
> > 
> > * tools/common/check-rst-syntax.sh
> > * tools/rst2pdf/generate-pdfs.sh
> > * toplevel README.md
> > 
> > Adding the document to `generate-pdfs.sh` will have the added benefit that it will be picked up by CI, which will include the PDF in the PDFs.zip zipfile of the CI run, which you can access if you click on the Details link for the CI run on this page, and scroll to the bottom part of the summary page: https://github.com/ARM-software/abi-aa/actions/runs/2959474522
> 
> Thanks, done all.
> 
> Folks who have left comments that are still open (and that I've responded to) - would you mind taking a quick look and making sure that your comments were sufficiently addressed?
> 
> I think the only outstanding thing here is to address the pauth+memtag pointers.
> 
> Who'd be the right person to approve/merge this document? Have they had an opportunity to review it?

We're working through the process at the moment, I think the major things we need are:
* The copyright according to the recently updated CONTRIBUTION guide. I think this has been done.
* Assign a document owner to the table in https://github.com/ARM-software/abi-aa/blob/main/CONTRIBUTING.md . Would you be willing to be the owner, or one of the owners? I think this modification could be done in a follow up patch, possibly by someone from Arm, so no need to update the PR yet.
* Make sure we have the document status correct as we are expecting this document to evolve over time. This should be good already (Alpha).
* Ensure that the document has been reviewed, which I think we're getting to the end of that process now.

Will hopefully make some progress on these tomorrow. We also need to work out a process for how to get an Alpha merged in to the main ABI once it has stabilised.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,773 @@
...
+  | ``DT_AARCH64_MEMTAG_STACK``     | 0x7000000c | d\_ptr | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_GLOBALS``   | 0x7000000d | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+  | ``DT_AARCH64_MEMTAG_GLOBALSSZ`` | 0x7000000f | d\_val | Platform specific | Platform Specific |
+  +---------------------------------+------------+--------+-------------------+-------------------+
+
+``DT_AARCH64_MEMTAG_MODE`` indicates the initial MTE mode that should be set. It
+has two possible values: ``0``, indicating that the desired MTE mode is
+Synchronous, and ``1``, indicating that the desired mode is Asynchonous. This
```
MaskRay, line 363: What "Synchronous" and "Asynchronous" mean are not explained.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,773 @@
...
+entry is only valid on the main executable, usage in dynamically loaded objects
+is ignored.
+
+The presence of the ``DT_AARCH64_MEMTAG_HEAP`` dynamic array entry indicates
+that heap allocations should be protected with memory tagging. Implementation of
+this logic is left to the heap allocator. This entry is only valid on the main
+executable, usage in dynamically loaded objects is ignored.
```
MaskRay, line 370: "only valid on the main executable" makes ", usage in dynamically loaded objects is ignored" unneeded. Ditto below.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
MaskRay: I am a bit confused by static Linking and dynamic linking usage here. Are both for relocatable files? Is one for shared object linking (`-shared`) and one for executable (`-pie`, `-no-pie`)?

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
MaskRay: Making the linker inspect all relocatable files to check a symbol is an unusual request to the linker. For efficient linker design, it just discards all non-prevailing COMDAT groups. Is this really the responsibility of a linker?

Perhaps I have missed a big picture. A companion design doc about how this is imagined to be used in the llvm toolchain with a concrete example (test source code, linker command line, runtime behavior, etc) may help. Some components may not be the scope of this ABI doc, though.

memtagabielf64/memtagabielf64.rst
hctim: Sure, removed.

memtagabielf64/memtagabielf64.rst
hctim: Added some descriptions/references in the "Terms and Abbreviations" section.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
hctim: Updated some of the headers. If the following isn't clear, please let me know:

 - Compiler produces globals of `size % 16 == 0`, `alignment >= 16 && alignment % 16 == 0`.
 - Compiler produces `SHT_AARCH64_MEMTAG_GLOBALS_STATIC`.
 - Linker reads `SHT_AARCH64_MEMTAG_GLOBALS_STATIC` to identify tagged globals.
 - Linker produces `SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`.
 - Linker produces extended-semantics relocations to tagged globals.
 - Linker ensures tagged globals are referenced via. the GOT.
 - Dynamic loader reads `SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC`, assigns random tags at runtime during loading.
 - Dynamic loader also populates GOT entries with correct address tags using extended relocation semantics.

memtagabielf64/memtagabielf64.rst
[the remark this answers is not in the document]
hctim: The intention is not to scan symbols inside of discarded COMDAT groups - it's just to keep track of the tagged-ness of the declarations and use that information when selecting the definition. This is seemingly pretty cheap in LLD, a small update to `Symbol::mergeProperties`. I'll upload the tentative LLD patches soon so you can take a look.

hctim:
> We're working through the process at the moment, I think the major things we need are:
> 
> * The copyright according to the recently updated CONTRIBUTION guide. I think this has been done.
> * Assign a document owner to the table in https://github.com/ARM-software/abi-aa/blob/main/CONTRIBUTING.md . Would you be willing to be the owner, or one of the owners? I think this modification could be done in a follow up patch, possibly by someone from Arm, so no need to update the PR yet.

Wasn't sure on first glance that the owner could be non-ARM, but I've added myself there.

> * Make sure we have the document status correct as we are expecting this document to evolve over time. This should be good already (Alpha).
> * Ensure that the document has been reviewed, which I think we're getting to the end of that process now.
> 
> Will hopefully make some progress on these tomorrow. We also need to work out a process for how to get an Alpha merged in to the main ABI once it has stabilised.

Great, thanks very much.

memtagabielf64/TRADEMARK_NOTICE
```diff
@@ -1,6 +1,3 @@
-The text of and illustrations in this document are licensed by Arm
```
kbeyls: I agree that the "by Arm" words here aren't fully correct now that other entities are also contributing to these specifications.
I'll address that issue for the TRADEMARK_NOTICE files in the other ABI docs in a separate pull request that I will put up shortly.
Here, please only remove the 2 words "by Arm" at the end of the first line. The separate pull request I'll put up shortly will make the same change to the TRADEMARK_NOTICE files in the other ABI docs.

memtagabielf64/LICENSE
```diff
@@ -1,22 +1,202 @@
-This work is licensed under the Creative Commons
```
kbeyls: Thank you for responding to the review feedback regarding copyright statements, but please do not change the LICENSE file.
The contribution you're making is still under the same CC with patent grant license used in the other parts of the ABI specifications in this project; not under the Apache2.0 license as the proposed change to this LICENSE file would imply.
The only thing that you need to change is to add a line to the copyright header - I'll point to the right place to do so in another comment.

kbeyls (changes requested):

memtagabielf64/TRADEMARK_NOTICE
kbeyls: I've now created pull request #170 to address this issue for the TRADEMARK_NOTICE files in the other ABI docs.

sallyarmneale (commented):
LGTM

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+   untagged stack.
+
+Global variable tagging
+-----------------------
+
+Because allocation tags can only be applied to tag-capable memory, the dynamic
+loader should ensure that any segments containing global variables are mapped as
+tag-capable. On Linux, this is done by ensuring that ensuring that the mappings
+are created as anonymous mappings (``MAP_ANONYMOUS``) with the ``PROT_MTE``
+flag. Note that this requires special handling for when the linker is invoked
```
sallyarmneale, line 738:
```suggestion
flag. Note: this requires special handling for when the linker is invoked
```

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+         i += bytes_consumed
+
+      write_random_allocation_tag(virtual_address, ngranules * tag_granule_size)
+
+dlopen/dlsym
+------------
+
+On many platforms, programs can load shared libraries at run-time via ``dlopen``
+and access symbols in that library via ``dlsym`` or ``dlvsym``. Tagged globals
+should be assigned allocation tags at ``dlopen``-time, and symbol addresses
```
sallyarmneale, line 794: must?

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+returned by ``dlsym`` should have the correct logical tag materialized using the
```
sallyarmneale, line 795: will, or must? Avoid should as it creates ambiguity

sallyarmneale (commented):
Quite a few suggestions

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+AAELF64_.
+
+Tagging Global Variables
+========================
+
+Compilation
+-----------
+
+Global variables that wish to be tagged must meet MemtagABI-specific guidelines.
+In particular, a compiler that destines a global variable to be tagged at
```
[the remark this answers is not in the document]
hctim: patched up slightly differently.

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+--------------------------------------------------
+
+During process intialization, the dynamic loader should use the MemtagABI
+dynamic table entries to enable the correct MTE features at runtime.
+
+If ``DT_AARCH64_MEMTAG_MODE`` is present, the dynamic loader should (in a
+platform-specific specific way) enable MTE for the process. On Linux, this is
+done through ``prctl(PR_SET_TAGGED_ADDR_CTRL, PR_TAGGED_ADDR_ENABLE |
+PR_MTE_TCF_[A]SYNC ...)``. When ``DT_AARCH64_MEMTAG_MODE`` is ``0``, then
+synchronous MTE should be enabled. When ``DT_AARCH64_MEMTAG_MODE`` is ``1``,
```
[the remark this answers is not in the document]
hctim: rephrased slightly differently

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+asynchronous MTE should be enabled.
```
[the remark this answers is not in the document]
hctim: rephrased slightly differently

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+
+Process initialization using dynamic table entries
+--------------------------------------------------
+
+During process intialization, the dynamic loader should use the MemtagABI
+dynamic table entries to enable the correct MTE features at runtime.
+
+If ``DT_AARCH64_MEMTAG_MODE`` is present, the dynamic loader should (in a
+platform-specific specific way) enable MTE for the process. On Linux, this is
+done through ``prctl(PR_SET_TAGGED_ADDR_CTRL, PR_TAGGED_ADDR_ENABLE |
```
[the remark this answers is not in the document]
hctim: slightly rephrased, this is not a "you, developer should do this", it's a "the dynamic loader should do this on the behalf of you, the developer"

memtagabielf64/memtagabielf64.rst
```diff
@@ -0,0 +1,796 @@
...
+for heap allocations in this process.
+
+If ``DT_AARCH64_MEMTAG_STACK`` is present, the dynamic loader should enable
+tagging for the main stack and thread stacks. Re-mapping currently alive stack
+frames as taggable is dangerous, and so dynamic loader developers may find it
+easiest to implement this functionality by:
+
+1. Ensuring the implementation completes before any threads have spawned, and
+
+2. Creating a new MTE-protected main stack and calling a function that
```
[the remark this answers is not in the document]
hctim: Not sure this suggestion is right, given the concatenation of would be "implement this functionality by create".

smithp35:
I plan to check over these changes over the next couple of days, although please do give us a ping if this drags on too long. My plan is to:
* Check to find if there are any comments that are unanswered and summarise them. This may take a bit of time as I'm not that familiar with the github UI.
* Check over the most recent changes.
* With luck give a +1 to merging, from there we'll give some time for any other reviewers to comment.

At some point after approval but prior to merging, we'll ask you to squash the commits into one or more logical commits with a good commit message. We've recently found that the ABI history has been cluttered up with "address review comments" commits.

smithp35 (approved):
I've given this my approval. What I suggest is to leave this open till next week for any further comments/objections. Once that has been covered the changes can be squashed (may be possible to do that now if it doesn't lose any of the conversations) and merged.

I've run through the remaining opening conversations.
* One is purely language, I've left a suggestion but don't mind how it is resolved.
* Two open conversations are my own. One is an alternative suggestion to `SHT_AARCH64_MEMTAG_GLOBALS_DYNAMIC` but that will likely need implementation experience to decide if it is worthwhile. The other is about compatibility with the PAuthABI, I can't see any incompatibilities myself.
* I thought it would be worth clarifying if TLS variables are considered global or not. I'm expecting not.

I've also checked the license/trademark changes have been as per kbeyls suggestions.

hctim:
> I've given this my approval. What I suggest is to leave this open till next week for any further comments/objections. Once that has been covered the changes can be squashed (may be possible to do that now if it doesn't lose any of the conversations) and merged.

Thanks. AIUI, only the owners (or people with merge permissions) of the repo can merge the change. Even with the changes approved, I can't merge the branch, and it'd need to be done by you or someone from your team.

I was under the impression at that point, the default option was "squash and merge". I think a variant of the original commit message from the first version of the pull request might be a good one:

```
Add the Alpha version of MemtagABI.
    
MemtagABI is an AArch64 ELF ABI extension that allows for tagging of
stack, heap, and global variables. The ABI includes:

  1. Dynamic array entries for instructing the dynamic loader to control
     the MTE mode, heap, stack, and global tagging.
  2. A compressed table of metadata that's used to tag global variables
     at runtime.
  3. Extensions to existing relocations to allow tagging of global
     variables at runtime, while still allowing correct behaviour when
     run on non-MTE hardware or by non-MemtagABI aware loaders.
```

smithp35:
> > I've given this my approval. What I suggest is to leave this open till next week for any further comments/objections. Once that has been covered the changes can be squashed (may be possible to do that now if it doesn't lose any of the conversations) and merged.
> 
> Thanks. AIUI, only the owners (or people with merge permissions) of the repo can merge the change. Even with the changes approved, I can't merge the branch, and it'd need to be done by you or someone from your team.
> 
> I was under the impression at that point, the default option was "squash and merge". I think a variant of the original commit message from the first version of the pull request might be a good one:
> 
> ```
> Add the Alpha version of MemtagABI.
>     
> MemtagABI is an AArch64 ELF ABI extension that allows for tagging of
> stack, heap, and global variables. The ABI includes:
> 
>   1. Dynamic array entries for instructing the dynamic loader to control
>      the MTE mode, heap, stack, and global tagging.
>   2. A compressed table of metadata that's used to tag global variables
>      at runtime.
>   3. Extensions to existing relocations to allow tagging of global
>      variables at runtime, while still allowing correct behaviour when
>      run on non-MTE hardware or by non-MemtagABI aware loaders.
> ```

Thanks for the update. It looks like the only option I have is Rebase and Merge. If you are able to squash the commits using your suggest commit message then I'll try and merge tomorrow since it has been a week since my "any last comments" message.

hctim:
> Thanks for the update. It looks like the only option I have is Rebase and Merge. If you are able to squash the commits using your suggest commit message then I'll try and merge tomorrow since it has been a week since my "any last comments" message.

Alright, I force-pushed to my own branch to reset the history. Github UI now only shows a single commit in the pull request, which seems right to me.

I think the only outstanding question I had was with respect to the MEMTAG_GLOBALS_STATIC section vs. `st_other`: https://github.com/ARM-software/abi-aa/pull/166/files#diff-b74373b2ee6de4c5ae65712c0be8f666b7f192e3772f1a7b82126c00e320153c

smithp35 (approved):
Thanks for update, I've resolved the remaining conversations involving me. I think it would be worth an experiment to use new GOT+TAG generating relocations for tagged globals as I think it could simplify the linker implementation but this could be added later after gaining experience with the implementation.

PR #188, opened 2022-11-19
Fix formatting of v6 cell in core registers table

EleanorNB:
The alias `v6` was mistakenly in the `Special` column; I have moved it to `Synonym`. I have also included its other role in the `Role` column.

### DIFF
--- a/aapcs32/aapcs32.rst
+++ b/aapcs32/aapcs32.rst
@@ -819,11 +819,11 @@ one status register (CPSR) that is available for use in conforming code.
    +----------+---------+---------+-----------------------------------------+
    | r10      | v7      |         | Variable-register 7.                    |
    +----------+---------+---------+-----------------------------------------+
-   | r9       |         | v6      | Platform register.                      |
-   |          |         |         |                                         |
-   |          |         | SB      | The meaning of this register is defined |
-   |          |         |         | by the platform standard.               |
+   | r9       | v6      | SB      | Platform register or Variable-register  |
+   |          |         |         | 6.                                      |
    |          |         | TR      |                                         |
+   |          |         |         | The meaning of this register is defined |
+   |          |         |         | by the platform standard.               |
    +----------+---------+---------+-----------------------------------------+
    | r8       | v5      |         | Variable-register 5.                    |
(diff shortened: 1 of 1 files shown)


smithp35:
Thanks for the pull request. It looks right to me. It has been that way for as long as I can find the history of the document so I'd like to check to see if it wasn't intentionally put there. Hopefully will have an answer soon.

aapcs32/aapcs32.rst
```diff
@@ -819,11 +819,11 @@ one status register (CPSR) that is available for use in conforming code.
    +----------+---------+---------+-----------------------------------------+
    | r10      | v7      |         | Variable-register 7.                    |
    +----------+---------+---------+-----------------------------------------+
-   | r9       |         | v6      | Platform register.                      |
-   |          |         |         |                                         |
-   |          |         | SB      | The meaning of this register is defined |
-   |          |         |         | by the platform standard.               |
+   | r9       | v6      | SB      | Platform register or Variable-register  |
+   |          |         |         | 8.                                      |
```
smithp35, line 823: The 4th column has Variable-register 8, this should be Variable-register 6 to match v6.

smithp35 (commented):
As this is a trivial change, please amend the commit rather than making a new one for the change. The repository is rebase and merge so we're trying to avoid small typo fix review commits.

aapcs32/aapcs32.rst
EleanorNB: Whoops. Well, I just pulled and amended locally, but now it won't let me push. Apparently I'm a commit behind as well as being a commit ahead. I'm really not good at Git it seems.

EleanorNB: Whoops. Well, forced that down, should be good now.

stuij: you should be able to just force-push: `git push -uf <repo> <branch>`

EleanorNB: Yep, did that, thanks. I really should properly learn Git.

smithp35 (approved):
Thanks very much for the update. I'll merge the change.

PR #191, opened 2023-02-20
Adding specification to handle the new C2x _BitInt type

mmalcomson:
Here we add mappings from the new `_BitInt` type to machine types.

All that I believe needs to be changed in the ABI documents themselves are entries in the Mapping tables for the PCS rules.
I explain in the commit messages below the reasoning that this is all that is needed.

This PR also contains a rationale document.
As discussed I'm putting the initial proposed document up and going to add the comments I've already heard onto this github request (to keep the discussion as in the open as possible).


aapcs32/aapcs32.rst
```diff
@@ -301,6 +301,9 @@ changes to the content of the document for that release.
   | 2021Q1| 12\ :sup:`th` April 2021            | Clarify what it means for a VFP CPRC argument to be correctly     |
   |       |                                     | aligned.                                                          |
   +-------+-------------------------------------+-------------------------------------------------------------------+
+  | 2023Q1| 5\ :sup:`th` February 2022          | Document mapping from C2x _BitInt(N) and unsigned _BitInt(N) types|
+  |       |                                     | to machine types.                                                 |
+  +-------+-------------------------------------+-------------------------------------------------------------------+
```
mmalcomson, line 306: On reviewing my own patch -- I guess it shouldn't include an update in this table since this is the table.
I'm now starting to think that maybe this table is just for changes in releases, and that should be made upon a release.
Would appreciate a confirmation/refutation on that from someone more in touch with the releases @stuij ?

Similar for the aapcs64.rst document.

aapcs64/aapcs64.rst
```diff
@@ -249,6 +249,9 @@ changes to the content of the document for that release.
 | 2022Q3     | 20\ :sup:`th`      | - Add alpha-level support for SME.                               |
 |            | October 2022       | - Across the document, use “thread” rather than “process”.       |
 +------------+--------------------+------------------------------------------------------------------+
+| 2023Q1     | 5\ :sup:`th`       | - Document mapping from C2x _BitInt(N) and unsigned _BitInt(N)   |
+|            | February 2022      |   types to machine types.                                        |
++------------+--------------------+------------------------------------------------------------------+
```
mmalcomson, line 254: As above with the aapcs32.rst changelog.

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+This document describes the rationale behind the ABI choices made for using the
+bit-precise integral types defined in C2x.  These are ``_BitInt(N)`` and
+``unsigned _BitInt(N)``.  These are defined for integral ``N`` and each ``N`` is
+a different type.
+
+The proposal for these types can be found in following link.
+https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2763.pdf
+
+As the rationale mentioned, some applications have uses for a specific bit-width
+type.  In the case of writing C code which can be used to determine FPGA
```
mmalcomson, line 24: nitpick: /As the rationale *in that proposal* mentions/

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+hardware these specific bit-width types can lead to large performance and space
+savings.
+
+From the perspective of the Arm ABI we have some trade-offs to determine.  We
+need to choose a representation for these objects in memory and in registers
+along with the size and alignment of the objects.  The main trade-offs we have
+identified in this case are on performance between different types of C-level
+operations, whether certain hardware-level atomic operations are possible, and
+general familiarity of programmers with the representation.
+
```
mmalcomson, line 34: nitpick: maybe it reads more easily if this were a 3-item bullet list?

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+For this particular type we are estimating that the use of ``_BitInt`` types
+will not be such that operations on these types are performance critical.
+
```
mmalcomson, line 37: Less of a nitpick: I'm not sure we agreed that operations on these types
will never be performance critical, which the above sentence seems to
suggest?
I'd probably start by explaining the difficulty in defining ABI
rules/understanding the trade-offs as there is almost no code currently
using _BitInt, and it is next-to-impossible to predict exactly how it
will be used most often in the future.

Nonetheless, we need to make trade-off decisions.

Therefore, let's analyze some of the characteristics and features of
this type to try and make a best guess of how these types may get used
when target Arm CPUs.

(The specifics on not much code with _BitInt being performance critical
 could be repeated after the below analysis of use cases?)

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+There seem to be two different regimes for these types.  The "small" regime
+where bit-precise types could be stored in a single register, and the "large"
+regime where bit-precise types must span multiple registers.
```
mmalcomson, line 40: nitpick: maybe say general purpose register? Is the term "general
purpose register" used in other parts of the ABI docs?

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+
+These types must be at least byte-aligned so they are addressable, and at least
+rounded to a byte boundary in size for ``sizeof``.  Since these types have an
+aesthetic similarity to bit-fields one might expect better packing in an array
+of ``_BitInt(24)`` than an array of ``int32_t`` types (i.e. packing as good as a
+byte-array).  However, this would require a low alignment of such types and that
+would mean loading and storing of even "small" sized ``_BitInt``'s crossing
+cache boundaries -- leading to an unnecessary performance hit and hindering any
+atomic operations on these.
+
```
mmalcomson, line 53: nitpick on style: I'm wondering if it would be better to more explicitly
surface the pros and cons of a trade-off decision in the structure of
this document, rather than using full prose? For the example above, I'm
thinking of for example something like:

Trade-off decision 1: alignment of small-regime _BitInt

There are 2 obvious options:

option 1: 1 byte alignment
pros: list them
cons: list them

option 2: alignment rules same as 8/16/32/64-byte larger datatype (todo:
   maybe improve the wording of this title/option)
pros: list them
cons: list them

Discussion of option 1 vs option 2.
Describe why we chose option 2 based on the information and estimates we
currently have available.

(I noticed you have something similar further down for "Representation
in bits". It might be good to have the same structure for analyzing
different trade-off choices to be made in this document?)

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+the largest type defined on that platform.  This falls out of the fact that
+double-register size maps to the largest integral Fundamental Data Type defined
+on both platforms.
+
+Hence for "large" sizes we are choosing to define a ``_BitInt(N)`` size and
+alignment by treating them "as if" they are an array of double-register sized
+Fundamental Data Types.
+
+Representation in bits
+----------------------
```
mmalcomson, line 81: Should the above title be "Representation in registers" instead of "...
in bits"?

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,213 @@
...
+the most surprising representation for developers on an Arm platform.
+
+Between option ``B`` and option ``C`` there is not a great difference in
+performance characteristics.  However it should be noted that option ``C`` is
+likely the most natural extension of the AArch32 PCS rules for unspecified bits
+in a register containing a small Fundamental Data Type, while option ``B`` is
+the most natural extension of the similar rules in AArch64 PCS.
+
+As mentioned above, we do not expect operations on ``_BitInt`` types to be
+performance critical.  Given that providing a productive environment for
```
mmalcomson, line 210: Overall, I'm not sure if we can expect _BitInt to never be used in
performance-critical code.
If people will write code targeting FPGAs with _BitInt, presumably that
code is performance-critical (why would anyone otherwise target an FPGA
for that code?). If such code becomes popular, at some point, there may
be good reasons to use the same source code to target CPUs too. In that
case, the _BitInt code would be performance critical...

mmalcomson (commented):
Added comments from Kristof Beyls gave me out-of-band.

Also added two comments that I noticed myself around improperly updating the ChangeLogs of the ABI documents.

workingjubilee:
Regarding `_BitInt`, I would like to raise a concern about whether C's `_BitInt(128)` and AArch64's `__int128` will be compatible in terms of layout, alignment, and parameter passing. It may significantly complicate ABI handling and type definitions, especially for other languages that are not C if they are not, and on AArch64, FFI compatibility with `__int128` is much more important than on x86-64 because it is used in the "hardware ABI" (it appears a lot in the layout of essential types). It also seems like a programmer footgun in general if they are inconsistent, even for just normal C code.

It's been a while since I've read the AAPCS64 last so apologies if this has already been addressed and the spec as-is will guarantee that, it wasn't immediately obvious from my skimming of this PR whether this was the case.

I agree that I don't think these will be performance-critical types otherwise.

design-documents/bit-precise-types.rst
workingjubilee: The justifications in N2763 mention that `_BitInt` to some degree exists to allow code that is used in high-level synthesis to have better performance by not having to participate in the usual arithmetic promotions, like promoting up to `int` (which is about 32 bits), and only promoting up to the other `_BitInt` in size (so that adding `_BitInt(4)` and `_BitInt(7)` gives you a `_BitInt(7)`, not a 32-bit `int`, which is easier to translate to an FPGA).

At larger sizes, it's more of a nicety (being able to "natively" specify "bigints" that the compiler handles for you), but an extremely appealing nicety.

design-documents/bit-precise-types.rst
workingjubilee: Perhaps it's more coherent to say "As mentioned above, we do not expect operations on ``_BitInt`` types _at large sizes_ to be performance critical." since the benefits of `_BitInt` allowing sidestepping the arithmetic promotions disappear for sizes larger than `_BitInt(32)`.

workingjubilee: Perhaps it would be more coherent to say "As mentioned above, we do not expect operations on ``_BitInt`` types _with N greater than 32 bits_ to be performance critical." due to N2763's rationale.

mmalcomson:
@workingjubilee Thanks for the feedback!

W.r.t. the `_BitInt(128)` yeah, the current spec does have them compatible with AArch64 `__int128`, but it's still good to hear that it was on someones mind since it increases our confidence in that choice.

The "performance critical" statement was a bit of a clumsy glossing-over of some analysis on my part.
I'm in the process of adding the details of the analysis to the rationale document, hopefully that would make it clear what the current thought process is -- I'd appreciate hearing your thoughts on whether the adjustment makes sense and is clear when I post it.

Cheers!

workingjubilee:
Aha, excellent!

> and on AArch64 ``_BitInt(128)`` has the same alignment and size as a ``__int128`` which is
the largest type defined on that platform.  This falls out of the fact that
double-register size maps to the largest integral Fundamental Data Type defined
on both platforms.

Ah, good, this was the part I missed. Thank you!

I'll be happy to look at things when you follow up!

design-documents/bit-precise-types.rst
```diff
@@ -20,72 +20,327 @@ a different type.
...
+  packing.
+
+Option ``B`` has the following benefits (both following from the alignment being
+greater than or equal to the size of the object in memory):
+
+- Avoid a performance hit since loading and storing of these "small" sized
+  ``_BitInt``'s will not cross cache boundaries.
+- Atomic loads and stores can be made on these objects.
+- The representation of bit-precise types of the same size as standard integer
+  types will have the same alignment and size in memory.
```
Wilco1, line 284: That last point is really the most important. The first 2 points are not too convincing since unaligned accesses are practically for free on modern cores, and LSE2 allows atomics on unaligned data.

design-documents/bit-precise-types.rst
```diff
@@ -135,16 +390,19 @@ It has the following negatives:
```
Wilco1, line 390: Again, mentioning atomics as the first, most important thing does not make any sense. For large bitints you can't do atomics at all (no atomics on int arr[100] either!), small bitints up to int128 could support atomics similar to how bitfields could support atomics (typically using CAS due to needing extra normalization step). And it's not like LDSMAX/LDUMAX is a popular atomic given it's not even a supported builtin in GCC...

design-documents/bit-precise-types.rst
```diff
@@ -194,20 +458,28 @@ It has the following negatives:
...
-performant.  However, when acting on "large" values (i.e. greater than the size
-of one register) it loses some of that benefit.  Storing to and from memory
-would also come at a cost for this representation.  This is also likely to be
-the most surprising representation for developers on an Arm platform.
+
+Overall it seems that option ``A`` is more performant for operations on small
+values.  However, when acting on "large" values (i.e. greater than the size of
+one register) it loses some of that benefit.  Storing to and from memory would
+also come at a cost for this representation.  This is also likely to be the most
+surprising representation for developers on an Arm platform.
```
Wilco1, line 469: Note there is no reason why a compiler couldn't do register operations in format A, while passing arguments/storing as C and assuming B when loading bitints or using parameters/return values. A seems most useful as an optimized internal representation (which is entirely private) rather than as the storage format.

Wilco1 (changes requested):
I agree with choosing option B and C for AArch64/AArch32 but I'd reduce the major focus on atomics since I'd be surprised if they would ever be used - it's not even clear whether they are supported by the mid-end since bitints are like bitfields and would require extra work.

design-documents/bit-precise-types.rst
mmalcomson: Ooh -- thanks!
I did not realise the cost of unaligned accesses were essentially free (useful to know).

design-documents/bit-precise-types.rst
mmalcomson: Yeah -- I had not realised the order of bullet points would imply a priority order.
The atomics were not supposed to be presented as the most important reason -- will adjust.

To make sure I'm not missing something -- what extra normalization step are you referring to?
Is it the normalisation required for the set of operations which are not OK for a given representation (e.g. if choosing option `B` there would be a normalisation required for `/`, but not `+`) or is it something else you've noticed?

design-documents/bit-precise-types.rst
Wilco1: Yes, optimization guides mention things like "The Neoverse N2 handles most unaligned accesses without performance penalties.", so you are unlikely to notice a penalty in typical compiled code.

design-documents/bit-precise-types.rst
```diff
@@ -162,14 +420,17 @@ It has the following negatives:
...
+- Operations ``/,%,==,<,>,<=,>=,>>`` and widening conversions on operands coming
+  from an ABI boundary would require masking the operands.
 
 - On AArch32 this could cause surprises to developers, given that on this
   architecture small Fundamental Data Types are have zero- or sign-extended
   extra bits.  So a ``char`` would not have the same representation as a
   ``_BitInt(8)`` on this architecture.
 
+- If used in calls to variadic functions which were written for standard
+  integral types this can give surprising results.
```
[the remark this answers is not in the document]
mmalcomson: Hmm, honestly I'd not seriously considered this since the way that the PCS is defined is through a two step process of
"source code type" -> "fundamental data type"
then
"fundamental data type" -> "rule for PCS passing".

Now you've pointed it out it does seem reasonable.
Not sure whether the exceptional case could lead to confusion -- will invite other feedback to see what people think.
@smithp35 @stuij @rsandifo-arm

design-documents/bit-precise-types.rst
Wilco1: Yes, I mean the standard normalization. So while you can't use LDADD with option C, you can still use a CAS or load exclusive loop with an extra zero/sign-extend. Since you can do this for every atomic operation, being able to use LDADD or LDSMAX directly might be nice but shouldn't count as an advantage. Not being able to use atomics at all would be a disadvantage for a particular layout if atomic support is essential (and I'm not convinced it is for bitint).

design-documents/bit-precise-types.rst
mmalcomson: :+1:  I'll add a paragraph explicitly mentioning that the compiler could always choose to use a different representation at non ABI boundaries.

design-documents/bit-precise-types.rst
algrant-arm: It's not just about the atomic RMW operations, you also (I think) want ordinary reads and writes of _BitInt variables to be single-copy-atomic and implemtable with ordinary loads and stores. This implies they should be aligned so as not to cross cache boundaries, and also not packed along with other objects (i.e. no packing a _BitInt(24) together with a _BitInt(8) in a 32-bit word).

design-documents/bit-precise-types.rst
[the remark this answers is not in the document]
mmalcomson: Test @rearnsha (wanted to tag you, the comment box doesn't want to auto-complete your name, don't know if typing it fully out will properly link -- this is a test).

stuij:
@rearnsha

mmalcomson:
Just pinging everyone that has been watching this.
Things seem to have settled on a decision here -- if there are still objections that I've missed could someone raise them again.

Otherwise could someone @rearnsha @smithp35 @stuij have a double-check that everything seems fine to them and approve this?

sallyarmneale (approved):
Looks good to me.

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+
+Background
+----------
+
+This document describes the rationale behind the ABI choices made for using the
+bit-precise integral types defined in C2x.  These are ``_BitInt(N)`` and
+``unsigned _BitInt(N)``.  These are defined for integral ``N`` and each ``N`` is
+a different type.
+
+The proposal for these types can be found in following link.
```
sallyarmneale, line 20: following link:

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+- We need to choose a representation, size and alignment of these objects in memory.
+
+The main trade-offs we have identified in this case are:
+
+- Performance of different C-level operations.
+- Whether certain hardware-level atomic operations are possible.
+- Size cost of storing values in memory.
+- General familiarity of programmers with the representation.
+
+Since this is a new type there is large uncertainty on how it will be used by
```
sallyarmneale, line 41: new type,

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+programmers in the future.  Decisions we make here may also influence future
+usage.  Nonetheless we must make trade-off decisions with this uncertainty.  The
```
sallyarmneale, line 43: Nonetheless -> Therefore,

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+below attempts to analyze possible use-cases to make our best guess as to how
+these types may be used when targeting Arm CPU's.
+
+
+Use-cases known of so far
+-------------------------
+
+There seem to be two different regimes for these types.  The "small" regime
```
sallyarmneale, line 51:
```suggestion
There are two different regimes for these types.  The "small" regime
```

design-documents/bit-precise-types.rst
sallyarmneale, line 43:
```suggestion
usage.  Therefore, we must make trade-off decisions with this uncertainty.  The
```

design-documents/bit-precise-types.rst
sallyarmneale, line 41:
```suggestion
Since this is a new type, there is large uncertainty on how it will be used by
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+necessary when performing operations, especially given that operations on small
+integral types promote their operands to ``int``.
+
+If these wider than necessary operations end up in the FPGA they would use many
+more many more logic gates than necessary.  Using ``_BitInt`` allows programmers
+to write code which directly expresses what is needed.  This can ensure the FPGA
+description generated saves space and has better performance.
+
+The notable thing about this use-case is that though the C code may be run on an
+Arm architecture (e.g. for testing), the most critical use is when transferred
```
sallyarmneale, line 75:
```suggestion
Arm architecture (for example for testing), the most critical use is when transferred
```

sallyarmneale: Avoid using Latin contractions

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+to an FPGA (i.e. not an Arm architecture).
+
+That said, if the operation that this FPGA performs becomes popular there may be
```
sallyarmneale, line 78:
```suggestion
However, if the operation that this FPGA performs becomes popular there may be
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+a need to run the code directly on CPU's in the future.
```
sallyarmneale, line 79:
```suggestion
a need to run the code directly on CPUs in the future.
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+
+The requirements on Arm ABI's from this use-case are relatively small since the
```
sallyarmneale, line 81:
```suggestion
The requirements on the Arm ABI from this use-case are relatively small since the
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+bit-fields rather than bit-precise types for storage, since the bit-precise types
+must be accessible and hence at least byte-aligned.
+
+The incentives to use bit-precise integral types for networking code would be in
+order to maintain the best representation of the operation that is being
+performed.
+
+One negative of using bit-precise integral types for networking code would be
+that idioms like ``if (x + y > max_representable)`` where ``x`` and ``y`` have
+been loaded from small bit-fields would no longer be viable.  We have seen such
```
sallyarmneale, line 122:
```suggestion
been loaded from small bit-fields, would no longer be viable.  We have seen such
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+arithmetic manipulation performed on such values will not be the main
+performance bottleneck in networking protocols.  This estimation comes from the
+belief that networking is often IO bound, and that small packed values in
+networking protocols tend to have limited arithmetic performed on them.
+
+Hence we believe that ease of debugging of values in registers may be more
+critical than performance concerns in this use-case.
+
+
+To help the compiler optimize (e.g. for auto vectorization)
```
sallyarmneale, line 137:
```suggestion
To help the compiler optimize (for example, for auto vectorization)
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+our trade-offs is correspondingly lower than others.
+
+Possible transparent BigNum libraries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We have heard of interest in using the new bit-precise integer types to
+implement transparent BigNum libraries in C.
+
+Such a use-case unfortunately does not directly correspond to what kind of code
+will be using this (e.g. would this be algorithmic code or I/O bound code).
```
sallyarmneale, line 225:
```suggestion
will be using this (for example, would this be algorithmic code or I/O bound code).
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+  appropriately shifted).
+
+- Operations ``+,-,%,==,<=,>=,<,>,<<`` all work without any extra instructions
+  (which is more of the common operations than other representations).
+
+It has the following negatives:
+
+- This would be a less familiar representation to programmers.  Especially the
+  fact that a ``_BitInt(8)`` would not have the same representation in a
+  register as a ``char`` could cause confusion (e.g. when debugging, or writing
```
sallyarmneale, line 414:
```suggestion
  register as a ``char`` could cause confusion (For example, when debugging, or writing
```

sallyarmneale (changes requested):
Some minor changes - mostly phrasing. Please do avoid using latin contractions (e.g., i.e, and etc) they're not good for non-native speakers.

mmalcomson:
> Please do avoid using latin contractions (e.g., i.e, and etc) they're not good for non-native speakers.

Noted -- thanks for the info!

design-documents/bit-precise-types.rst
sallyarmneale, line 20:
```suggestion
The proposal for these types can be found in following link:
```

design-documents/bit-precise-types.rst
```diff
@@ -0,0 +1,506 @@
...
+integral types promote their operands to ``int``.
+
+If these wider than necessary operations end up in the FPGA they would use many
+more many more logic gates than necessary.  Using ``_BitInt`` allows programmers
+to write code which directly expresses what is needed.  This can ensure the FPGA
+description generated saves space and has better performance.
+
+The notable thing about this use-case is that though the C code may be run on an
+Arm architecture (e.g. for testing), the most critical use is when transferred
+to an FPGA (i.e. not an Arm architecture).
```
sallyarmneale, line 76:
```suggestion
The notable thing about this use-case is that though the C code may be run on an
Arm architecture (for example. for testing), the most critical use is when transferred
to an FPGA (that is, not an Arm architecture).
```

design-documents/bit-precise-types.rst
sallyarmneale, line 137:
```suggestion
To help the compiler optimize (for example, for auto vectorization)
```

design-documents/bit-precise-types.rst
sallyarmneale, line 225:
```suggestion
will be using this (for example, would this be algorithmic code or I/O bound code).
```

sallyarmneale (changes requested):
Some minor changes - mostly phrasing. Please do avoid using latin contractions (e.g., i.e, and etc) they're not good for non-native speakers.

design-documents/bit-precise-types.rst
mmalcomson: Would like to not make this change -- I believe the change puts much more emphasis on the second point than it did before, and that's not emphasis that I believe should be there (at least that's what "However" sounds like to me).

Is this suggestion in order to make things easier to read for non-native speakers (as the suggestions to avoid latin contractions), or is it for some other reason?
(Would make the change if the point is ease-of-understanding for non-native speakers, reserve judgement for other motivations ;-).

mmalcomson:
@sallyarmneale Thanks for the reviews -- I've updated all the things you pointed at and upon the re-read found some places that I felt could be better worded as well.
N.b. to anyone else reading -- I also re-ordered the reasons in a few places.
A while ago Wilco pointed out that the ordering of reasons gave an impression that we weighted the behaviour around atomics higher than we actually did, and on re-reading it seems I had only re-ordered some of the mentions.

design-documents/bit-precise-types.rst
mmalcomson: I adjusted to "We believe there are two different regimes" (point being that we could be mistaken and want to convey that this is how we view things rather than exactly what is happening).

design-documents/bit-precise-types.rst
mmalcomson: Removed the `e.g.` indirection entirely -- just said "may be run on an Arm architecture for testing".

design-documents/bit-precise-types.rst
mmalcomson: Adjusted this to `(possibly for auto vectorization)` -- hope that isn't falling into another ease-of-parsing pitfall ;-)

design-documents/bit-precise-types.rst
mmalcomson: For some reason the "for example" reads clunkier than "e.g." to me.
I went with `cause confusion (we imagine when debugging or writing assembly code)`.
Hopefully that avoids the ease-of-parsing difficulty for non-native speakers.

sallyarmneale (commented):
Looks good to me

stuij (approved):

PR #192, opened 2023-02-21
[vfabia64] Replace hardcoded section number with link

avieira-arm:
This is just a minor fix to replace a hardcoded section number with a link to what I believe was the intended section it was originally pointing to.


mmalcomson:
Thanks!
To be slightly annoying -- do you mind doing the same for the mention of `section 4.1` and `chapter 4` in the `Parameter mapping (Advanced SIMD)` section just above where you made this change?

mmalcomson:
Looks good to me!  Thanks!

Issue #143, opened 2022-04-19
[ehabi32] Fix "trade off"


vhscampos:
"trade off" is not a correct term. It should be either "tradeoff" or "trade-off"

rearnsha:
Well that depends on the context of its use.  If it's used as a verb then "trade off" is clearly correct.  But if used as a noun then you're correct.

vhscampos:
It's used as a noun. It's in the second paragraph of Section 3.

Issue #142, opened 2022-04-19
[ehabi32] "toolchain" should be consistent


vhscampos:
"toolchain" is not consistent across the document. Sometimes it's "tool-chain", others it's "tool chain".

It should be made consistent.

PR #193, opened 2023-03-07
Spelling canonicalisation

mmalcomson:
Issues #143 and #142 are both simple typographic changes to make.

Here I suggest we choose "trade-off" for #143 and "toolchain" for #142 , but perfectly happy to adjust if there are other opinions.

"toolchain" was not the existing most common form, but I think that's the most commonly used term -- at least searching the web for the other forms gives pages using "toolchain".


vhscampos (approved):
LGTM. Thanks

PR #199, opened 2023-04-06
Mention size and alignment of "place" for some relocations

mmalcomson:
Most relocations act on immediates of instructions and specify the bits of the calculated value that are put into that instruction.

The "Static Data relocations" table is for data relocations which do not act on instructions.  These seem to have not explicitly mentioned the number of bits written by the linker.  They also do not mention the alignment of the "place" that is written (something which is written in the aaelf32 document).

Here we mention both the above in the "Comment" field of the relocation tables.  This differs from where the aaelf32 document describes the size and alignment of static data simply because I felt this way would be easier to find the information and would follow the existing style in the document better.
Am happy to change that if requested.

N.b. I have not seen the GOTREL relocations implemented in any linker implementation and so have not been able to double-check my understanding of the behaviour matches that which tools implement. Hence am going based on the information I've heard.  Would appreciate special attention to this part to make sure that they are indeed designed to be byte-aligned.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1001,20 +1001,27 @@ See also table `GOT-relative data relocations`_.
   | ELF64   | ELF32   | Name                 | Operation | Comment                                                          |
   | Code    | Code    |                      |           |                                                                  |
   +=========+=========+======================+===========+==================================================================+
-  | 257     | \-      | R\_<CLS>\_ABS64      | S + A     | No overflow check                                                |
+  | 257     | \-      | R\_<CLS>\_ABS64      | S + A     | Write bits [63:0] of X at byte-aligned place P.                  |
+  |         |         |                      |           | No overflow check.                                               |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 258     | 1       | R\_<CLS>\_ABS32      | S + A     | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`                     |
+  | 258     | 1       | R\_<CLS>\_ABS32      | S + A     | Write bits [31:0] of X at byte-aligned place P.                  |
+  |         |         |                      |           | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 259     | 2       | R\_<CLS>\_ABS16      | S + A     | Check that -2\ :sup:`15` <= X < 2\ :sup:`16`                     |
+  | 259     | 2       | R\_<CLS>\_ABS16      | S + A     | Write bits [15:0] of X at byte-aligned place P.                  |
+  |         |         |                      |           | Check that -2\ :sup:`15` <= X < 2\ :sup:`16`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 260     | \-      | R\_<CLS>\_PREL64     | S + A - P | No overflow check                                                |
+  | 260     | \-      | R\_<CLS>\_PREL64     | S + A - P | Write bits [63:0] of X at byte-aligned place P.                  |
+  |         |         |                      |           | No overflow check.                                               |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 261     | 3       | R\_<CLS>\_PREL32     | S + A - P | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`                     |
+  | 261     | 3       | R\_<CLS>\_PREL32     | S + A - P | Write bits [31:0] of X at byte-aligned place P.                  |
+  |         |         |                      |           | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -1001,20 +1001,25 @@ See also table `GOT-relative data relocations`_.
...
+  |         |         |                      |           | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 262     | 4       | R\_<CLS>\_PREL16     | S + A - P | Check that -2\ :sup:`15` <= X < 2\ :sup:`16`                     |
+  | 262     | 4       | R\_<CLS>\_PREL16     | S + A - P | Write bits [0:16] of X at byte alignment.                        |
+  |         |         |                      |           | Check that -2\ :sup:`15` <= X < 2\ :sup:`16`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 314     | 29      | R\_<CLS>\_PLT32      | S + A - P | Check that -2\ :sup:`31` <= X < 2\ :sup:`31` see call and jump   |
-  |         |         |                      |           | relocations                                                      |
+  | 314     | 29      | R\_<CLS>\_PLT32      | S + A - P | Write bits [0:31] of X at byte alignment.                        |
```
smithp35, line 1020: In this case the PLT32 is designed for relative vtable entries which are always 4-byte aligned. If we enforced that then we could enforce that the result was also 4-byte aligned. However that would tie the relocation to its current use case.

smithp35 (commented):
The current wording, example "Write bits [0:63] of X at byte alignment" is a bit ambiguous as to whether we are referring to the "place" we write the relocation value, or the alignment of the result of the relocation calculation, "X" in this case.

Not sure how best to word this in the comment without being overly long. This might imply that a separate "Unless otherwise specified there are no alignment requirements for the place P for data relocations."

In the comment, perhaps "Place is byte aligned."

I think the GOTREL relocations are described correctly. The closest equivalent in x86_64 is GOTOFF64 which is described (https://eli.thegreenplace.net/2012/01/03/understanding-the-x64-code-models) as being part of the large code model. In this case it can be loading a byte at some offset from the GOT so I don't think alignment requirements are needed.

aaelf64/aaelf64.rst
mmalcomson: Thanks -- I didn't have a great handle on where this is usually used.
Based on what you said I personally would prefer to not add that restriction based on leaving flexibility for other use-cases.
Do you have a preference?
(I read your comment as mentioning possibilities rather than suggesting a change -- so didn't notice a preference in it).

FWIW the current implementation in `lld` (`ld.bfd` does not seem to implement this) allows 1-byte aligned values.

aaelf64/aaelf64.rst
```diff
@@ -1001,20 +1001,25 @@ See also table `GOT-relative data relocations`_.
...
-  | 260     | \-      | R\_<CLS>\_PREL64     | S + A - P | No overflow check                                                |
+  | 260     | \-      | R\_<CLS>\_PREL64     | S + A - P | Write bits [0:63] of X at byte alignment.  No overflow check.    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 261     | 3       | R\_<CLS>\_PREL32     | S + A - P | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`                     |
+  | 261     | 3       | R\_<CLS>\_PREL32     | S + A - P | Write bits [0:31] of X at byte alignment.                        |
+  |         |         |                      |           | Check that -2\ :sup:`31` <= X < 2\ :sup:`32`.                    |
   +---------+---------+----------------------+-----------+------------------------------------------------------------------+
-  | 262     | 4       | R\_<CLS>\_PREL16     | S + A - P | Check that -2\ :sup:`15` <= X < 2\ :sup:`16`                     |
+  | 262     | 4       | R\_<CLS>\_PREL16     | S + A - P | Write bits [0:16] of X at byte alignment.                        |
```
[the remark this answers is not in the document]
mmalcomson: Ah, yes, thanks ;-)

mmalcomson:
> The current wording, example "Write bits [0:63] of X at byte alignment" is a bit ambiguous as to whether we are referring to the "place" we write the relocation value, or the alignment of the result of the relocation calculation, "X" in this case.
> 
> Not sure how best to word this in the comment without being overly long. This might imply that a separate "Unless otherwise specified there are no alignment requirements for the place P for data relocations."
> 
> In the comment, perhaps "Place is byte aligned."

I've tried swapping "at byte alignment" to "at byte-aligned place P" attempting to re-use the term `P` that was defined in `Relocation operations`.

Does that look good to you?

algrant-arm:
Just an observation, but the architecture when describing bit fields, puts the high bit first e.g. x[15:0]. This includes some instruction fields which might be updated by relocations.

mmalcomson:
> Just an observation, but the architecture when describing bit fields, puts the high bit first e.g. x[15:0]. This includes some instruction fields which might be updated by relocations.

And so are all the existing descriptions of bit fields used in this document that I was trying to emulate.
Thanks for the catch!

aaelf64/aaelf64.rst
smithp35: I'm happy to leave as is. AFAIK the only production use of PLT32 is in Fuchsia for relative vtables so probably not a high priority for ld.bfd.

PR #204, opened 2023-04-25
[aaelf64] Expand STO_AARCH64_VARIANT_PCS documentation

rsandifo-arm:
The intention here is to generalise the STO_AARCH64_VARIANT_PCS documentation slightly to include “processor state”, not just register contents.  I've also explicitly listed some types of functions that need the marker.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -18,6 +18,7 @@
 .. _SYM-VER: http://www.akkadia.org/drepper/symbol-versioning
 .. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
 .. _MTEEXTENSIONS: https://www.kernel.org/doc/html/latest/arm64/memory-tagging-extension.html#core-dump-support
+.. _VFABI64: https://github.com/ARM-software/abi-aa/releases
 
 ELF for the Arm® 64-bit Architecture (AArch64)
 **********************************************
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -684,14 +685,54 @@ The defined processor-specific ``st_other`` flag values are listed below:
...
+
+* The function might guarantee that extra register state is preserved
+  by the call.
+
+* PSTATE on entry to the function might be different from normal.
+
+* PSTATE on return from the function might be different from normal.
+
+The following types of function must be marked with
+``SOT_AARCH64_VARIANT_PCS``, although the list is not intended
```
sdesmalen-arm, line 713: nit: `STO_AARCH64_VARIANT_PCS` (not `SOT_...`)

sdesmalen-arm (approved):
I don't know if I'm the right person to approve the patch, but I'm happy to see the meaning of STO_AARCH64_VARIANT_PCS strengthened for state other than callee-saved/argument/return registers, which is needed for SME.

aaelf64/aaelf64.rst
```diff
@@ -684,14 +685,54 @@ The defined processor-specific ``st_other`` flag values are listed below:
...
+
 If a subroutine is called via a symbol reference that is marked with
 ``STO_AARCH64_VARIANT_PCS``, then code that runs between the calling routine
 and the called subroutine must preserve the contents of all registers except
-for IP0, IP1, and the condition code flags [AAPCS64_].
+for IP0, IP1, and the condition code flags [AAPCS64_].  It must also
+preserve the processor execution mode, such as PSTATE.SM and PSTATE.ZA.
+It is not possible to provide a definitive list of which state must be
+preserved, since the intention is to allow it to grow as the architecture
+evolves.
```
mmalcomson, line 735: To ask a naive question:   Do we know if there's any way in which processor execution mode was changed in such intermediate code before?

I.e. while we're calling out `PSTATE.{SM,ZA}` it seems like we're also putting a new restriction on all state in some class "processor execution mode", and TBH I don't know what that class encompasses.

My first guess for what that class of state encompasses would be `PSTATE.{SM,ZA}` based on context.
But without context I would have guessed the PSTATE field group `execution state controls` as mentioned in the Arm ARM (couldn't find what "field group" the SM and ZA would end up belonging to).

mmalcomson: What do you think about including something like "at the time of writing this entails `PSTATE.SM` and `PSTATE.ZA`" rather than "such as `PSTATE.SM` and `PSTATE.ZA`"?

Agreed we can't make it fool-proof, just trying to make it a bit more definitive about what we understand the situation to be right now.

rsandifo-arm: I'd rather avoid putting an upper bound on what's needed, since the point is that any current upper bound isn't guaranteed to be permanent.  I've also had pushback in the past about using “at the time of writing”, since that time is more obvious to the writer than the reader.

I think we should focus not on what the list currently is (for PSTATE or for registers), but instead on explaining why certain examples are included.  If it isn't obvious from the current text why PSTATE.SM and PSTATE.ZA are included, I can try to make it clearer.

mmalcomson: I can see your point -- I was hesitant about using "at the time of writing" too (though I still prefer it to the existing text).

Honestly when reading this I imagined state that I'm unaware of which might start to be relied upon for a users own special functions yet our runtime clobbers because we didn't think it was relevant.

I didn't originally pick up on the viewpoint of "if it's not mentioned by the `PCS` then it's either irrelevant to the `VARIANT_PCS` or the `PCS` has not been updated yet" at all.

That said -- this is an argument based on my lack of knowledge rather than anything else, so if you're confident the pitfall I'm concerned about is not a possibility then I won't push further than the suggestions below ;-)


Few final attempts at trying to avoid that particular misread (but if you don't like them then fair enough):

Maybe a lower bound instead of an upper bound -- could we explicitly exclude all `PSTATE` bits up to the architecture version at which SME was introduced?

Alternatively, trying to focus on the why -- maybe something like modifying the last sentence here to "the intention is to allow it to grow alongside the PCS specification as the architecture evolves"?

Issue #158, opened 2022-07-05
[Morello] Proposal: New ABI for functions with more than 8 arguments


dpgao:
When a C function takes more than 8 arguments, the current Morello ABI [requires](https://github.com/ARM-software/abi-aa/blob/main/aapcs64-morello/aapcs64-morello.rst#parameter-passing) these extra arguments to be 'spilled' onto the stack and passed implicitly to the callee.

Before the introduction of the new varargs ABI (#112), variadic arguments were also passed via the stack. Now, they are passed via a buffer (which may or may not be allocated on the stack) pointed to by `c9`. Thus, variadic functions no longer need to scan through the stack to retrieve its variadic arguments, which used to be a source of spatial safety violations.

I propose that the spilled arguments also be passed via this buffer, thereby completely eliminating any implicit argument passing via the stack. Here's a sketch of the implementation:

1. The spilled arguments are put at the front of the buffer.
2. Padding space is inserted.
3. Variadic arguments (if there're any) are put after the padding.
4. Since the types of the spilled arguments are known statically, the offset at which the variadic arguments begin is also known statically.
5. The `va_start` macro can thus be implemented to correctly set the `va_list` capability.

The rationale for this proposal is as follows. The current ABI necessarily exposes the caller's stack pointer to the callee. Under situations where the callee is untrusted code, this poses a security risk. The proposed ABI does not mandate that the argument buffer be allocated on the stack, allowing flexibility in its location depending on the level of security needed. For example, if the callee is trusted by the caller, the buffer may well be allocated on the stack. Otherwise, the buffer can be allocated on a single-use page that is deallocated by the caller as soon as the callee returns.

# Technicalities

1. When the `va_start` macro sets the `va_list` capability, it would be desirable to have that capability's lower bound increased _past_ the region containing the spilled arguments. A carefully designed padding scheme for step 2 above will be needed to make sure that this is the case despite any address rounding.

sbaranga-arm:
Just trying to gather more information about what we need this for (not saying this shouldn't be possible but this would be an ABI break).
For compartment calls we would have more registers besides SP that shouldn't be exposed (any register not used for argument passing, some system registers etc). For clearing argument registers we need a different calling convention. which could be made to pass arguments via c9 instead of the stack.
So we only need to change the way we're passing arguments for non-compartment calls if there is a requirement that compartment calls pass arguments the same way as non-compartment calls?

dpgao:
Hello,

Thanks for your questions.

[...]

PR #205, opened 2023-06-08
[Morello] Change the PCS to pass memory arguments through C9 instead of passing them on the stack

sbaranga-arm:
In practice for non-compartment calls the arguments are still passed on the stack, although now there is no explicit requirement for this.

Variadic arguments are still passed through C9 in the same way, at the first 16-byte aligned location after the last non-variadic argument.

This is done is such a way that reading more variadic arguments than were passed is guaranteed to cause an out-of-bounds fault.

Implements https://github.com/ARM-software/abi-aa/issues/158.


aapcs64-morello/aapcs64-morello.rst
```diff
@@ -521,22 +526,22 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
...
-    |                              | If the callee is variadic in AAPCS64-cap, the Anonymous Arguments Memory area is         |
-    |                              | allocated in memory with a size of at least 16 * (number of Anonymous arguments) bytes.  |
-    |                              | The Anonymous Arguments Memory area is 16-byte aligned. The Anonymous Arguments          |
-    |                              | Capability points to the Anonymous Arguments Memory area and is copied to C9. If there   |
-    | A.4                          | are no Anonymous arguments passed, C9 is set to NULL.                                    |
+    |                              | If any parameters are memory passed in AAPCS64-cap, the Arguments Memory Area is         |
+    |                              | allocated.                                                                               |
+    |                              | The Arguments Memory area is 16-byte aligned. The Arguments Capability has the           |
+    |                              | bounds of the Arguments Memory area and Arguments Memory area and the address of the     |
```
kevin-brodsky-arm, line 532: "and Arguments Memory area" is repeated.

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -808,7 +815,7 @@ The ``va_list`` type may refer to any parameter in a parameter list. All Anonymo
 The va_start() macro
 --------------------
 
-The ``va_start`` macro shall initialize the ``va_list`` argument to the value of C9 as seen in the entry of the callee.
+The ``va_start`` macro shall initialize the ``va_list`` argument to the value of derived from C9 with the address of the first Anonymous argument.
```
kevin-brodsky-arm, line 818: s/of derived/derived/
It also feels strange to say *the* value derived from [...], rather *a* value? Using "value" feels a bit confusing too as it seems to refer to the capability address in other places.

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -521,22 +526,22 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
...
-    |                              |                                                                                          |
-    | A.5                          |                                                                                          |
+    |                              | The next memory argument address (NMAA) is set to the value of C9 in AAPCS64-cap or the  |
+    | A.4                          | current stack-pointer value (SP) otherwise.                                              |
     +------------------------------+------------------------------------------------------------------------------------------+
 
 
+.. note::
+   The Arguments Capability is a capability that has the bounds of the arguments memory area and its address will be that of the first memory passed argument. The address of the Arguments Capability is 16-byte aligned. The capability offset of the Arguments Capability may not be zero. The capability offset is such that the first 16-byte aligned location after the last memory passed argument will be larger or equal than capability limit of the Arguments Capability.
```
kevin-brodsky-arm, line 542: "larger than or equal to the capability limit..."

I wonder if this paragraph shouldn't start by explaining how the Arguments Memory Area is defined. I don't really understand how it all fits in terms of bounds representability right now.

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -674,26 +679,28 @@ The differences in language bindings used for AAPCS64 and AAPCS64-cap are descri
...
     |                               | size of the argument. The argument has now been allocated.                             |
     | C.18                          |                                                                                        |
     +-------------------------------+----------------------------------------------------------------------------------------+
 
 .. note::
+    In AAPCS64-cap all Anonymous arguments are memory passed.
 
-    In AAPCS64-cap if the callee is variadic and there are fewer than 4096 Anonymous arguments, the length of C9 divided by 16 is equal to the number of Anonymous arguments. The length of C9 divided by 16 is always greater or equal to the number of Anonymous arguments.
+.. note::
+    In AAPCS64-cap if the callee is variadic the number of Anonymous arguments is equal to (length(C)-offset(C)/16) where C is the adjusted value of C9 that has the address of the first Anonymous Argument.
```
dpgao, line 703: Should this be `(length(C)-offset(C))/16` instead?

sbaranga-arm: Yes, thanks for catching that.

aapcs64-morello/aapcs64-morello.rst
jacobbramley, line 818: Rather than C9, consider referring to "the Arguments Capability". A compiler might need to spill it (or move it to something callee-saved) before `va_start` is called.

aapcs64-morello/aapcs64-morello.rst
jacobbramley: I agree about the precision issues, but also it'd be very useful to include a diagram here. I don't know if we have the infrastructure for that in this document, but it would really help.

I'm a bit confused about the relationship between the offset and bounds at various points. For example, here the "address will be that of the first memory passed argument", but also its offset may not be zero. However, won't the first memory-passed argument be allocated to the first slot in the arguments memory area, and therefore require an offset of zero?

It's likely that I've just misunderstood something obvious!

sbaranga-arm: Thanks! I'll try to clarify this in some way and look at adding a diagram. We need to add some padding at the start of the arguments area to make the length representable. So the 'offset' (maybe this should be 'capability offset'?) is just the size of the padding, and the first argument (and first slot) is located at the address (capability value in Morello terms) of C9.

We don't say how the (capability) offset is computed and leave that up to the implementation. In practice the way to do this would be to run the algorithm once to get the number of bytes that we need for argument passing and compute from that the minimal amount of padding required to make everything representable. With the current wording we can have multiple legal values of offset, but just one being minimal/optimal.

Hopefully this clarifies things?

jacobbramley: Ok, that does clarify things, but couldn't the offset still be zero in the (common) case where the area is precisely representable without padding?

sbaranga-arm: At the moment any sane implementation should give a zero offset in this case. We could add the additional constraint that the offset needs to be minimal, which would then require a zero offset. Any thoughts?

aapcs64-morello/aapcs64-morello.rst
```diff
@@ -496,7 +496,25 @@ The A64 and C64 branch instructions are unable to reach every destination in the
...
+Arguments Capability
+^^^^^^^^^^^^^^^^^^^^
+
+In AAPCS64-cap if there are any memory-passed parameters the Arguments Memory Area is allocated.
+
+The Memory Arguments Length (MAL) is the number of bytes required for memory-passed arguments. This can be obtained by subtracting the final and initial values of NMAA as part of Stage C from `Parameter Passing Rules`_.
+
+The Rounded Memory Arguments Length (RMAL) is the least representable capability length greater or equal to MAL.
+
+The size of the Arguments Memory Area is equal to RNML. The minimal alignment of the Arguments Memory Area is 16 bytes. The alignment of the Arguments Memory Area is large enough to allow the bounds of the Arguments Memory Area to be exactly representable by a capability.
```
kevin-brodsky-arm, line 515: I can't find the defiinition of RNML, could it be RMAL?

sbaranga-arm: Thanks! Yes, it should have been RMAL.

Issue #197, opened 2023-03-27
[aaelf64] BTI PLT requirement should be weaker


nsz-arm:
text says

"To support Branch Target Identification mechanism, in the presence of a GNU_PROPERTY_AARCH64_FEATURE_1_BTI all PLT entries generated by the linker must have a BTI instruction as the first instruction."

however in shared libraries PLT entries are not called indirecty, except when the linker inserts a veneer with indirect jump (for far calls), but the linker clearly knows about that case so it can either emit PLT differently then or do the veneer differently (double veneer).

in practice bfd ld does not emit BTI PLT in shared libraries only in executables (where PLT maybe the canonical address and thus indirect calls to it is possible).

i think the text should say that "If a PLT entry may be called indirectly then all PLTs must start with BTI c" (but a linker may choose to ensure that PLTs are never called indirectly). The reason all PLTs must be consistent is only to have same PLT size, so dynamic linker knows how to compute the index of the related GOT PLT entry, so the requirement may be even weaker depending on what `DT_AARCH64_BTI_PLT` should mean and if we want to allow tools to disas the first PLT entry and make assumptions about other PLT entries based on that.

smithp35:
"If a PLT entry may be called indirectly then all PLTs must start with BTI c" sounds reasonable to me.

From memory the v4t Arm PLT entries optionally inserted a BX PC, NOP into specific PLT entries called from Thumb, but not from Arm although I could have that wrong.

In LLD at least, the way the offsets are calculated means that all PLT entries need to be the same size. This is an implementation   constraint not a specification constraint though.

mmalcomson:
Addressed by https://github.com/ARM-software/abi-aa/pull/212 so closing here.

PR #212, opened 2023-07-21
Relax BTI PLT requirement in aaelf64

smithp35:
The ABI states that all PLT entries must start with BTI when the DT_AARCH64_BTI_PLT dynamic tag is present. This is too strict as the static linker can prove that some PLT entries are never indirectly called so do not need to start with a BTI. Both LLD and GNU ld do this.

Relax the wording to bring the specification in line with implementations.

https://github.com/ARM-software/abi-aa/issues/197


mmalcomson:
FWIW I'm for this change going in.
Mentioning to be explicit since we recently discussed in-person whether to double-check the existing linker implementations are indeed doing the right thing in this area as part of updating the spec.

DanielKristofKiss (approved):
LGTM

PR #214, opened 2023-08-03
[PAuthABI] Change DT_AARCH64_AUTH_RELR* to avoid conflict with DT_AARCH64_VARIANT_PCS

MaskRay:
We need to pick two odd numbers for DT_AARCH64_AUTH_RELRSZ and DT_AARCH64_AUTH_RELRENT. The recent DT_AARCH64_MEMTAG_* tags have occupied a few numbers.
Let's use 0x70000011 to 0x70000013. The numbers are consecutive to be similar to DT_RELRSZ/DT_RELR/DT_RELRENT.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -231,6 +231,9 @@ changes to the content of the document for that release.
   | 2021q3     | 1\ :sup:`st` November 2021  | Clarify the concepts of a default and alternate PAuth elf signing|
   |            |                             | schemes.                                                         |
   +------------+-----------------------------+------------------------------------------------------------------+
+  | 2023q3     | 16\ :sup:`th` August 2023   | Update tags in `Dynamic Section`_ to avoid conflict with         |
+  |            |                             | DT_AARCH64_VARIANT_PCS.                                          |
+  +------------+-----------------------------+------------------------------------------------------------------+
 
 References
 ----------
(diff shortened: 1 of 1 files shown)


smithp35 (commented):
Change looks good to me, it doesn't clash with the tags in https://github.com/ARM-software/abi-aa/blob/main/memtagabielf64/memtagabielf64.rst#6dynamic-section. I've raised https://github.com/ARM-software/abi-aa/issues/215 to cover text to reserve the values in aaelf64.rst

Even though the change is small. Please could you update the release history table at https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#212change-history 

I suggest using 2023q3 as the release.

MaskRay:
> Even though the change is small. Please could you update the release history table at https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#212change-history

Thanks for the suggestion. Done.

pauthabielf64/pauthabielf64.rst
```diff
@@ -231,6 +231,9 @@ changes to the content of the document for that release.
   | 2021q3     | 1\ :sup:`st` November 2021  | Clarify the concepts of a default and alternate PAuth elf signing|
   |            |                             | schemes.                                                         |
   +------------+-----------------------------+------------------------------------------------------------------+
+  | 2023q3     | 16\ :sup:`st` August 2023   | Update DT_AARCH64_AUTH_ values to avoid conflict with            |
```
smithp35, line 234: The DT_AARCH64_AUTH_ (I think it is the trailing _ in DT_AARCH64_AUTH_ ) is causing the word to be interpreted as a link which doesn't exist. This is then failing the rstcheck.

I suggest: Update tags in `Dynamic Sections`_ to avoid conflict with DT_AARCH64_VARIANT_PCS.

smithp35 (changes requested):
The CI rstcheck is failing for this file. The output has a lot of noise so it is difficult to see, I found searching for pauthabi64 brought up the specific error. If there are no errors in this specific document the CI will pass.

smithp35:
My apologies it looks like it should have been 
```
`Dynamic Section`_ 
``` 
to match the section heading later on. I mistyped and used Dynamic Sections.

The bit the link should match is:
```
Dynamic Section
===============
```

smithp35:
Thanks, I've merged the commit. I've also raised https://github.com/ARM-software/abi-aa/pull/216 to reserve the values used in the ELF extensions.

Issue #217, opened 2023-08-22
static linkers (lld and GNU ld) out of sync with aaelf64 for GOT relocations with addends.


smithp35:
Raised due to LLVM issue https://github.com/llvm/llvm-project/issues/63418 

Raising this as an ABI issue rather than a two separate toolchain issues as it may be simpler to amend the ABI than try and fix the tools.

When doing a `:got:` and `:got_lo12:` expression to a local symbol, the GNU and LLVM assemblers convert this to a GOT generating relocation to the section symbol + addend. 

For example:
```
        .global main
        .type   main, %function
main:
        adrp    x0, :got:x
        ldr     x0, [x0, #:got_lo12:x]
        adrp    x1, :got:y
        ldr     x1, [x1, #:got_lo12:y]
        adrp    x2, :got:z
        ldr     x2, [x2, #:got_lo12:z]
        add     x0, x1, x1
        add     x0, x0, x2
        ret
        .size   main, .-main
        .bss
        .align  3
        .size   x, 4
        .size   y, 8
        .size   z, 8
x:
        .zero   4
y:
        .zero 8
z:
        .zero 8
```
has the following relocations.
```
Relocation section '.rela.text' at offset 0x128 contains 6 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000000000  0000000500000137 R_AARCH64_ADR_GOT_PAGE 0000000000000000 .bss + 0
0000000000000004  0000000500000138 R_AARCH64_LD64_GOT_LO12_NC 0000000000000000 .bss + 0
0000000000000008  0000000500000137 R_AARCH64_ADR_GOT_PAGE 0000000000000000 .bss + 4
000000000000000c  0000000500000138 R_AARCH64_LD64_GOT_LO12_NC 0000000000000000 .bss + 4
0000000000000010  0000000500000137 R_AARCH64_ADR_GOT_PAGE 0000000000000000 .bss + c
0000000000000014  0000000500000138 R_AARCH64_LD64_GOT_LO12_NC 0000000000000000 .bss + c
```

The ABI https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#576static-aarch64-relocations search for GOT-relative instruction relocations

Defines these to be:
* Page(G(GDAT(S+A)))-Page(P)
* G(GDAT(S+A))

With
```
[...]

PR #223, opened 2023-09-28
[aaelf64] Define GOT-Relative data relocation

PiJoules:
This introduces a new relocation `R\_<CLS>\_GOTPCREL32` which follows the existing wording to “R\_<CLS>\_GOTREL32”, but instead evaluates to the 32-bit offset between a GOT entry for a given symbol and the current location where the relocation is applied, so its equation would be “G(GDAT(S+A))- P”.


aaelf64/aaelf64.rst
```diff
@@ -1240,14 +1240,17 @@ The following tables record single instruction relocations and relocations that
...
+  | 307        | \-         | R\_<CLS>\_GOTREL64   | S+A-GOT          | Write bits [63:0] of X at byte-aligned place P.  This represents a 64-bit offset relative to the GOT.                   |
+  +------------+------------+----------------------+------------------+-------------------------------------------------------------------------------------------------------------------------+
+  | 308        | \-         | R\_<CLS>\_GOTREL32   | S+A-GOT          | Write bits [31:0] of X at byte-aligned place P.  This represents a 32-bit offset relative to GOT, treated as signed;    |
+  |            |            |                      |                  | Check that -2\ :sup:`31` <= X < 2\ :sup:`31`.                                                                           |
+  +------------+------------+----------------------+------------------+-------------------------------------------------------------------------------------------------------------------------+
+  | 315        | \-         | R\_<CLS>\_GOTPCREL32 | G(GDAT(S+A))- P  | Write bits [31:0] of X at byte-aligned place P.  This represents a 32-bit offset relative to GOT entry for an address, treated as signed;    |
```
smithp35, line 1251: Although difficult to see unless you have a wide editor the Comment column is too long and the RST treats this as a malformed table.

I think 
```
  | 315        | \-         | R\_<CLS>\_GOTPCREL32 | G(GDAT(S+A))- P  | Write bits [31:0] of X at byte-aligned place P.  This represents a 32-bit offset relative to GOT entry for an address,  |
  |            |            |                      |                  | treated as signed; Check that -2\ :sup:`31` <= X < 2\ :sup:`31`.                                                        |
```
Will fix it. 

There's a script in tools/common/check-rst-syntax.sh which will run rstcheck over all the documents. Will be worth running that first to check.

smithp35 (commented):
The change text looks good to me, the CI is complaining about the table formatting. I've made a suggestion for how to correct that.

smithp35 (approved):
Thanks for the update, it passes CI now.

I've marked this as approved as we have got spare relocation codes and it does help make relative vtables be more efficient on AArch64.

Will wait for a colleague to double check and merge.

mmalcomson:
LGTM as well -- merging as I'm the second to look at it.

MaskRay:
Thanks for the addition. I think another advantage introducing the data relocation is that: `-shared -fpic -fexperimental-relative-c++-abi-vtables` builds will not create PLT entries for functions in the vtables, as long as they are not directly called.

smeenai:
I don't understand the equation for this relocation `G(GDAT(S+A)) - P`. Section 5.7.33 Relocation operations says:

> `GDAT(S+A)` represents a pointer-sized entry in the `GOT` for address `S+A`. The entry will be relocated at run time
with relocation `R_<CLS>_GLOB_DAT(S+A)`

Based on the LLD implementation of this relocation though (https://github.com/llvm/llvm-project/pull/72584), it seems the addend is applied to the place of the relocation, not the GOT entry. See [the added test case](https://github.com/PiJoules/llvm-project/blob/f0b271e4ebd2180f497694838ec9db0b2fd8ab4b/lld/test/ELF/aarch64-reloc-gotpcrel32.s#L15-L22) for what I mean. The relocations in the object file are:

```
$ bin/llvm-readelf -rW tools/lld/test/ELF/Output/aarch64-reloc-gotpcrel32.s.tmp.o

Relocation section '.rela.data' at offset 0xd0 contains 5 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000000000  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar + 0
0000000000000004  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar + 4
0000000000000008  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar - 4
000000000000000c  000000040000013b R_AARCH64_GOTPCREL32   0000000000000000 baz + ffffffff
0000000000000010  000000040000013b R_AARCH64_GOTPCREL32   0000000000000000 baz - ffffffff
```

The relocations in the output SO are:

```
$ bin/llvm-readelf -rW tools/lld/test/ELF/Output/aarch64-reloc-gotpcrel32.s.tmp.so

Relocation section '.rela.dyn' at offset 0x2c0 contains 2 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000020398  0000000100000401 R_AARCH64_GLOB_DAT     0000000000000000 baz + 0
0000000000020390  0000000200000401 R_AARCH64_GLOB_DAT     00000000000303a0 bar + 0
```

So the addend isn't propagated to the dynamic relocations, which `GDAT(S+A)` would imply in my reading. Instead, I believe the equation should be:

```
G(GDAT(S)) + A - P
```

This also better matches `R_X86_64_GOTPCREL`, which is defined as `G + GOT + A - P`, where `G` is the GOT offset for the symbol and `GOT` is the GOT address. I belive `R_ARM_GOT_PREL` is the AArch32 equivalent, and that's also defined as `GOT(S) + A - P`, where `GOT(S)` is the address of the GOT entry for S. Am I missing something?

PiJoules:
> I don't understand the equation for this relocation `G(GDAT(S+A)) - P`. Section 5.7.33 Relocation operations says:
> 
> > `GDAT(S+A)` represents a pointer-sized entry in the `GOT` for address `S+A`. The entry will be relocated at run time
> > with relocation `R_<CLS>_GLOB_DAT(S+A)`
> 
> Based on the LLD implementation of this relocation though ([llvm/llvm-project#72584](https://github.com/llvm/llvm-project/pull/72584)), it seems the addend is applied to the place of the relocation, not the GOT entry. See [the added test case](https://github.com/PiJoules/llvm-project/blob/f0b271e4ebd2180f497694838ec9db0b2fd8ab4b/lld/test/ELF/aarch64-reloc-gotpcrel32.s#L15-L22) for what I mean. The relocations in the object file are:
> 
> ```
> $ bin/llvm-readelf -rW tools/lld/test/ELF/Output/aarch64-reloc-gotpcrel32.s.tmp.o
> 
> Relocation section '.rela.data' at offset 0xd0 contains 5 entries:
>     Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
> 0000000000000000  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar + 0
> 0000000000000004  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar + 4
> 0000000000000008  000000020000013b R_AARCH64_GOTPCREL32   0000000000000000 bar - 4
> 000000000000000c  000000040000013b R_AARCH64_GOTPCREL32   0000000000000000 baz + ffffffff
> 0000000000000010  000000040000013b R_AARCH64_GOTPCREL32   0000000000000000 baz - ffffffff
> ```
> 
> The relocations in the output SO are:
> 
> ```
> $ bin/llvm-readelf -rW tools/lld/test/ELF/Output/aarch64-reloc-gotpcrel32.s.tmp.so
> 
> Relocation section '.rela.dyn' at offset 0x2c0 contains 2 entries:
>     Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
> 0000000000020398  0000000100000401 R_AARCH64_GLOB_DAT     0000000000000000 baz + 0
> 0000000000020390  0000000200000401 R_AARCH64_GLOB_DAT     00000000000303a0 bar + 0
> ```
> 
> So the addend isn't propagated to the dynamic relocations, which `GDAT(S+A)` would imply in my reading. Instead, I believe the equation should be:
> 
> ```
> G(GDAT(S)) + A - P
> ```
> 
> This also better matches `R_X86_64_GOTPCREL`, which is defined as `G + GOT + A - P`, where `G` is the GOT offset for the symbol and `GOT` is the GOT address. I belive `R_ARM_GOT_PREL` is the AArch32 equivalent, and that's also defined as `GOT(S) + A - P`, where `GOT(S)` is the address of the GOT entry for S. Am I missing something?

No you're correct. The addend should be applied to where the relocation is located, not to symbol we're taking the GOT entry for. The relocation is intended to be similar to `R_X86_64_GOTPCREL` as you said. I'll send out a followup fix. Sorry for the confusion.

PiJoules:
https://github.com/ARM-software/abi-aa/pull/247

smeenai:
Thanks for the confirmation and follow-up!

smithp35:
Apologies for the delayed response, just come back from vacation.

It is worth mentioning that there is a generic ABI issue with how GOT relative addends are represented in the AArch64 ABI https://github.com/ARM-software/abi-aa/issues/217 it looks like this instance may fall into that category. To summarise it looks like neither LLD, Mold, Gold or GNU ld implement `GDAT(S+A)` as written in the spec. Moreover they don't implement it in a consistent way either, with lld doing `GDAT(S) + A` and GNU ld just ignoring the `A`.

So far there hasn't been a satisfactory resolution as it's not clear what all the linker's can support. Probably best leave any further comments on https://github.com/ARM-software/abi-aa/issues/217 

I'll leave a comment in https://github.com/ARM-software/abi-aa/pull/247 too

PR #216, opened 2023-08-16
[AAELF64] Reserve dynamic tags used by memtagabielf64 and pauthabi64

smithp35:
The alpha ELF extension documents define dynamic tags. To avoid avoidable clashes with the main specification, mark the values used as reserved.

PR #225, opened 2023-10-09
Change MEMTAG_STACK to be d_val, and MEMTAG_GLOBALS to be d_ptr

hctim:
MEMTAG_STACK should be d_val (a bit to indicate that MTE stack is either on/off), and MEMTAG_GLOBALS should be d_ptr (a pointer offset to a section containing memtag globals descriptors).

This is the way that lld has implemented it, as it's the local semantics (`MEMTAG_GLOBALS` is a pointer-to-section that needs load bias adjustment, and `MEMTAG_STACK` is a single bit that enables `PROT_MTE` on the stack).

This unfortunatly does mean that we diverge from the accepted semantics of `odd == d_val`, and `even == d_ptr`, but this seems like the best trade-off as we've shipped a compiler that did the non-spec-but-more-logical thing and shipped binaries to end users. Much better than shipping a whole new set of `MEMTAG_GLOBALS_2` and `MEMTAG_STACK_2` to fix the mistake.

Found by eugenis@ in https://r.android.com/2765590

### DIFF
--- a/memtagabielf64/memtagabielf64.rst
+++ b/memtagabielf64/memtagabielf64.rst
@@ -366,12 +366,29 @@ MemtagABI adds the following processor-specific dynamic array tags:
   +---------------------------------+------------+--------+-------------------+-------------------+
   | ``DT_AARCH64_MEMTAG_HEAP``      | 0x7000000b | d\_val | Platform specific | Platform Specific |
   +---------------------------------+------------+--------+-------------------+-------------------+
-  | ``DT_AARCH64_MEMTAG_STACK``     | 0x7000000c | d\_ptr | Platform specific | Platform Specific |
+  | ``DT_AARCH64_MEMTAG_STACK``     | 0x7000000c | d\_val | Platform specific | Platform Specific |
   +---------------------------------+------------+--------+-------------------+-------------------+
-  | ``DT_AARCH64_MEMTAG_GLOBALS``   | 0x7000000d | d\_val | Platform specific | Platform Specific |
+  | ``DT_AARCH64_MEMTAG_GLOBALS``   | 0x7000000d | d\_ptr | Platform specific | Platform Specific |
   +---------------------------------+------------+--------+-------------------+-------------------+
   | ``DT_AARCH64_MEMTAG_GLOBALSSZ`` | 0x7000000f | d\_val | Platform specific | Platform Specific |
   +---------------------------------+------------+--------+-------------------+-------------------+
+
+.. warning::
+
+  Earlier revisions of this document had ``DT_AARCH64_MEMTAG_STACK`` and
+  ``DT_AARCH64_MEMTAG_GLOBALS`` with incorrect ``d_un`` (``d_ptr`` and ``d_val``
+  respectively). Binaries compiled with clang-17 and lld-17 produced the dynamic
+  entries with ``DT_AARCH64_MEMTAG_STACK`` as ``d_val`` and
+  ``DT_AARCH64_MEMTAG_GLOBALS`` as ``d_ptr`` as this was the intended semantics,
+  and they were shipped on Android devices. The values were thus updated to
+  their more fitting semantic types (without updating the ``Value``), but this
+  does mean that ``DT_AARCH64_MEMTAG_STACK`` is a ``d_val`` with an even
+  ``Value``, and ``DT_AARCH64_MEMTAG_GLOBALS`` is a ``d_ptr`` with an odd
+  ``Value`` (where the normal semantics are ``odd == d_val``, and ``even ==
+  d_ptr``). Implementations of dynamic loaders need to be careful to apply these
+  semantics correctly - notably the load bias should not be applied to
+  ``DT_AARCH64_MEMTAG_STACK``, as it's a ``d_val``, even though the ``Value`` is
+  even.
 
 ``DT_AARCH64_MEMTAG_MODE`` indicates the initial MTE mode that should be set. It
 has two possible values: ``0``, indicating that the desired MTE mode is


memtagabielf64/memtagabielf64.rst
```diff
@@ -366,9 +366,9 @@ MemtagABI adds the following processor-specific dynamic array tags:
   +---------------------------------+------------+--------+-------------------+-------------------+
   | ``DT_AARCH64_MEMTAG_HEAP``      | 0x7000000b | d\_val | Platform specific | Platform Specific |
   +---------------------------------+------------+--------+-------------------+-------------------+
-  | ``DT_AARCH64_MEMTAG_STACK``     | 0x7000000c | d\_ptr | Platform specific | Platform Specific |
+  | ``DT_AARCH64_MEMTAG_STACK``     | 0x7000000c | d\_val | Platform specific | Platform Specific |
```
smithp35, line 369: The ELF rules for d_val and d_ptr for tags in the LOOS and HIOS range are inferred from whether the tag is even or odd.

As the spec is in Alpha I'm assuming ABI breaks are possible? If so I recommend switching the order of the tags in the table.

I'm not sure quite what to do if ABI breaks are not possible. We may need to add more tags with the right even/odd value so that if this ever got deplyed outside of Android it could be used by a dynamic loader that inferred val/ptr from that.

From https://www.sco.com/developers/gabi/latest/ch5.dynamic.html
```
To make it simpler for tools to interpret the contents of dynamic section entries, the value of each tag, except for those in two special compatibility ranges, will determine the interpretation of the d_un union. A tag whose value is an even number indicates a dynamic section entry that uses d_ptr. A tag whose value is an odd number indicates a dynamic section entry that uses d_val or that uses neither d_ptr nor d_val. Tags whose values are less than the special value DT_ENCODING and tags whose values fall between DT_HIOS and DT_LOPROC do not follow these rules.
```

smithp35: Thanks for the update I'll need to update my pull request to reserve the numbers in the main specification https://github.com/ARM-software/abi-aa/pull/216

smithp35 (approved):
I've marked as approved. LGTM, I'll leave it open just in case my colleagues want to add any additional comments.

smithp35:
I think that you are right that a dynamic linker that did not recognize the tag couldn't hope to run the binary. I think the cases where it could be a problem are:
* A pretty printing tool like readelf might have different displays for d_val and d_ptr. This would not be the end of the world if it got it wrong though.
* A dynamic linker could do a RelocateMachineSpecificDynamicTagEntries() and add the programs displacement to all even entries. Later a switch statement could process the relocated entries.

In practice it looks like no open-source ELF reader or dynamic loader that I checked makes use of the even entries being d_ptr. It seems like every one uses a switch statement and combines the relocation and processing at the same time.

It does seem like if a dynamic linker did make use of the property for relocation, it could compensate later by subtracting/adding the displacement when doing the memtag specific processing later on.

To summarise, I personally would be happy with your proposal to put a note explaining that the two tags have the wrong parity (if that's the right word) according to the ELF standard. I'll open it up to my colleagues to see if there are any objections.

MaskRay:
https://www.sco.com/developers/gabi/latest/ch5.dynamic.html says

> A tag whose value is an odd number indicates a dynamic section entry that uses d_val or that uses neither d_ptr nor d_val. Tags whose values are less than the special value DT_ENCODING and tags whose values **fall between DT_HIOS and DT_LOPROC do not follow these rules**.

Some processors, such as MIPS and PPC64, have values that do not follow the even/odd convention. It's probably still a good idea to follow the convention, but I agree that as the values as been deployed, there is not much need to change it now.

hctim:
I've updated the patch, commit message, and top post (which I'm guessing should be the squashed merge commit) now with the wording that I think is correct.

The patch:
 - Changes `MEMTAG_STACK` to be `d_val` instead of `d_ptr` without changing the number,
 - Changes `MEMTAG_GLOBALS` to be `d_ptr` without changing the number,
 - Explains why this is strange but why it's the least-awful option over introducing a whole new set of tags.

smithp35 (approved):
Thanks for the update. I've approved the final version. The way the project is set up is a rebase and merge, so all 3 commits will be applied in the current state.

Internally we've been doing a force push of the squashed commit prior to clicking the merge button in github. I'm just double checking that with my colleagues. Hope to get you an answer soon.

Update: Yes, please do squash the changes so we'll be able to apply a single commit.

memtagabielf64/memtagabielf64.rst
hctim: (resolving, looks like #216 doesn't have the new types)

hctim (commented):
Wow, squashing was difficult. Sorry for any noise. Looks like I squashed it properly.

smithp35 (approved):
Thanks for the update, and apologies for the faff.

memtagabielf64/memtagabielf64.rst
```diff
@@ -366,13 +366,30 @@ MemtagABI adds the following processor-specific dynamic array tags:
...
   +---------------------------------+------------+--------+-------------------+-------------------+
 
+.. warning::
+
+  Earlier revisions of this document had ``DT_AARCH64_MEMTAG_STACK`` and
+  ``DT_AARCH64_MEMTAG_GLOBALS`` with incorrect ``d_un`` (``d_ptr`` and ``d_val``
+  respectively). Binaries compiled with clang-17 and lld-17 produced the dynamic
+  entries with ``DT_AARCH64_MEMTAG_STACK`` as ``d_val`` and
+  ``DT_AARCH64_MEMTAG_GLOBALS`` as ``d_ptr`` as this was the intended semantics,
+  and they were shipped on Android devices. The values were thus updated to
```
sallyarmneale, line 383: remove thus - it's archaic language and difficult for non-native speakers.

"The values were updated to more fitting semantic types.

memtagabielf64/memtagabielf64.rst
```diff
@@ -366,13 +366,30 @@ MemtagABI adds the following processor-specific dynamic array tags:
...
+  their more fitting semantic types (without updating the ``Value``), but this
+  does mean that ``DT_AARCH64_MEMTAG_STACK`` is a ``d_val`` with an even
+  ``Value``, and ``DT_AARCH64_MEMTAG_GLOBALS`` is a ``d_ptr`` with an odd
+  ``Value`` (where the normal semantics are ``odd == d_val``, and ``even ==
+  d_ptr``). Implementations of dynamic loaders need to be careful to apply these
```
sallyarmneale, line 388: change "need to be" to "should be"

_Implementations of dynamic loaders should be careful to apply these semantics correctly. Most importantly, you should not apply the load bias to......_

memtagabielf64/memtagabielf64.rst
```diff
@@ -366,13 +366,30 @@ MemtagABI adds the following processor-specific dynamic array tags:
...
+  semantics correctly - notably the load bias should not be applied to
+  ``DT_AARCH64_MEMTAG_STACK``, as it's a ``d_val``, even though the ``Value`` is
+  even.
+
 ``DT_AARCH64_MEMTAG_MODE`` indicates the initial MTE mode that should be set. It
 has two possible values: ``0``, indicating that the desired MTE mode is
 Synchronous, and ``1``, indicating that the desired mode is Asynchronous. This
```
sallyarmneale, line 395: Change to a bulleted list:

It has two possible values:
* 0 - indicates that the 
* 1 - indicates that the....

sallyarmneale (commented):
Couple of minor changes

memtagabielf64/memtagabielf64.rst
stuij: addressed by https://github.com/ARM-software/abi-aa/pull/278

memtagabielf64/memtagabielf64.rst
stuij: addressed by https://github.com/ARM-software/abi-aa/pull/278

PR #227, opened 2023-10-23
[AAELF64][PAUTHABI64] Assign PAuthABI relocation codes.

smithp35:
Now that there will be a concrete implementation of PAuthABI in upstream LLVM we need to move the PAuthABI relocation codes out of the space reserved for private experiments to their final codes. These are in the range of codes already reserved for the PAuthABI in AAELF64.

The description of the operations will remain in the PAuthABI64 as this remains an extension.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -236,6 +236,7 @@ changes to the content of the document for that release.
   |            |                             | DT_AARCH64_VARIANT_PCS.                                          |
   +------------+-----------------------------+------------------------------------------------------------------+
   | 2024Q1     | 29\ :sup:`th` January 2024  | Update preferred ELF marking scheme to be GNU property based     |
+  | 2023Q4     | 18\ :sup:`th` March 2024    | Update relocation codes to move out of private experiments space.|
   +------------+-----------------------------+------------------------------------------------------------------+
 
 References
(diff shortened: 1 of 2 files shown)


pauthabielf64/pauthabielf64.rst
```diff
@@ -698,7 +700,7 @@ Static Data relocations
   |             |                          |                |                                                     |
   |             |                          |                |                                                     |
   +=============+==========================+================+=====================================================+
-  | 0xE100      | R\_AARCH64\_AUTH\_ABS64  | ``PAUTH(S+A)`` | Signing schema encoded in the contents of the place |
+  | 0x244       | R\_AARCH64\_AUTH\_ABS64  | ``PAUTH(S+A)`` | Signing schema encoded in the contents of the place |
```
asl, line 703: We are having two relocations named `R_AARCH64_AUTH_ABS64` with different ids (static and dynamic). Should this be named `R_AARCH64_AUTH_ABS64_STATIC` as otherwise there is an ambiguity??

smithp35: This is the same as R_AARCH64_ABS64 which can be both a static and dynamic relocation. I followed that convention as R_AARCH64_AUTH_ABS64 is its PAuth equivalent.

Changing codes is possible but I thought it better to stick the existing convention.

asl: It's not the code, it's the name. Essentially we cannot put these two relocations into a single enum – we'd need to rename one entry.

smithp35: Actually I think it is the code, and I've made a mistake, sorry.

The intention is that the name and code remain the same. In the original both had 0xE100, but in the new one I used the first code in the dynamic range, it needed to retain its original 0x244. I'll update the pull request.

smithp35: I've updated the value. I had it right in aaelf64.rst but not here.

asl:
Just to note: the current implementation in LLVM is done according to this spec.

smithp35:
I'll set a review deadline for 12/05/2024, I'll merge unless there are any objections.

smithp35:
Rebased changes prior to merge

sallyarmneale (commented):
No changes required.

PR #228, opened 2023-10-24
[AAELF64][SYSVABI64] Move dynamic linking contents to sysvabi64

smithp35:
The Elf for the Arm 64-bit Architecture (aaelf64) describes the processor specific parts of the ELF specification common to all ELF environments from a statically linked freestanding environment with no OS to a hosted environment that supports dynamic linking.

The dynamic linking parts that are specific to SystemV style operating environments such as Linux, and in particular the GNU properties which are based on a Linux ABI extension are better documented in the SystemV ABI for Arm 64-bit Architecture.

PR #229, opened 2023-11-03
[DesignDocuments] Rationale for GNU Properties in sysvabi

smithp35:
Add a design rationale for use of GNU properties as well as guidelines for how these should be used for properties in the AArch64 processor space.

Pull request https://github.com/ARM-software/abi-aa/pull/228 moves the GNU properties and other dynamic section properties specific to SystemV ABI to the SystemV ABI document.

Arm has typically left metadata in exectuables and shared-libraries to the platform. Only defining metadata for relocatable objects. With platforms such as Linux the most frequently run software on AArch64, Arm needs to document the metadata that it is using for SystemV platforms.

We have chosen to use GNU properties and to document these in the sysvabi64.rst document.

### DIFF
--- a/design-documents/elf64-loadable-unit-marking-rationale.rst
+++ b/design-documents/elf64-loadable-unit-marking-rationale.rst
@@ -1,0 +1,496 @@
+..
+   Copyright (c) 2023, Arm Limited and its affiliates.  All rights reserved.
+   CC-BY-SA-4.0 AND Apache-Patent-License
+   See LICENSE file for details
+
+.. |copyright-date| replace:: 2023
+
+.. _AAELF32: https://github.com/ARM-software/abi-aa/releases
+.. _AAELF64: https://github.com/ARM-software/abi-aa/releases
+.. _ACLEMULTIVERSION: https://arm-software.github.io/acle/main/acle.html#function-multi-versioning
+.. _ARMARM: https://developer.arm.com/documentation/ddi0487/latest
+.. _ADDENDA32: https://github.com/ARM-software/abi-aa/releases
+.. _ARMV8V9: https://developer.arm.com/documentation/102378/0201/Armv8-x-and-Armv9-x-extensions-and-features
+.. _BUILDATTR64: https://github.com/ARM-software/abi-aa/releases
+.. _GCCOPTS: https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html
+.. _LINUX_ABI: https://github.com/hjl-tools/linux-abi/wiki/Linux-Extensions-to-gABI
+.. _HWCAPS: https://docs.kernel.org/arch/arm64/elf_hwcaps.html
+.. _PAUTHABI64: https://github.com/ARM-software/abi-aa/releases
+.. _SYSVABI64: https://github.com/ARM-software/abi-aa/
+.. _X86_64PSABI: https://raw.githubusercontent.com/wiki/hjl-tools/x86-psABI/x86-64-psABI-draft.pdf
+
+Design Rationale for Loadable-unit marking for Arm 64-bit Architecture (AARCH64)
+********************************************************************************
+
+Preamble
+========
+
+Licence
+-------
+
+This work is licensed under the Creative Commons
+Attribution-ShareAlike 4.0 International License. To view a copy of
+this license, visit http://creativecommons.org/licenses/by-sa/4.0/ or
+send a letter to Creative Commons, PO Box 1866, Mountain View, CA
+94042, USA.
+
+Grant of Patent License. Subject to the terms and conditions of this
+license (both the Public License and this Patent License), each
+Licensor hereby grants to You a perpetual, worldwide, non-exclusive,
+no-charge, royalty-free, irrevocable (except as stated in this
+section) patent license to make, have made, use, offer to sell, sell,
+import, and otherwise transfer the Licensed Material, where such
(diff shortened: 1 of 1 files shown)


smithp35:
I don't think the PDF creation is to do with this pull-request as I didn't add the directory to the generate-pdfs script and the same attribute error happens in other documents.

design-documents/elf64-loadable-unit-marking-rationale.rst
```diff
@@ -0,0 +1,496 @@
+..
+   Copyright (c) 2023, Arm Limited and its affiliates.  All rights reserved.
```
yury-khrustalev: Nit: update copyright year

design-documents/elf64-loadable-unit-marking-rationale.rst
```diff
@@ -0,0 +1,496 @@
...
+recorded in the ELF file. This document will use the team ELF
+loadable-unit marking scheme to describe the metadata.
+
+Design Goals
+============
+
+The design goals of an ELF marking scheme are:
+
+- Utilize existing ELF marking standards where possible.  Platforms
+  may already have an implementation existing standards that can be
```
yury-khrustalev: Nit: did you mean "implementation **of** existing..."

design-documents/elf64-loadable-unit-marking-rationale.rst
```diff
@@ -0,0 +1,496 @@
...
+
+A feature may also require use of an architectural extension in a
+sufficient number of functions that would make runtime selection of
+function implementations impractical. For example an implementation of
+the pointer authentication ABI (PAuthABI64_) requires all functions in
+the program to agree on the signing schema.
+
+To permit an ELF loader to reason about feature compatibility and take
+extra actions as a result of features some additional metadata must be
+recorded in the ELF file. This document will use the team ELF
```
Wilco1: What is "team ELF"?

smithp35: A typo for "term ELF"

design-documents/elf64-loadable-unit-marking-rationale.rst
```diff
@@ -0,0 +1,496 @@
...
+The SystemV ABI for the Arm 64-bit architecture (SYSVABI64_) will use
+GNU program properties in executables and shared-libraries to
+communicate architectural and ABI properties to the execution
+environment.
+
+GNU program properties
+======================
+
+GNU program properties are a linux extension defined in
+(``LINUX_ABI_``). They are encoded in ``SHT_NOTE`` sections with a
```
yury-khrustalev: Nit: some references are written with double inverse quotes ` and as a result they are not rendered as links.

design-documents/elf64-loadable-unit-marking-rationale.rst
```diff
@@ -0,0 +1,496 @@
...
+can be recorded in a property.
+
+Focus on userspace architectural features
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Userspace programs are expected to be portable between software
+platforms running on different hardware.  Platform specific software
+can be customized for the hardware that it runs on.
+
+Only require features that can't be tested at runtime
```
yury-khrustalev: We have properties for PAC, BTI, and GCS all of which can be tested at run time via HWCAP. The next paragraph about hint space features explains this but this paragraph seems to contradict what exists in practice.

The properties for PAC, BTI, and GCS inform the loader that the binary is supposed to work with the respective features if they are present in the system (so it is safe to enable them when the run time check passes).

design-documents/elf64-loadable-unit-marking-rationale.rst
smithp35: Ack

design-documents/elf64-loadable-unit-marking-rationale.rst
smithp35: Now updated to "term ELF"

design-documents/elf64-loadable-unit-marking-rationale.rst
smithp35: Yes, now updated.

design-documents/elf64-loadable-unit-marking-rationale.rst
smithp35: Removed the double inverse quotes from the references.

design-documents/elf64-loadable-unit-marking-rationale.rst
smithp35: I think what I need to do is restrict this paragraph to "use of non-hint space instructions from an architectural extension".

Main thing to get over is that if function multiversioning is used then we don't need to record all the architecture features used in the alternative implementations, just the baseline.

BTI and GCS are (or can be) orthogonal to this. For example BTI only uses hint-space instructions, but requires action from the loader to enable BTI. In this case we record "I am compatible with BTI", and not "I require armv8.5-a for BTI instructions".

I'll have a go at rewriting.

yury-khrustalev: Thanks, this makes sense to me

yury-khrustalev (approved):
LGTM. Let's merge this!

smithp35:
Thanks for the review. Now merging.

PR #230, opened 2023-11-03
AArch64 build attributes specification

smithp35:
Introduce a new specification document for AArch64 build attributes.

The 32-bit ABI makes extensive use of the build attributes defined in addenda32. These attributes defined for relocatable objects are used for a number of purposes including:

* Checking ABI compatibility.

* Checking required processor feature compatibility.

* Automatic static library selection.

With AArch64's focus on software platforms, there has not been a strong requirement to introduce extra metadata to denote the requirements and optional processor features used by a relocatable object. Programs are expected to do runtime testing to check if processor features are available.

With the advent of security features like PAuthABI, BTI and GCS we need a way to record additional metadata that a static linker can use to set metadata like GNU properties in a loadable-unit (executable or shared-library).

A brief summary:

* Made a generic "aeabi" prefix for Arm subsections. Including a header that states optionality and encoding. All ULEB or all NTBS.

* Describes a single subsection that maps directly to the existing .note.gnu.properties and uses language similar to AAELF64. This includes the existing BTI and PAC properties.

* Is accompanied by a design rationale with a suggestion for tool syntax.


smithp35:
I don't think the PDF creation is to do with this pull-request as I didn't add the directory to the generate-pdfs script and the same attribute error happens in other documents.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+   long long int and pointer are 64-bit.
+
+
+This document uses the following terms and abbreviations.
+
+Link-unit
+   An executable or shared library. Also known as loadable-unit in
+   this document.
+
+Loadable-unit
```
MaskRay, line 331: The existing docs don't use the term "loadable-unit". "link unit" has been used in several docs. Is there a rationale  to introduce this alias?

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,540 @@
...
+
+- ``GNU_PROPERTY_AARCH64_FEATURE_1_AND`` transfers the ``AND`` of all
+  the feature bits from the relocatable object inputs to the output
+  property. In practice an additional input for some the feature bits
+  has come from the static linker, making the implementation less
+  generic than it could be.
+
+The major differences between build attributes and GNU properties are:
+
+- Build attributes are relocatable object only and have an encoding
```
MaskRay, line 270: This sentence appears to contradict with "Encoding of meta data in a loadable-unit" in `buildattr64.rst`. If the recommended practice is for loaders not to inspect the platform-specific build attributes, mention it in `buildattr64.rst`?

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,540 @@
...
+attributes.  This can be used as the basis of finding the best set of
+pre-compiled libraries out of a selection of pre-compiled
+alternatives. Intuitively the best candidate library makes the most
+demands of the execution environment without exceeding the
+capabilities of the exception environment.
+
+- All candidate libraries with build attributes ≤ selection build
+  attributes are compatible.
+
+- For each compatible candidate library L with best candidate B then
```
MaskRay, line 309: I am puzzled by this sentence. Perhaps clarify it?

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+Default values for public tags
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The effect of omitting a public tag in file scope is identical to
+including it with a value of 0 or “”, depending on its parameter type.
+
+Formal Syntax of aeabi prefixed subsections
+-------------------------------------------
+
+<uint8: optional> <uint8: parameter type>
```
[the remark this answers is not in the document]
smithp35: That is true. In this particular case I'm not expecting too many different subsections, currently we only define one so my instinct is to keep it simple for now.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+have a prefix of "aeabi". If the header is optional then a consumer
+may skip the subsection if they do not recognize the full vendor-name.
+
+Default values for public tags
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The effect of omitting a public tag in file scope is identical to
+including it with a value of 0 or “”, depending on its parameter type.
+
+Formal Syntax of aeabi prefixed subsections
```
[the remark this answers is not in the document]
smithp35: Yes, will fix.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+
+*  If a1 ≤ a2, a1 + a2 = a2, else if a2 ≤ a1,
+   a1 + a2 = a1. (‘+’ behaves like the *maximum* function).
+
+*  If a1 <> a2 there are two mutually exclusive sub-cases.
+
+   *  There is a least a3 such that a1 ≤ a3 and a2 ≤ a3.
+      Then a1 + a2 = a3.
+
+      Example: ``Tag_CPU_arch`` from (ADDENDA32_) when a1 = v6KZ, a2 = v6T2, and a3 = v7.
```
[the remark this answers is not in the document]
smithp35: Yes. I'm somewhat limited with the examples right now.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+The compatibility between an entity with Tag_PAuth_Platform = 0,
+Tag_PAuth_Schema = 0, and Tag_Pauth_Platform != 0, Tag_PAuth_Schema !=
+0 is implementation defined.
+
+aeabi-feature-and-bits and GNU Program Properties
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+GNU Program Properties as defined by [LINUX_ABI_] are a similar
+concept to Build Attributes. Program properties are encoded in
+.note.gnu.property sections. A static linker combines the program
```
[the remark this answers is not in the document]
smithp35: Yes, will fix.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,901 @@
...
+     <id> The version number of the Schema.
+
+Where the *<id>* is the version number in the context of
+Tag_PAuth_Platform.
+
+The signing schema is not optional.
+
+The parameter type is ULEB128.
+
+The lattice value is custom. Two entities are compabitible if both
```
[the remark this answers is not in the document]
smithp35: Looking it up, it seems like join-semilattice is the formal term. What I'll do for now is explain that it is a join-semilattice but we'll use lattice for brevity.

buildattr64/buildattr64.rst
smithp35: In some internal feedback some of my colleagues didn't like link-unit :-) . Loadable unit was felt to be more inituitively understandable than link-unit.

I'm confortable with link-unit, but that's many years of familiarity.

design-documents/buildattr64-rationale.rst
smithp35: In `buildattr64.rst` there is:
```
Encoding of meta data in a loadable-unit
----------------------------------------

Build attributes are only defined for relocatable objects.  The
encoding of metadata in a loadable-unit (executable or shared-library)
is platform specific.  For example a platform may choose to use GNU
program properties as defined in (LINUX_ABI_). This document specifies
the meaning of build attributes for relocatable objects only.  The
presence of ``.ARM.attributes`` sections in a loadable-unit is
permitted, but the interpretation of the contents of the section is
outside the scope of this document.
```

I think that is consistent with `Build Attributes are relocatable only`.

Arm has traditionally taken the approach that the executable and shared-library are mostly "owned" by the platform holder, be that Android or the Linux community and the AArch64 ABI. I've written https://github.com/ARM-software/abi-aa/pull/229 which is a rationale for using GNU properties in sysvabi64.rst files which we equate to Linux/BSD.

Hope that makes sense. Any suggestions for improvement are welcome though.

design-documents/buildattr64-rationale.rst
smithp35: OK, I'll try and rewrite it.

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,540 @@
...
+
+- Subsections may share common implementation properties such as all
+  tags having the same partial order which could make them easier to
+  process. For example if all tags in a subsection have the value 0
+  or 1, and have a partial order that is identical or reverse of the
+  tags arithmetic value, then an implementation can represent this as
+  a bit vector and use the binary ``or`` or ``and`` operator to do the
+  join.
+
+Alternative one .aeabi_attributes section
```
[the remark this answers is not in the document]
smithp35: OK, will fix.

buildattr64/buildattr64.rst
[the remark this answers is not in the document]
smithp35: I ended up using partial order to remove the use of lattice. While build attributes are modelled on a lattice, the addenda32 only mentions it once.

MaskRay:
Thanks for starting https://discourse.llvm.org/t/rfc-aarch64-build-attributes-for-elf-relocatable-objects/75161 and I came here while reading that post. I have done a lot of object file format/assembler/linker work in llvm-project and I am in favor of AArch64 Build Attributes. I haven't studied much about PAuth, but if loaders don't need to know the relevant build attributes, using build attributes should be good.

> LLD processing of Build Attributes sections similar to the ARM target.

In `lld/ELF/Driver.cpp`, I added `mergeRISCVAttributesSections`. For AArch64 we will need something similar.

MaskRay (commented):
Forgot to push some "pending comments"...

buildattr64/README.md
```diff
@@ -0,0 +1,36 @@
+<div align="center">
+   <img src="Arm_logo_blue_RGB.svg" />
+</div>
+
+# Build Attributes for the Arm® 64-bit Architecture (AArch64)
+
+## About this document
+
+The [Build Attributes for the Arm 64-bit Architecture](buildattr64.rst) describes the
```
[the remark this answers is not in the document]
smithp35: ACK

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,905 @@
...
+
+For a platform that uses GNU Program Properties in loadable-units the
+mapping table above can be used to translate build attributes back to
+GNU properties.
+
+Private subsections
+-------------------
+
+The tool interface for private subsections is implementation defined.
+
```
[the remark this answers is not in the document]
smithp35: ACK

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,547 @@
...
+attributes can have advantages when modelling properties that:
+
+- can be represented by a value where the order of values represents
+  the degree to which the property has the value. For example if there
+  is a monotonically increasing set of hardware capabilities this can
+  be represented as a single attribute with a single value rather than
+  multiple feature bits with forcing functions. A concrete example is
+  ``FEAT_LSE2`` which can only exist if ``FEAT_LSE`` exists.
+
+- require more then one attribute to describe. For example the
```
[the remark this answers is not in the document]
smithp35: ACK

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,547 @@
...
+Similarly there would not need to be tool support for decoding and
+printing build attributes.
+
+There are several downsides:
+
+- Humans make textual errors, requiring validation and error handling.
+
+- Textual formats are slower to parse and take up more space.
+
+
```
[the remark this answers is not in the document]
smithp35: ACK

MaskRay (approved):

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+    +=====================+============================+=======+===========+======+============+
+    | ``.ARM.attributes`` | ``SHT_AARCH64_ATTRIBUTES`` | 0x0   | 1         | 0    | 0          |
+    +---------------------+----------------------------+-------+-----------+------+------------+
+
+A consumer may not assume the natural alignment of a data type such as ``uint32``.
+
+The syntactic structure of an attributes section is::
+
+   <format-version: <version-letter>>
+   [ <uint32: subsection-length> NTBS: vendor-name
```
sivan-shani, line 660: The first byte will be <version> followed by 0 or more subsections.
```
[ <uint32: subsection-length> NTBS: vendor-name
  <bytes: vendor-data>
]
```
Is the equivalent of a single `.aeabi_subsection` in the assembly.
for each `.aeabi_subsection` there must be single
```
[ <uint32: subsection-length> NTBS: vendor-name
  <bytes: vendor-data>
]
```
In the ELF.

Each unique .aeabi_subsection vendor-name will result in a single subsection. That permits an assembler to allow .aeabi_subsection to appear multiple times in the same file with the same vendor name, like a .section directive. Therefore it is possible to change subsections, but be able to return to the previous one. I don't think this is likely in practice, but it is often how "stateful" directives are implemented in assembler.

For example:

// switch to subsection aeabi-feature-and-bits
.aeabi_subsection aeabi-feature-and-bits , 0, 0
.aeabi_attribute Tag_Feature_BTI, 1

// switch to subsection aeabi-pauthabi
.aeabi_subsection aeabi-pauthabi, 1, 0
.aeabi_attribute Tag_PAuth_Platform, 1
.aeabi_attribute Tag_PAuth_Schema, 1

// switch back to subsection aeabi-feature-and-bits
// Note this doesn't create a new subsection.
.aeabi_subsection aeabi-feature-and-bits , 0, 0
.aeabi_attribute Tag_Feature_GCS, 1

There will be two subsections in the ELF file aeabi-feature-and-bits and aeabi-pauthabi.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+    | Name                | Type                       | Flags | Alignment | Info | Entry Size |
+    +=====================+============================+=======+===========+======+============+
+    | ``.ARM.attributes`` | ``SHT_AARCH64_ATTRIBUTES`` | 0x0   | 1         | 0    | 0          |
+    +---------------------+----------------------------+-------+-----------+------+------------+
+
+A consumer may not assume the natural alignment of a data type such as ``uint32``.
+
+The syntactic structure of an attributes section is::
+
+   <format-version: <version-letter>>
```
sivan-shani, line 659: The  <format-version: <version-letter>> (double '<, >' also around 'version-letter') and <uint32: subsection-length> (single '<, >', not around 'subsection-length') does not indicated a different between the cases.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+==============================================================
+
+Encoding
+--------
+
+Encoding build attributes are encoded in a vendor-specific section of
+type ``SHT_AARCH64_ATTRIBUTES`` and name ``.ARM.attributes`` (for
+further details see [AAELF64_]).
+
+An attribute is encoded in a <tag, value> pair.
```
sivan-shani, line 617: A tag with the same name in a different .aeabi_subsections is considered a different tag.

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,545 @@
...
+Assembler
+---------
+
+Where possible the assembler can derive build attributes from the
+assembler's command line options in the same way as the compiler.  For
+options that cannot be derived the following directives can be used to
+construct "aeabi" prefixed subsections.
+
+::
+   .aeabi_subsection name [optional] [, parameter type]
```
sivan-shani, line 473: The syntax is not final. There are consideration around whether the values should be optional and with what exact syntax.
Options:
- as it is now, and to denote optional arguments: 
 .aeabi_subsection MySectionA 0 -> 'optional'
 .aeabi_subsection MySectionA, 0 -> 'parameter type'
- The syntax might be .aeabi_subsection name [, optional] [, parameter type]
and to denote optional arguments:
.aeabi_subsection MySectionA, 0,   or: 
.aeabi_subsection MySectionA, ,0
- Having both the arguments being mandatory
- Make encoding and parameter type either both present or absent

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+attributes in a single pass over the section, then all of its private
+data in a second pass.
+
+Public aeabi prefixed subsections
+=================================
+
+About public tags
+-----------------
+
+Subsections containing a vendor name with a prefix of "aeabi" are
```
sivan-shani, line 701: Subsections containing a vendor name with a prefix of "aeabi"

"containing vendor name"
If it is correct to understand that aeabi-feature-and-bits and aeabi-pauthabi are some of those "public subsections" that starts with aeabi mentioned here, then where is the "vendor name"? Isn't vendor name supposed to be something like "Arm", "Intel" etc.?

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+
+Formally, there are no constraints on the order or number of vendor
+subsections. A consumer can collect the public ("aeabi" prefixed)
+attributes in a single pass over the section, then all of its private
+data in a second pass.
+
+Public aeabi prefixed subsections
+=================================
+
+About public tags
```
sivan-shani, line 698: Here is a suggestion for this paragraph that reflect my understanding of public subsections:
```suggestion
Public subsections

Subsections are denoted with the tag: `.aeabi_subsection <name>` 
Public subsections are subsections with <name> starting with a prefix of "aeabi".
Public subsections have a header that indicates whether the subsection is mandatory or not, and what is the type of the value for all the subsection's attributes.  

A consumer must be able to parse the header of all the mandatory public subsections. A consumer may skip an optional subsection.
```

Several points:
- The header does not indicates only optionality, but also the attribute's values type.
- The part about the subsection that do not need to be read is confusing. The essence is, if I am not mistaken: If subsection is mandatory, it has to be read, not being able to read it is an error. If it is optional, it can not be read, from any reason (e.g. not being able to recognize the full vendor-name or any other reason, what is special about the vendor-name?)

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+will indicate whether the section is required or optional.
+
+A consumer must be able to parse the header of all subsections that
+have a prefix of "aeabi". If the header is optional then a consumer
+may skip the subsection if they do not recognize the full vendor-name.
+
+Default values for public tags
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The effect of omitting a public tag in file scope is identical to
```
sivan-shani, line 713: It is unclear to me whether this is true only for optional subsections, only for mandatory subsections,  or for both, optional and mandatory and subsections.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+Subsections containing a vendor name with a prefix of "aeabi" are
+public attributes defined by this specification. All public aeabi
+sections have the same format, which includes a header. The header
+will indicate whether the section is required or optional.
+
+A consumer must be able to parse the header of all subsections that
+have a prefix of "aeabi". If the header is optional then a consumer
+may skip the subsection if they do not recognize the full vendor-name.
+
+Default values for public tags
```
sivan-shani, line 710: Might it be a good idea to call 'tags' -> subsections and therefore 'public tags' -> public subsections?
"Tags" might mean quite a few things.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,904 @@
...
+The specification will note known forcing functions. Implementation of
+forcing functions is Q-o-I.
+
+Representing build attributes in a relocatable ELF Object file
+==============================================================
+
+Encoding
+--------
+
+Encoding build attributes are encoded in a vendor-specific section of
```
sivan-shani, line 613:
```suggestion
Build attributes are encoded ...
```

smithp35:
Thanks for all the comments. I'll aim to get a response tomorrow and I'll post an update to the spec and rationale this week.

buildattr64/buildattr64.rst
smithp35: Thanks for the suggestion, I'll have a think about what I can incorporate.

I'm thinking that I could move the assembly support into the main specification as an appendix.

buildattr64/buildattr64.rst
smithp35: I'll see if I can make the mandatory part clearer. I think an examples section in the rationale would help too. All tags get an implicit default of 0 whether in a mandatory or optional section.

A subsection should be mandatory if a static linker must be able to understand at least one of the tags defined in it to produce correct output. By making the subsection mandatory a static linker is expected to give an error message if it can't process the subsection.

If the subsection is optional and a static linker doesn't understand either the subsection or a tag within it then it can just skip over it. The output will be correct, if not optimal.

buildattr64/buildattr64.rst
smithp35: I'll see if I can define these terms a bit better. A public tag is just a tag defined in a public (name starting with aeabi) subsection.

I agree that tag is a general term, I've taken a lot of the language from the 32-bit ABI.

design-documents/buildattr64-rationale.rst
smithp35: Internally with a conversation with the GNU team we decided that it should be `[, optional]` I think that while possible to disambiguate between
```
.aeabi-subsection vendor-name optional
```
and
```
.aeabi-subsection vendor-name ,ntbs
```
Is possible, given that optional and ,ntbs can be digits then it will be difficult for a human to spot.

smithp35 (commented):
I'll start work on an update tomorrow.

design-documents/buildattr64-rationale.rst
```diff
@@ -0,0 +1,540 @@
...
+
+
+*name*
+
+Create or switch the current subsection to *name*.
+
+*optional*
+
+This field is optional and only applies to subsection names with a
+prefix of "aeabi".  The default value is 1 for optional.
```
sivan-shani: Starting to doubt whether it is a good idea to have optional parameters with default values.
It creates the possibility for user mistakes when writing build attributes as assembly and implementation mistakes. 
Are there any real benefits for allowing the omission of parameters?

smithp35: Default values are necessary for reasoning about the properties of old object files in new toolchains and vice-versa. If we introduce a new attribute in ABI release N, then toolchains that implement some version prior to N won't have that attribute, toolchains need to be able to reason about that case as we can't guarantee that everyone recompiles all their objects, including all their dependencies, all of the time. An explicit default value of 0 forces the attribute definition to think about backwards compatibility. With the most common kinds of attribute combination  being "all objects must have property X to set X in output", "If at least one object has property X to set X in output" or "all objects must have the same property for the combination to be valid" then 0 works well.

sivan-shani: Thank you for the reply, it is a bit difficult to understand.

Given old object file O1 which does not have attrX,
Being handled with new object files O2, O3 that has attrX.
The way to handle O1, O2 and O3 together will be implemented in the logic of the handling tool (e.g. linker)
Something like "if no attrx, then..."

It is not so clear how having default values for attrX helps this process.
1) Having default value for attrX can not imply that old object has to be treated like it has the default value, because there is no way to know if it will make sense for future attributes.
2)  "An explicit default value of 0 forces the attribute definition to think about backwards compatibility" - Why? when there is an old object involved, it is not the same as having new object with no explicit attribute (see (1)) It is a completely different case that has to be handled anyhow. e.g. even in a world without old object files, there could be a case of a missing attribute, either buy design (if the attribute is optional) or mistake (if it is mandatory)

It seems to be completely unrelated?

sivan-shani: There is some logic to the idea that the default is '0', and it means that no attribute -> value is '0', and then it imply that all other objects can not have '1'.
However, if this is the intention, do we really want to force all future developers to comply with this logic? It is a bit micromanagy on our part, and it might be better to leave those decisions to the future implementers.

sivan-shani: Sorry, I am gathering my thoughts slowly. Just want to add this:
Handling old object files has to be done anyhow, and I believe that having default values will hardly effect it or will hint anything to those who will implement it, while it will add the need to remember that 1) there are default values and 2) what they are to everyone who read/write assembly, or handle the tools that parse/write assembly/elf. We adding complexity and making things more error prone for benefits that does not seems quite there.

sivan-shani: (sorry, a very last comment ) It will be clear from the partial order defined on the attribute, what should be the way to handle the case of missing attribute. (or "what should be the default value for missing attribute") and in the process of implementation and review, it should be done correctly, is there a need to hint it by setting a default value?

buildattr64/buildattr64.rst
sivan-shani: Continuing the discussion regarding default values per tag:
If we have the role that omitting a tag is like having it with value "0", then this is all we need in order to handle old object files(?) 
Then it is not clear what is the part of having a default value per tag has to do with handling old object files?

design-documents/buildattr64-rationale.rst
smithp35: Hope this goes through all the questions:

A given tool will understand a non-strict subset of the documented attributes and documented subsections of the current specification.

A build attributes producer like an assembler or compiler can only set the attributes and subsections it knows about. A build attributes consumer like a linker can only take action  (the encoding permits it to read the attributes even if it doesn't understand them) on attributes and subsections it knows about.

On the producer side, the attributes and subsections that the tool doesn't know about will not be present. When the intended value of the attribute is the same as the default then it is up to the producer whether to emit the attribute or leave it out. Personally I would expect tools like compilers to just set the attribute explicitly to 0, however in hand-written assembly I would expect humans to only set the ones they care about. It isn't realistic to expect a human to look up all the attributes in the spec.

On the consumer side it is clear what happens if an attribute is encountered that the consumer doesn't understand. If it is in an optional subsection, it ignores it, if it is an non-optional subsection it gives a diagnostic.

If the consumer encounters an object file that is missing an attribute that it understands it uses the default 0 value for the object. As you suggest the partial order and combination rule will tell the consumer how to update the output value of the attribute.

sivan-shani: Thanks, I was confusing Assembly default values, such as subsection header values such as optional/value and ELF default value such as an entire tag.
ELF does not have default values like Assembly, therefore my point was not relevant.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,1115 @@
...
+* *tag* is either the ``Tag_tag_name`` such as ``Tag_Feature_GCS`` or
+  the ``Tag value`` such as ``3``. The matching of ``Tag_tag_name`` is case
+  insensitive.
+
+* *value* is either a `number` or a "string" depending on the
+  ``<parameter type>`` of the current subsection.
+
+In the current active subsection, set *tag* to *value*.
+
+It is an error if .aebi_attribute is encountered when the current
```
sivan-shani, line 1076: Typo, '.aeabi_attribute'

smithp35: Thanks for spotting, will update.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,1148 @@
...
+  +----------------------------+                                                                 +
+  | ``NTBS``                   | The parameter type field is set to 1.                           |
+  +----------------------------+-----------------------------------------------------------------+
+
+::
+
+   .aeabi_attribute *tag*, *value*
+
+* *tag* is either the ``Tag_tag_name`` such as ``Tag_Feature_GCS`` or
+  the ``Tag value`` such as ``3``. The matching of ``Tag_tag_name`` is case
```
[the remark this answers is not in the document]
smithp35: Following what discussion? Can you give me a public reference?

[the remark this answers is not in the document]
sivan-shani: The current [llvm implementation ](https://github.com/llvm/llvm-project/commit/ee99c4d4845db66c4daa2373352133f4b237c942)
And from my understanding also the upcoming gnu one treat Tag_tag_name as case sensitive.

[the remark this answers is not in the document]
smithp35: OK. I can change it if both implementations are treating it as case sensitive.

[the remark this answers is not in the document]
smithp35: Now done with latest version.

buildattr64/buildattr64.rst
```diff
@@ -0,0 +1,1162 @@
...
+     0 The user did not permit this entity to use the PAuthABI, or no information available.
+     <id> The version number of the Schema.
+
+Where the *<id>* is a number representing the version in the context
+of ``Tag_PAuth_Platform``.
+
+Combining attribute values of aeabi_pauthabi
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The partial order is custom. Two entities are compabitible if both
```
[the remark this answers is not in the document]
smithp35: Thanks for spotting, will update with a fix.

buildattr64/buildattr64.rst
```diff
@@ -1150,7 +1150,10 @@ Directives
 
 *name*
 
-Create or switch the current subsection to *name*.
+Create or switch the current subsection to *name*. When a subsection
+with *name* already exists, *comprehension* and *parameter type* can
+be omitted; if they are present then they must match the existing
```
sivan-shani: Maybe:
```
Switching to an existing subsection requires only *name*,  *comprehension* and *parameter type* can
be omitted;
```

smithp35:
There are now implementations of Build Attributes in GNU and LLVM. Specification has been updated to Beta and will be merged.

PR #231, opened 2023-11-22
[SYSVABI64] Add Guarded Control Stack (GCS) Feature Bit

smithp35:
Add GNU_PROPERTY_AARCH64_FEATURE_1_GCS to the GNU property GNU_PROPERTY_AARCH64_FEATURE_1_AND.

This permits executables and shared libraries to declare that they are compatible with GCS, this permits a platform to enable GCS for a program when the executable and all of its dependecies are compatible with it.

When GCS is enabled, use of dlopen of a shared library that is not GCS compatible is platform defined. Some platforms may be able to disable GCS, others may decide to refuse to load the library.

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -200,8 +200,11 @@ Change History
  | 02Alpha    | 6\ :sup:`th` April 2023      | Define the processor specific interpretation for      |
  |            |                              | DT_PLTGOT                                             |
  +------------+------------------------------+-------------------------------------------------------+
- | 2023Q4     | 24\ :sup:`th` 2023           | Added Program Property, Program Loading and           |
+ | 2023Q4     | 22\ :sup:`nd` November 2023  | Added Program Property, Program Loading and           |
  |            |                              | Dynamic Linking contents from `AAELF64`_              |
+ |            |                              |                                                       |
+ |            |                              | Added Dynamic linking content for GCS including       |
+ |            |                              | GNU_PROPERTY_AARCH64_FEATURE_1_GCS Feature Bit        |
  +------------+------------------------------+-------------------------------------------------------+
 
 References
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -1643,11 +1648,25 @@ Its use is optional, meaning that an ELF file where this feature bit
 is unset can still have Return Address signing enabled in some or all of
 its executable sections.
 
+``GNU_PROPERTY_AARCH64_FEATURE_1_GCS`` This indicates that all
+executable sections are compatible with the Guarded control stack
+(GCS) mechanism. Minimum requirements for setting this feature bit
+include:
+
+* The number of ``procedure return address push operations`` and the
```
john-brawn-arm, line 1656: The terms "procedure return address push operation" and "procedure return address pop operation" are defined only in the GCS specification (as far as I know) and are perhaps a bit misleading here (on first reading I thought it was talking about the STP at function start and LDP at function end, but it's actually talking about BL and RET). Also having a matching number of BL/RET isn't enough, every RET has to match up with the BL that called the function.

I think rewording this to be something like the following would be clearer:
> Each function that is called using a BL instruction (or other instruction that is a GCS procedure return address push operation) returns using a RET instruction (or other instruction that is a GSC procedure return address pop operation). This means that RET instructions are used for function returns, and not as an indirect branch.

smithp35: Thanks for the suggestion, I've used an almost identical wording in a fixup commit that I'll squash prior to merging.

sysvabi64/sysvabi64.rst
```diff
@@ -1643,11 +1648,27 @@ Its use is optional, meaning that an ELF file where this feature bit
...
 
+``GNU_PROPERTY_AARCH64_FEATURE_1_GCS`` This indicates that all
+executable sections are compatible with the Guarded control stack
+(GCS) mechanism. Minimum requirements for setting this feature bit
+include:
+
+* Each function that is called using a BL instruction (or other
+  instruction that is a GCS ``procedure return address push
+  operation``) returns using a RET instruction (or other instruction
+  that is a ``GCS procedure return address pop operation``). This
```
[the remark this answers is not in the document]
smithp35: Apologies for the mistake. Looking at the spec it should be outside. I want to highlight that `procedure return address push operation` is a term from a specification, that I expect will eventually be in the Arm ARM as a highlighted term. Without it procedure return address push operation could just be a generic sentence.

Will fix.

DanielKristofKiss (approved):

sallyarmneale (commented):
No changes required

PR #232, opened 2023-11-30
Define soft-float PCS for AArch64

ostannard:
With the release of the Cortex-R82, we now have an AArch64 CPU which can be built without any floating-point instructions or registers, so I think we should re-consider our decision to not have a soft-float variant of the AArch64 PCS.

I've worded this as a variant over the base PCS, to keep it simple.

No compiler implements this yet, but clang does already allow passing floating-point types without an FPU, and seems to implement the first bullet point correctly. I'm working on LLVM patches which will implement bullet points 2 and 3.

GCC doesn't currently allow any use of floating-point types without an FPU for AArch64, so we don't need to worry about compatibility with existing code there.

For the fourth bullet point, I don't think we need to define the PCS for vector types, because these are only defined by the ACLE when the relevant hardware exists.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -253,6 +253,7 @@ changes to the content of the document for that release.
 |            | October 2023       |                                                                  |
 +------------+--------------------+------------------------------------------------------------------+
 |            |                    | - Change the status of the SME support from Alpha to Beta.       |
+|            |                    | - Add soft-float PCS variant.                                    |
 +------------+--------------------+------------------------------------------------------------------+
 
 References
@@ -2027,6 +2028,21 @@ size\_t, ptrdiff\_t
 -------------------
 
 See `Arm C and C++ language mappings`_.
+
+Soft-float
+----------
+
+The soft-float variant is defined for Arm v8-R implementations that do not have floating-point instructions or registers. This variant is incompatible with the base procedure call standard and toolchains are not required to support it.
+
+The soft-float variant is defined identically to the base variant, with these changes:
+
+* All floating-point types are passed as if they are an Integral machine type with the same byte-size and alignment.
+
+* No types are considered to be HFAs or HVAs.
+
+* The definition of ``va_list`` is unchanged. The ``__vr_top`` and ``__vr_offs`` fields may be left uninitialised by ``va_start``, and their values must not be relied on by ``va_arg``.
+
+* The calling convention for Short Vector, Scalable Vector and Scalable Predicate machine types is left undefined.
 
 .. raw:: pdf
 


aapcs64/aapcs64.rst
```diff
@@ -2026,6 +2026,21 @@ size\_t, ptrdiff\_t
...
+Soft-float variant
+------------------
+
+For v8-R CPUs which do not have floating-point instructions or registers, the soft-float variant of the PCS can be used. Arm recommends that this is not used more widely, as most existing AArch64 implementations do have floating-point registers, and this variant is incompatible with the base variant.
+
+The soft-float variant is defined identically to the base variant, with these changes:
+
+* All floating-point types are passed as if they are an integer type with the same bit-width.
+
+* No types are considered to be HFAs or HVAs, these are treated as if they are non-homogeneous composite types.
```
smithp35, line 2038: Does the implementation treat these as just structs of integers (of the appropriate size) for example a struct of two floats would be treated as if it were a struct of two integers.

From the specification
```
struct S {
  int x; // substitute for float
  int y; // substitute for float
};
```
Would be considered a homogenous aggregate, just not a homogenous floating point aggregate. I can't see anything in the argument passing rules that does anything special with this definition so it would behave as if a non-homogenous composite type. However for future changes would we prefer it to be defined as a homogenous aggregate or a non-homogenous composite?

If we do want to go down that route perhaps:
``Types that the base procedure call standard would consider HFAs or HVAs are treated as homogenous aggregates of signed Integral type with the same byte size and alignment."

aapcs64/aapcs64.rst
```diff
@@ -2026,6 +2026,21 @@ size\_t, ptrdiff\_t
 
 See `Arm C and C++ language mappings`_.
 
+Soft-float variant
```
smithp35, line 2029: I think this document and aapcs32 hasn't put variant in the title. Presumably inheriting it from "The standard variants". Perhaps just "Soft-float"

aapcs64/aapcs64.rst
```diff
@@ -2026,6 +2026,21 @@ size\_t, ptrdiff\_t
...
+------------------
+
+For v8-R CPUs which do not have floating-point instructions or registers, the soft-float variant of the PCS can be used. Arm recommends that this is not used more widely, as most existing AArch64 implementations do have floating-point registers, and this variant is incompatible with the base variant.
```
smithp35, line 2032: I think we could use stronger language here. Perhaps
The soft-float variant is defined for Arm v8-R implementations that do not have floating-point instructions or registers. This variant is incompatible with the base procedure call standard and toolchains are not required to support it.

smithp35 (commented):
The variant definition looks good to me. I've made some comments/suggestions on the language, some of which could probably do with a second opinion so don't feel that you need to apply them. My intent for most of them is to try and use as similar language as the rest of the document.

aapcs64/aapcs64.rst
ostannard: Done

aapcs64/aapcs64.rst
ostannard: Done

aapcs64/aapcs64.rst
```diff
@@ -2026,6 +2026,21 @@ size\_t, ptrdiff\_t
...
+
+The soft-float variant is defined identically to the base variant, with these changes:
+
+* All floating-point types are passed as if they are an integer type with the same bit-width.
```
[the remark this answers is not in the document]
ostannard: Done, but I've dropped the word `signed` to avoid implying that the integer representation is meaningful, other then the bit-pattern matching the floating point value.

aapcs64/aapcs64.rst
ostannard: I agree that the original wording conflicts with the existing definition of `homogenous aggregate`, but I think that change could be confusing, since `homogenous aggregates of signed Integral type` isn't used anywhere else. Maybe it would be better to cut this down to just "No types are considered to be HFAs or HVAs"?

smithp35: I'd be happy with just "No types are considered to be HFAs or HVAs" I think that is sufficient for someone to work out what they need to do by following the rest of the rules in the doc and avoids a potentially confusing definition.

smithp35 (approved):
I've approved the changes. I've added mmalcomson as a reviewer. Will wait to see if he has any further comments before merging.

mmalcomson (approved):
Appreciate waiting on me (especially given my holidays causing delays).
I don't really have anything to add -- from what I see we've had good discussion on trying to ensure it is clear the scope across which we expect this variant PCS to act.
I appreciate the work -- thanks both!

aapcs64/aapcs64.rst
```diff
@@ -2028,6 +2029,21 @@ size\_t, ptrdiff\_t
...
+
+The soft-float variant is defined identically to the base variant, with these changes:
+
+* All floating-point types are passed as if they are an Integral machine type with the same byte-size and alignment.
+
+* No types are considered to be HFAs or HVAs.
+
+* The definition of ``va_list`` is unchanged. The ``__vr_top`` and ``__vr_offs`` fields may be left uninitialised by ``va_start``, and their values must not be relied on by ``va_arg``.
+
+* The calling convention for Short Vector, Scalable Vector and Scalable Predicate machine types is left undefined.
```
sallyarmneale, line 2045: Add the final comma in the list:

The calling convention for Short Vector, Scalable Vector, and Scalable Predicate machine types is left undefined.

sallyarmneale (commented):
Minor changes required.

aapcs64/aapcs64.rst
ostannard: #277

PR #233, opened 2023-12-05
[sysvabi] Make hard-float a requirement for sysvabi

smithp35:
The aapcs64 base procedure call standard requires floating point and simd registers. With the advent of a soft-float ABI for R-profile CPUs without floating point, we are making it explicit that platforms based on the System V ABI are required to use the base procedure call standard.

Soft-float pull request in https://github.com/ARM-software/abi-aa/pull/232

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -210,6 +210,8 @@ Change History
  |            |                              |   used by `PAuthABIELF64`_ and                        |
  |            |                              |   `MemTagABIELF64`_.                                  |
  +------------+------------------------------+-------------------------------------------------------+
+ | 2025Q1     | 04\ :sup:`th` February 2025  | Require hard-float ABI for sysvabi platforms          |
+ +------------+------------------------------+-------------------------------------------------------+
 
 References
 ----------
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -363,6 +365,16 @@ AArch64 system.
...
 
+Hardware Requirements
+---------------------
+
+The AArch64 System V ABI requires the presence of SIMD and FP registers.
+
+Procedure call standard requirements
+------------------------------------
+
+The AArch64 System V ABI uses The Base Procedure Call Standard from (AAPCS64_).
```
ostannard, line 376: There are some other ABI variations mentioned in the `The standard variants` section of AAPCS64 (fp16 format, sizes of some types), should we mention any of them here too? I think it would be enough to say that the `LP64` data model must be used, maybe with an option to use `ILP32`.

smithp35: I'll have a think. I am a little worried that the statement is a bit too strong, and could rule out things like the vector PCS.

The opening paragraph of https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#6the-base-procedure-call-standard has:
```
Application code is expected to conform to one of three data models defined in this standard; ILP32, LP64 or LLP64.
```
I think that is something close to what I want to say, without duplicating the text exactly. Perhaps I can find a way of referencing it more directly.

stuij (approved):
Sorry for the long wait. LGTM.

smithp35:
Rebased with to resolve conflict with version history.

sallyarmneale (commented):
Looks good to me

PR #242, opened 2024-01-31
Clarify R_<CLS>_RELATIVE optimization wording

smeenai:
The optimization is applied by the static linker, so we should reword
the sentence from that perspective. Thanks to @smithp35 for suggesting
the wording.


MaskRay:
I think "resolves to the current dynamic shared object" conveys the same meaning: rtld binds the symbol reference to the current DSO.

smeenai:
Hmm. I was thinking about this from the perspective of the static linker, since it'll be the one emitting RELATIVE vs. GLOB_DAT as an optimization (the context of the sentence). I guess a better wording for the entire sentence from that perspective would be:

> This relocation represents an optimization; it can be used to replace ``R_<CLS>_GLOB_DAT`` when the symbol ~resolves to~ *is defined in* the current dynamic shared object and is not pre-emptable.

You're right from the rtld perspective, but I don't understand the optimization discussion from that perspective. The static linker is the one deciding what relocation type to emit, and rtld is just acting on it, right?

aaelf64/aaelf64.rst
```diff
@@ -1738,7 +1738,7 @@ The need for copy relocations can be avoided if a compiler generates all code re
...
+``R_<CLS>_RELATIVE`` represents a relative adjustment to the place based on the load address of the object relative to its original link address. All symbols defined in the same segment will have the same relative adjustment. If ``S`` is the null symbol (ELF symbol index 0) then the adjustment is based on the segment defining the place. On systems where all segments are mapped contiguously the adjustment will be the same for each reloction, thus adjustment never needs to resolve the symbol. This relocation represents an optimization; it can be used to replace ``R_<CLS>_GLOB_DAT`` when the symbol resolves to the current dynamic shared object and is not pre-emptable.
```
smithp35, line 1741: I interpreted resolves to the current dynamic shared object as not including the cases when the symbol could be pre-empted. However it maybe one of those cases where knowing the answer makes it easier to make that interpretation.

I do think the current wording could do with some clarification, particularly as RELATIVE relocations are used in position independent executables.

May I suggest the alternative wording:
```
This relocation represents an optimization; a static linker can replace ``R_<CLS>_GLOB_DAT`` when the symbol is known at static link time to always resolve to the current link unit.
```
I've not explicitly mentioned pre-emptable but I think the always resolve catches that.

smithp35 (commented):
Thanks very much for patch. 

I've made an alternative suggestion as I think the text should be more specific about static linking and less specific about the current shared object.

aaelf64/aaelf64.rst
smeenai: That alternative wording sounds great to me; I'd maybe say "a static linker can **use it to** replace" or "a static linker can replace `R_<CLS>_GLOB_DAT` **with it** when" just to be explicit, but it's exactly what I was going for. Did you want to create a separate PR/commit for that or just edit this one?

smithp35: It would be easiest for me if you can choose one of your alternatives (use it to) or (with it) and update this PR, I'm happy with either variation! The ABI unfortunately uses merge and rebase rather than merge and squash, so to avoid a load of fixup commits we do a squash before merging. With such a simple change you may as well just force-push a replacement.

The company prefers for us not to edit PRs, so it is either update this one or submit a new PR which seems like more work than is necessary.

smeenai:
Thanks, I updated the wording with your suggestion. I also updated the PR title and description to match the commit, because I wasn't sure which one would be ultimately used.

smithp35 (approved):
Thanks for the update LGTM.

I'll leave it open for a week before merging to give any of my colleagues a chance to comment.

smeenai:
Can we merge this now? :)

stuij:
Yes :) As per @smithp35's previous comments.

PR #243, opened 2024-02-02
Promote the SME(2) ABI status from Alpha to Beta

rsandifo-arm:
The SME and SME2 ABI have been implemented in upstream GCC and Clang/LLVM, so I think they're ready to be treated as Beta.


stuij (approved):
LGTM

stuij:
thanks!
FWIW, similar changes have already been merged in the ACLE:
https://github.com/ARM-software/acle/pull/299

sallyarmneale (commented):
No changes required. But just a note to say that I *think* you have two spaces after each full stop - this isn't necessary.  A single full stop is standard now.

PR #246, opened 2024-02-15
[aaelf32] Note that R_ARM_PREL31 can create veneers

smeenai:
Since https://reviews.llvm.org/D25721, LLD has assumed this. I assume
other linkers behave similarly, and LLVM's ARM backend also assumes that
R_ARM_PREL31 is PLT-generating. Update the ABI to reflect that.

### DIFF
--- a/aaelf32/aaelf32.rst
+++ b/aaelf32/aaelf32.rst
@@ -2231,8 +2231,8 @@ to effect the transition to Thumb state. Conditional function call instructions
 
 A linker may use a veneer (a sequence of instructions) to implement the
 relocated branch if the relocation is one of R_ARM_PC24, R_ARM_CALL,
-R_ARM_JUMP24, (or, in Thumb state, R_ARM_THM_CALL, R_ARM_THM_JUMP24, or
-R_ARM_THM_JUMP19) and:
+R_ARM_JUMP24, R_ARM_PREL31, (or, in Thumb state, R_ARM_THM_CALL,
+R_ARM_THM_JUMP24, or R_ARM_THM_JUMP19) and:
 
 * The target symbol has type STT_FUNC
 


aaelf32/aaelf32.rst
```diff
@@ -2231,8 +2231,8 @@ to effect the transition to Thumb state. Conditional function call instructions
 
 A linker may use a veneer (a sequence of instructions) to implement the
 relocated branch if the relocation is one of R_ARM_PC24, R_ARM_CALL,
-R_ARM_JUMP24, (or, in Thumb state, R_ARM_THM_CALL, R_ARM_THM_JUMP24, or
-R_ARM_THM_JUMP19) and:
+R_ARM_JUMP24, R_ARM_PREL31, (or, in Thumb state, R_ARM_THM_CALL,
```
smithp35, line 2234: I'm not sure this is the right place as R_ARM_PREL31 is not a call or jump relocation (https://github.com/ARM-software/abi-aa/blob/main/aaelf32/aaelf32.rst#L2037)

There is a section in the ehabi32 which states that most uses of R_ARM_PREL31 must be via the section symbol + addend to avoid the symbol getting indirected via the PLT or a Thunk so the only cases where there is R_ARM_PREL31 is the address of the personality routine.

I think if we were going to put this anywhere then something near:
https://github.com/ARM-software/abi-aa/blob/main/aaelf32/aaelf32.rst#L2557 might work

```
In addition to the data generating relocations listed above the call and branch
relocations (R_ARM_CALL, R_ARM_THM_CALL, R_ARM_JUMP24, R_ARM_THM_JUMP24, R_ARM_THM_JUMP19), and the data relocation R_ARM_PREL31, may also require a proxy to be generated if the symbol will
be defined in an external executable or may be pre-empted at execution time.
```

smithp35 (commented):
Apologies for missing this a few weeks ago. I think we would be better stating this in the section on PLTs as R_ARM_PREL31 is a data relocation.

PR #250, opened 2024-03-19
[PAUTHABIELF64] Remove alternative ELF marking scheme

smithp35:
No implementation is using the alternative marking scheme; take the opportunity to remove it, so that toolchains do not need to support it.

The alternative marking scheme started as the default one back when the spec was written. At the time there were only prototype implementations for which signing schema compatibility and versioning weren't important.

Over time Arm's ELF marking for SysV has evolved such that the preference for ELF executables and shared-libraries is to use GNU Program Properties. We intend to migrate to Build Attributes for relocatable objects, but until that change is made GNU Program Properties should be used in relocatable objects too.

The alternative ELF marking scheme was preserved for backwards compatibility, however no current implementation needs that so it can be removed.


asl:
Confirmed, we are not doing any "alternative" bits.

asl (approved):

smithp35:
Thanks for the confirmation. Merging.

PR #251, opened 2024-03-21
[CLIBABI32] Restore missing reference to EDOM.

statham-arm:
I just noticed a paragraph that was grammatically confusing because it contained a comma with no word before it, and an equality with no left-hand side.

It looks to me as if the <errno.h> macro name EDOM was accidentally deleted in both cases. (Rationale: EDOM is an error code closely related to ERANGE, and matches the claim in the paragraph that its consensus value is 33 – that's its value on Linux, for example.)

Searching for ERANGE in the rest of the document turned up another two places where there was a missing word that looked as if it ought to be "EDOM".

I've no idea how this happened! But it seems like an easy fix.

### DIFF
--- a/clibabi32/clibabi32.rst
+++ b/clibabi32/clibabi32.rst
@@ -828,9 +828,9 @@ Grey list items are a little more difficult. There are two options.
 
 -  Treat the consensus members as *white* and the remainder as *black*.
 
-Consider , ERANGE, and EILSEQ from errno.h. This is a grey list category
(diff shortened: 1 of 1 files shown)


smithp35 (approved):
LGTM

I agree this looks like a typo. The MS documentation for EDOM is also 33 https://learn.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=msvc-170

Issue #253, opened 2024-03-25
 [PAUTHABIELF64] Possible typo in `R_AARCH64_AUTH_GOT_LO12_NC`


kovdan01:
In https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#auth-variant-got-generating-relocations, AUTH variants of GOT-generating relocations are defined. Their names are in most cases basically the same as corresponding non-AUTH relocs names, but for `R_AARCH64_AUTH_GOT_LO12_NC`, it's not true: the non-AUTH variant contains `LD64` (`R_<CLS>_LD64_GOT_LO12_NC`), so the AUTH variant should probably be `R_AARCH64_AUTH_LD64_GOT_LO12_NC`. If `LD64` is considered redundant since we don't have 32-bit versions of AUTH relocations, it should be probably deleted from other AUTH relocs names: in `R_AARCH64_AUTH_LD64_GOTOFF_LO15` and `R_AARCH64_AUTH_LD64_GOTPAGE_LO15`, `LD64` is present.

Is `LD64` omitted in `R_AARCH64_AUTH_GOT_LO12_NC` intentionally? If yes, it would be nice if you explain rationale of such decision.

kovdan01:
Tagging @smithp35 
Tagging @asl

smithp35:
I'm on vacation this week. Will get back to you after Easter.

It is most likely a typo, my memory is that I wanted to match the existing dynamic relocations. I'll check when I get back to see if there is any other reason.

smithp35:
I agree this is a typo. I've created https://github.com/ARM-software/abi-aa/pull/255 to fix.

To the best of my knowledge the Appendix on GOT generating relocations hasn't been implemented before. At the time of writing the interested parties preferred to have a RELRO unsigned GOT.

kovdan01:
Closing the issue as resolved in #255

PR #255, opened 2024-04-02
[pauthabielf64] Fix typo in relocation name

smithp35:
As pointed out in https://github.com/ARM-software/abi-aa/issues/253 the R_AARCH64_AUTH_GOT_LO12_NC is meant to be the AUTH variant of R_AARCH64_LD64_GOT_LO12_NC. As there is also a
R_AARCH64_LD32_GOT_LO12_NC relocation rename the relocation to R_AARCH64_LD64_AUTH_GOT_LO12_NC.

These relocations are in the appendix as we are currently expecting the GOT to be RELRO and unsigned in most signing schemas.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -1167,7 +1167,6 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
   |             |                                        |                                  | value to bits [11:0] of  |
   |             |                                        |                                  | X. No overflow check.    |
   +-------------+----------------------------------------+----------------------------------+--------------------------+
-
 .. raw:: pdf
 
    PageBreak
@@ -1181,7 +1180,9 @@ is the PAuth ABI equivalent of ``R_AARCH64_RELATIVE``. The underlying
 calculation performed by the dynamic linker is the same, the only
 difference is that the resulting pointer is signed. The dynamic linker
 reads the signing schema from the contents of the place of the dynamic
-relocation.
+relocation. The ``R_AARCH64_AUTH_GOT_ADD_LO12_NC`` relocation is an
+addition for the PAuth ABI and has no equivalent in (AAELF64_). It is
+used with the ``:got_auth_lo12:`` operator on an add instruction.
 
 .. table:: Additional AUTH Dynamic relocations
 
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -1153,7 +1153,7 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
   |             |                                        |                                  | check that –2\ :sup:`32` |
   |             |                                        |                                  | <= X < 2\ :sup:`32`      |
   +-------------+----------------------------------------+----------------------------------+--------------------------+
-  | 0x811A      | R\_AARCH64\_AUTH\_GOT\_LO12_NC         | G(ENCD(GDAT(S + A)))             | Set the LD/ST immediate  |
+  | 0x811A      | R\_AARCH64\_AUTH\_LD64\_GOT\_LO12_NC   | G(ENCD(GDAT(S + A)))             | Set the LD/ST immediate  |
   |             |                                        |                                  | field to bits [11:3] of  |
   |             |                                        |                                  | X. No overflow check;    |
   |             |                                        |                                  | check that X&7 = 0       |


MaskRay:
Shall we use `GDAT(S) + A` ? But that can be done after https://github.com/ARM-software/abi-aa/issues/217 has a concrete resolution.

kovdan01:
I've just noticed that `RAARCH64_AUTH_GOT_ADD_LO12_NC` also contains a typo - `_` is missed after `R`.
BTW, is there a corresponding non-auth reloc? I don't see `R_AARCH64_GOT_ADD_LO12_NC` in https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst

smithp35:
> I've just noticed that `RAARCH64_AUTH_GOT_ADD_LO12_NC` also contains a typo - `_` is missed after `R`. BTW, is there a corresponding non-auth reloc? I don't see `R_AARCH64_GOT_ADD_LO12_NC` in https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst

I've updated to remove that relocation. I can't see anything it would be used for apart from calculating the address of a GOT slot, but user-code shouldn't be able to do that anyway and the compiler has other ways to do that if it needs to using the other relocations.

smithp35:
> Shall we use `GDAT(S) + A` ? But that can be done after #217 has a concrete resolution.

I've got an action to make an internal proposal for what to do about #217 this week.

MaskRay (approved):

kovdan01:
@smithp35 

> I can't see anything it would be used for apart from calculating the address of a GOT slot, but user-code shouldn't be able to do that anyway

Is it actually true that we shouldn't be able to do that? According to the spec, address of the GOT entry is used as a modifier when signing the contents of the entry. The example code for accessing signed GOT entry includes this calculation (see https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#default-signing-schema):

```
adrp x8, :got_auth: symbol
add x8, x8, :got_auth_lo12: symbol
ldr x0, [x8]
// Authenticate to get unsigned pointer
autia x0, x8
```

If `R_AARCH64_AUTH_GOT_ADD_LO12_NC` is removed, I suppose the code snippet should be changed correspondingly (not to include `add` with `:got_auth_lo12:` immediate). Personally, I feel that `adrp`+`add` is the most simple way to achieve the desired functionality, so the reloc should probably be left in the spec, just with the typo fixed. Please let me know if I miss smth and there is an easier way to re-implement the code block above.

smithp35:
> @smithp35
> 
> > I can't see anything it would be used for apart from calculating the address of a GOT slot, but user-code shouldn't be able to do that anyway
> 
> Is it actually true that we shouldn't be able to do that? According to the spec, address of the GOT entry is used as a modifier when signing the contents of the entry. The example code for accessing signed GOT entry includes this calculation (see https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#default-signing-schema):
> 
> ```
> adrp x8, :got_auth: symbol
> add x8, x8, :got_auth_lo12: symbol
> ldr x0, [x8]
> // Authenticate to get unsigned pointer
> autia x0, x8
> ```
> 
> If `R_AARCH64_AUTH_GOT_ADD_LO12_NC` is removed, I suppose the code snippet should be changed correspondingly (not to include `add` with `:got_auth_lo12:` immediate). Personally, I feel that `adrp`+`add` is the most simple way to achieve the desired functionality, so the reloc should probably be left in the spec, just with the typo fixed. Please let me know if I miss smth and there is an easier way to re-implement the code block above.

Thanks for reminding me, a lot of the rationale has fallen out of my memory in the time that it was first wrote. I agree with you and this is how the AArch64 PAC PLTs are implemented in LLD, albeit with just the calculation and not the relocation.

In this particular case there is no need to rename the relocation as the ABI only uses the LD64 and LD32 when there is a material difference in the relocation. The  `R_AARCH64_AUTH_GOT_ADD_LO12_NC` has no overflow check and would result
in bits [11:0] regardless of whether we were ILP32 and ILP64. The closest equivalent in the existing AAELF64 is `R_<CLS>_ ADD_ABS_LO12_NC`.

I'll update to put the relocation back in as was, and add a note below to match the relocation with the operator `:got_auth_lo12:`

asl:
@smithp35 Thanks! Looks like there was some discrepancy in the particular implementation. Now everything fully matches.

smithp35:
Thanks for the confirmation. Merging.

PR #256, opened 2024-04-05
[ATOMICSABI64]: Alpha Draft of Atomics ABI

lukeg101:
[atomicsabi64.pdf](https://github.com/ARM-software/abi-aa/files/14890730/atomicsabi64.pdf)
This is the Alpha draft of the ABI for the
"C/C++ Atomics Application Binary Interface Standard for the Arm® 64-bit Architecture"

This document describes the C/C++ Atomics Application Binary Interface for the Arm 64-bit architecture.

This document concerns the valid mappings from C/C++ Atomic Operations to sequences of A64 instructions. This document does not support Armv7.

For matters concerning the memory model, please consult §B2 of the Arm Architecture Reference Manual.

We focus only on a subset of the C11 atomic operations and their mapping to A64 instructions at the time of writing.

More atomics will be added.

Co-Authored with Wilco Dijkstra (@Wilco1).


lukeg101:
@stuij @smithp35 @Wilco1

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+#. If any *c* in *C* exhibits a compiler bug with respect to the outcomes *S*
+   then the given mappings are not interoperable.
+
+Using Mix Testing we now define ABI-Compatibility of Atomic Operations.
+
+
+Definition of ABI-Compatibility for Atomic Operations
+-----------------------------------------------------
+
+*A compiler that implements the above set of Mappings is ABI-Compatible with
```
smithp35, line 1136: As I understand it, the ABI is a binary interface. Tools that implement the interface must implement it correctly for compatibility. With the description here Mix Testing sounds like a compliance test and doesn't necessarily alter the definition of ABI compatibility. It could be argued that if Mix testing is just compliance testing then it doesn't belong in the specification, it would be best described in an adjunct document (for example the design-documents section).

Where I could see it coming in is if an alternative code sequence is to be added then it must be shown to be equivalent using Mix Testing.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+   PageBreak oneColumn
+
+
+Preamble
+========
+
+Abstract
+--------
+
+This document describes the C/C++ Atomics Application Binary Interface for the
```
smithp35, line 47: I'm trying to work out what part of this document is ABI, i.e. what are the requirements for an implementation of the C/C++ atomic operations to be ABI compatible with other ABI compatible implementations; and what part is a cookbook of known/tested implementations. Which while helpful, I'm not sure is strictly part of the ABI?

I think it could be that any assembly sequence is considered ABI compatible if it passes Mix testing against the C memory model, and this document lists the sequences that have passed that bar already?

Other things I was expecting to see in an ABI document is whether a conforming implementation can use locks? For example I think one of the problems for ABI is whether an operation is lock-free or not. I think all the assembly sequences here are lock-free. If I'm right (not an expert in this area) then could be worth making it explicit.

smithp35 (commented):
Thanks very much for putting this together. Most of my comments are based on first time reading rather than as a concurrency expert.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+By comparing ``W3`` and the local variable ``r0`` of the original Concurrent
+Program we see there is one additional Outcome of executing the compiled
+program that is not an outcome of executing the Concurrent Program. This is due
+to the fact that according to the Arm Architecture Reference Manual [ARMARM_] 
+*instructions where the destination register is WZR or XZR, are not regarded as
+doing a read for the purpose of a DMB LD barrier.*
+
+ISO C permits a conforming implementation to delete unused data, but in this
+case it introduces another Outcome of Execution. To fix this issue, a compiler
```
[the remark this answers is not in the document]
Wilco1: I think we should just shorten it to say you can't use atomics with the zero register as the destination and list the various cases. This is more of a cookbook item, however compilers get this wrong all the time, so clearly we have to explain how to do it correctly...

atomicsabi64/atomicsabi64.rst
Wilco1: Chapter 4 is the ABI. Yes we have to list the various sequences plus explain the pitfalls because using atomics correctly is extremely hard. Compilers get these sequences wrong all the time without anyone noticing. So if you do not emit the listed sequences, then you're mostly likely getting atomics wrong.

Locks are inherently inefficient, unfair and can easily lead to deadlock. The 8/16/32/64/128-bit atomic sequences shown are obviously lock-free irrespectively of the architecture version. We can make that more explicit, including how to deal with the wacky non-power of 2 structure types that C/C++ allow.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+appropriate <op> Assembly Sequence inside the loop. Exceptions, where Assembly 
+Sequences exist, are stated (for instance ``fetch_or`` can be implemented using
+``LDSETP`` when the LSE128 extension is enabled).
+
+Special Cases
+-------------
+
+There are special cases in the Mappings presented above, these must be handled
+in order to prevent unexpected outcomes of the compiled program.
+
```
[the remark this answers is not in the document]
Wilco1: Agreed, it needs a bullet list of the special cases.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+Assembly Sequences. Since there is a large number of ways these mappings may be
+combined, we break down the tables by the width of the access, and list
+compatible Assembly Sequences for each Atomic Operation.
+
+This is an open ABI, we encourage improvements to this specification to be
+submitted to the `issue tracker page on
+GitHub <https://github.com/ARM-software/abi-aa/issues>`_.
+
+These mappings are not exhaustive, but aim to cover the atomics we have tested.
+Please request more atomics using the issue tracker.
```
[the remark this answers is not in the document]
lukeg101: I'm on leave so I'l reply to the rest of these when I am back but I think it is important to clarify something here. Defining new instructions is a memory model issue and should be directed to the memory model team, the ABI is an issue of _compiler_ mappings which is more appropriate for this document and whoever maintains the compilers. In other words, the burden is not on us to implement or test the semantics of the instructions, that goes to whoever defines the instructions (which is separate from using them in the compiler). Adding mappings to the ABI different as it involves understanding the atomics AND compilers, and whether new instructions can be used to implement them, which is the realm of compiler folks. So here's a possible process with a boundary:

1. Someone implements the semantics of the new instruction in the tool of their choice (e.g. herd). 
---- boundary where the ABI comes in ----
2. Candidate mappings are then considered by whoever maintains the ABI (on an ad-hoc basis perhaps as the need/request arises as a PR?)
3.a) ABI team checks existing tests,
3.b) If the Mapping has new semantics (it uses an instruction with a special case or new ordering semantics e.g. SWP or LDAPR) then we may need to add new tests/special cases but this is rare. More likely a new instruction sequence has the same semantics as other instruction/sequences but is introduced into some new extension as an alternative.
4. If ok, add the mapping to the ABI.

I suspect the burden of maintaining or adding mappings is actually very small. Whilst Wilco is correct in that it is necessary to consider all interactions with existing sequences as well as potential (compatible) future instructions, the new sequences are often very similar (look at the repetition in the doc itself) and very rarely are new atomics mappings are proposed, let alone something that is semantically different in some way.

In any case, I enjoy this and so I am happy to take a look as needed and suggest tests.

atomicsabi64/atomicsabi64.rst
lukeg101: I think we should explain Mix Testing Technique in the ABI itself. The comment above ([here](https://github.com/ARM-software/abi-aa/pull/256#discussion_r1555923959) demonstrates why quite well. By this I mean to help answer your question:

> "(could it be multiple compiler profiles?) [...] Stitch together the assembly outputs from above into multiple (all?) possible (legal) interleavings of execution that when run produce C."

The Mix testing Technique compiles all source fragments using all input profiles, and then generate all the tests with the compiled sequences. Since a given Atomic operation has multiple mappings, there are multiple possible assembly sequences we could use. Only once this stitching is done do we check the executions of _all_ tests (1 input test : multiple output tests). We should document exactly what it is we are doing (mix testing) in the ABI so these same questions don't arise in the reader later on. Compatibility is deliberately defined in terms of Mix Testing because the latter is the technique by which we ensure the former (as far as testing can ensure).

Atomics are complicated as Wilco has suggested in other comments. The technique is equally tricky. I think stating exactly _how_ ABI compatibility is achieved in the same document is important to ensure there is no confusion.

Note also this is independent of discussions of any tool that might implement the mappings (compiler) or mix testing (a testing tool) which I suppose is where a design document comes in.

lukeg101: perhaps a middle ground is the appendix, and I'm happy to provide even more documentation, but Mix Testing certainly should remain in this doc.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+  +--------------------------------------------------------+--------------------------------------+
+  | Atomic Operation                                       | Assembly Sequence                    | 
+  +============================================+===========+======================================+
+  | ``atomic_store_explicit(loc,val,relaxed)`` | ARCH1     | ``option A``                         |
+  +                                            +-----------+--------------------------------------+
+  |                                            | ARCH2     | ``option B``                         |
+  +--------------------------------------------+-----------+--------------------------------------+
+
+Where ARCH is for example BASE (armv8), LSE, LSE2, LSE128, RCPC, or LRCPC3.
```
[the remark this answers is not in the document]
Wilco1: Yes it uses the `FEAT_` names from the Arm ARM, we could expand it a bit.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+Under ISO C, the above Concurrent Program finishes execution in one of three
+possible outcomes::
+
+  { thread_1:r0=0; y=1; }
+  { thread_1:r0=1; y=1; }
+  { thread_1:r0=1; y=2; }
+
+In this case the value read by the exchange on ``thread_1`` is not used, and a
+compiler is free to remove references to unused data. It is thus legal under
```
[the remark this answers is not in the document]
Wilco1: Indeed, it's not - it's only correct to remove unused variable assignments and computations.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  +-----------------------------------------+--------------------------------------+
+  | ``memory_order_seq_cst``                | ``sc``                               |
+  +-----------------------------------------+--------------------------------------+
+
+In what follows ``loc`` refers to the location, ``val`` refers to a value
+parameter.
+
+Arbitrary registers may be used in the Assembly Sequences that may change in
+compiler implementations. Cases where arbitrary registers may *not* be used are
+covered in the Special Cases section.
```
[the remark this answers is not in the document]
Wilco1: Yes it's best to list all the cases explicitly in section 4.7. We could improve the general comment here to state that XZR as a destination is generally invalid and point to 4.7, but listing all the exceptions in the table (there are also other register constraints etc) would make it too complex. Maybe RST allows footnotes?

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+----------
+
+This document refers to, or is referred to by, the following documents.
+
+.. table::
+
+  +-------------+--------------------------------------------------------------+-----------------------------------------------------------------------------+
+  | Ref         | External reference or URL                                    | Title                                                                       |
+  +=============+==============================================================+=============================================================================+
+  | ARMARM_     | DDI 0487                                                     | Arm Architecture Reference Manual Armv8 for Armv8-A architecture profile    |
```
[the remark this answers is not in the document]
lukeg101: done

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+      produced relocatable files must conform in order to be statically
+      linkable and executable.  For example, the C++ ABI for the Arm 64-bit
+      Architecture [CPPABI64_], or ELF for the Arm Architecture [AAELF64_].
+
+Arm-based
+   ... based on the Arm architecture ...
+
+Concurrent Program
+   A C or C++ program that consists of one or more Threads of Execution. Each
+   Thread of Execution must communicate with other threads in the Concurrent
```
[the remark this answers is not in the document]
lukeg101: I'm not sure what terms are best here, that is technically true but we are focusing on AArch64 only here. I will add a disclaimer in this definition.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+   ``atomic_thread_fence`` to synchronize the order of accesses made by atomic
+   operations on ``_Atomic`` qualified data.
+
+Atomic Instruction
+   An A64 instruction that may have Memory Order semantics. For instance an A64
+   LDR instruction has no atomicity, but the LDAR instruction has *acquire*
+   semantics. (see [ARMARM_]). 
+
+Assembly Sequence
+   A sequence of Atomic Instructions.
```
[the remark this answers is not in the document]
lukeg101: changed to 'A sequence of A64 instructions, optionally including Atomic Instructions.'

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+Mapping
+   A pair of Atomic Operation and Assembly Sequence. A compiler generates the
```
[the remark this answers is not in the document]
lukeg101: yes this `((Atomic Operation, Compiler Profile), Assembly Sequence)` type is correct. Will use your definition

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+   Assembly Sequence, given an Atomic Operation and Compiler Profile as input.
+
+Compiler Profile
```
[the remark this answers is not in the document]
lukeg101: replaced with `A Compiler implementation and command-line flags or attributes that use
   Mappings.`

atomicsabi64/atomicsabi64.rst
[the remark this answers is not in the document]
lukeg101: I will add `ARCH describes the required extension, with BASE meaning Armv8-A with no extensions.`

and clarify that e.g. `RCPC3` means `FEAT_RCPC3`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  +------------------------------------------+--------------------------------------+
+  | ``fence(relaxed)``                       | ``NOP``                              |
+  +------------------------------------------+--------------------------------------+
+  | ``fence(acq)``                           | ``DMB ISHLD``                        |
+  +------------------------------------------+--------------------------------------+
+  | | ``fence(rel)``                         | ``DMB ISH``                          |
+  | | ``fence(acq_rel)``                     |                                      |
+  | | ``fence(sc)``                          |                                      |
+  +-------------------------------+----------+--------------------------------------+
+  | ``exchange(loc,val,relaxed)`` | ``BASE`` | | ``loop:``                          |
```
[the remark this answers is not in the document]
lukeg101: I removed all the extra bars and generally cleaned up the tables

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  +---------------------------------+-------------+--------------------------------------+
+  | ``compare_exchange_strong(``    | ``BASE``    | | ``loop:``                          |
+  |   ``loc,&exp,val,relaxed,``     +             + |   ``LDXP   X6, x7, [X4]``          +
+  |   ``relaxed)``                  +             + |   ``CMP    X6, X0``                +
+  +                                 |             | |   ``CCMP   X7, X1, 0, EQ``         |
+  |                                 +             + |   ``CSEL   X8, X2, X6, EQ``        +
+  |                                 +             + |   ``CSEL   X9, X3, X7, EQ``        +
+  |                                 +             + |   ``STXP   W5, X8, X9, [X4]``      +
+  |                                 +             + |   ``CBNZ   W5, loop``              +
+  +                                 +             + | ``MOV   X0, X6``                   +
```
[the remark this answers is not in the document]
lukeg101: The indentation differences are expected, is there a convention we adopt in the ABI for this? I will keep indentation unless you'd prefer a flat list?

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+There are special cases in the Mappings presented above, these must be handled
+in order to prevent unexpected outcomes of the compiled program.
+
+Re-Ordering of Read-Modify-Write Effects and Acquire Fence
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Consider the following Concurrent Program::
+
+  // Shared-Memory Locations
```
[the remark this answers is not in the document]
lukeg101: I haven't used code-block for now, if it still renders strangely when I upload, then I'll try the code-block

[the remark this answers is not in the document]
Wilco1: Without the bars I used, the pdf renders incorrectly. If code-block works for both then that's probably what we need to use.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+   may not be the same as the order they are written in the program.
+   Synchronization Operations are statements that constrain the order of
+   accesses made to Shared-Memory Locations by each thread. Synchronization
+   Operations include Thread Fences, and certain control flow structures.
+
+Shared-Memory Location
+   A memory location that can be accessed by any Thread of Execution in the
+   program.
+
+Memory Order Parameters
```
walhua, line 299: Memory Order Parameter

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  above by generating C/C++ Concurrent Programs that permute combinations of
+  Atomic Operations on each Thread of Execution. We bound our test size between
+  2 and 5 Threads of Execution, where each Thread has at least 1 Atomic
+  Operation or Synchronization Operation and at most 5 Atomic Operations or
+  Synchronization Operations. We do not make any statement about the
+  ABI-Compatibility of Concurrent Programs outside these bounds.
+* We test Concurrent Programs with a fixed initial state, loop unroll factor
+  (equal to 1 loop unroll), and function calls or recursion. 
+* The above Mappings are not exhaustive, We hope that Arm's partners will
+  submit requests for other Mappings to the ABI team using the issue tracker
```
walhua, line 1153: it should be a link for consistency: [issue tracker
page on GitHub](https://github.com/ARM-software/abi-aa/issues)

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  bounded testing. Atomics ABI-compatibility is thus tested for the Mappings
+  above by generating C/C++ Concurrent Programs that permute combinations of
+  Atomic Operations on each Thread of Execution. We bound our test size between
+  2 and 5 Threads of Execution, where each Thread has at least 1 Atomic
+  Operation or Synchronization Operation and at most 5 Atomic Operations or
+  Synchronization Operations. We do not make any statement about the
+  ABI-Compatibility of Concurrent Programs outside these bounds.
+* We test Concurrent Programs with a fixed initial state, loop unroll factor
+  (equal to 1 loop unroll), and function calls or recursion. 
+* The above Mappings are not exhaustive, We hope that Arm's partners will
```
walhua, line 1152: we recommend

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+  |                                 +             + |   ``STLXP  W5, X8, X9, [X4]``      +
+  |                                 +             + |   ``CBNZ   W5, loop``              +
+  +                                 +             + | ``MOV   X0, X6``                   +
+  +                                 +             + | ``MOV   X1, X7``                   +
+  +                                 +-------------+--------------------------------------+
+  |                                 | ``LSE``     | ``CASPAL  X0, X1, X2, X3, [X4]``     |
+  +---------------------------------+-------------+--------------------------------------+
+
+
+We do not list other variants of ``fetch_<op>`` since their mappings should be
```
walhua, line 927: Mappings

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+#. Split it into its representative Atomic Operations.
+#. Compile each Atomic Operation separately using a Compiler Profile that
+   generates Assembly Sequences under a given Mapping.
+#. Combine the Assembly Sequences into *multiple* possible Compiled Programs.
+#. Compute the outcomes of executing the Source Concurrent Program under the
+   ISO C memory model. Get source program outcomes *S*.
+#. Compute the outcomes of each compiled program under the AArch64 memory model
+   [ARMARM_]. Get a *set* of compiled program outcomes *C*.
+#. If any *c* in *C* exhibits a compiler bug with respect to the outcomes *S*
+   then the given mappings are not interoperable.
```
walhua, line 1128: Mappings

walhua (commented):
LGTM, just added some trivial comments.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+The Mix Testing Process
+-----------------------
+
+We test for Compiler bugs, a compiler bug is defined as an Outcome of a
+compiled program execution (under the AArch64 model) that is not an Outcome of
+execution of the source Concurrent Program (under the ISO C model). Consider
+the hypothetical example where a source Concurrent Program finishes execution
+in one of three possible outcomes::
+
+  { thread_0:r0=0, thread_1:r0=1 }
```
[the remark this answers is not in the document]
lukeg101: Added a reference

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+ensure compatibility we therefore test for the absence of such Outcomes of the
+compiled programs when mixing all combinations of the above Mappings. We define
+the *Mix Testing* process as follows:
+
+#. Given a C/C++ Concurrent Program.
+#. Split it into its representative Atomic Operations.
+#. Compile each Atomic Operation separately using a Compiler Profile that
+   generates Assembly Sequences under a given Mapping.
+#. Combine the Assembly Sequences into *multiple* possible Compiled Programs.
+#. Compute the outcomes of executing the Source Concurrent Program under the
```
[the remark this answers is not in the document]
lukeg101: There will be once it is public - I will add that reference once it is public. I hope we can merge this document in the meantime

atomicsabi64/atomicsabi64.rst
lukeg101: good spot, thanks!

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+
+Thread Fence 
+   A Thread Fence is a Synchronization Operation that constrains the order of
+   Accesses made by Atomic Operations on a given Thread of Execution. Fences
+   are equipped with a Memory Order Parameter that specifies which kinds of
+   accesses may be reordered before or after the fence. ISO C defines the
+   ``atomic_thread_fence`` to synchronize the order of accesses made by atomic
+   operations on ``_Atomic`` qualified data.
+
+Atomic Instruction
```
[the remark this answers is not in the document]
lukeg101: removed.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+#. Given a C/C++ Concurrent Program.
+#. Split it into its representative Atomic Operations.
+#. Compile each Atomic Operation separately using a Compiler Profile that
+   generates Assembly Sequences under a given Mapping.
+#. Combine the Assembly Sequences into *multiple* possible Compiled Programs.
+#. Compute the outcomes of executing the Source Concurrent Program under the
+   ISO C memory model. Get source program outcomes *S*.
+#. Compute the outcomes of each compiled program under the AArch64 memory model
+   [ARMARM_]. Get a *set* of compiled program outcomes *C*.
+#. If any *c* in *C* exhibits a compiler bug with respect to the outcomes *S*
```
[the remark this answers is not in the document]
lukeg101: clarified

atomicsabi64/atomicsabi64.rst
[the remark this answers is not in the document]
lukeg101: Ah yes, a good spot, my attempts at explaining what is valid wrt ISO C make the narrative confusing. Will rephrase

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+state (where ``x`` and ``y`` are ``0``) produces one of the following outcomes,
+according to the AArch64 Memory Model contained in §B2 of the Arm Architecture
+Reference Manual [ARMARM_]::
+
+  { thread_1:r0=0; [y]=1; }
+  { thread_1:r0=1; [y]=1; }
+  { thread_1:r0=1; [y]=2; }
+
+As such the unexpected outcome has disappeared. There are multiple Mappings
+that exhibit this behaviour, those effected make use of ``SWP`` and ``LD<OP>``
```
smithp35, line 1047: I think `effected` should be `affected`.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+Executing the compiled program on an Arm-based machine from a fixed initial
+state (where ``x`` and ``y`` are ``0``) produces one of the following outcomes,
+according to the AArch64 Memory Model contained in §B2 of the Arm Architecture
+Reference Manual [ARMARM_]::
+
+  { thread_1:r0=0; [y]=1; }
+  { thread_1:r0=1; [y]=1; }
+  { thread_1:r0=1; [y]=2; }
+
+As such the unexpected outcome has disappeared. There are multiple Mappings
```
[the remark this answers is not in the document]
lukeg101: That's actually a good point, so @gonzalobg and @Wilco1 suggested we annotate each mapping in that table with a footnote, paraphrasing 'don't zero out the dest reg', and you're suggesting a pointer to those mappings.

For the purposes of minimising pointer chasing, perhaps I can make a statement after we summarise the cases where this applies, saying 'please refer' to the table starting at line 602 for a complete list of mappings, each is annotated with a footnote'. Then we have both a summary of these cases, and pointers to a complete list. Unless there is a better idea?

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1161 @@
...
+By comparing ``X3`` and the local variable ``r0`` of the original Concurrent
+Program in this example we see there is one additional outcome of executing the
+compiled program that is not an outcome of executing the source program (under
+the respective models). This suggests the Mappings under question are
+incompatible, and a compiler that implements them exhibits a compiler bug. To
+ensure compatibility we therefore test for the absence of such Outcomes of the
+compiled programs when mixing all combinations of the above Mappings. We define
+the *Mix Testing* process as follows:
+
+#. Given a C/C++ Concurrent Program.
```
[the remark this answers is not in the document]
lukeg101: I clarified my description using these, it does not need to be identical

atomicsabi64/atomicsabi64.rst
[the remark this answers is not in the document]
lukeg101: I will explain for each special case what must be done as a 'one liner' before the example

lukeg101:
Apologies for my delay folks, I think I've addressed everything here now with the new version 
[atomicsabi64.pdf](https://github.com/ARM-software/abi-aa/files/15189954/atomicsabi64.pdf)

On the footnote front, I had to use a `*` in a table above the main table of mappings, since the default `[1]_` syntax moves the footnote all the way to the bottom of the document itself.

atomicsabi64/atomicsabi64.rst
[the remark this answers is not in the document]
tmatheson-arm: Personally, I'd prefer we use the full `FEAT_` names throughout. I'm not convinced the space saved is worth potential ambiguity.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+Use-cases known of so far
+-------------------------
+
+
+A Baseline: Describing current implementations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ABI we provide is a baseline specification that compilers should or do implement.
+The ABI provides a grounds to be compatible across all versions of the Armv8 architecture. Most
```
smithp35, line 70: "provides a grounds to be compatible across"  is a bit loose. I think you could use something like
Compilers that implement the baseline specification are compatible across all versions of the Armv8 architecture.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+arise with each new architecture release, and implementers quickly find new candidate mappings
+that are performant on their machines. Such mappings are proposed or added to mainstream
+compilers. However due to the lack of a baseline specification or widespread
+concurrency expertise, testing such mappings has been a challenge and concurrency bugs have been
+unintentionally introduced into compilers when new mappings are added.
+
+We need a baseline ABI in order to determine if a given sub-ABI respects or departs
+from the baseline. Adding command-line options is a logical consequence of defining such an ABI, 
+and makes it possible to track ABI compatibility of concurrent programs at compile or link-time,
+rather than runtime. It is the responsibility of the sub-ABI maintainer to ensure code built
```
smithp35, line 91: I think it is the responsibility of the user of the sub-ABI, the sub-ABI maintainer is responsible for defining the sub-ABI and not responsible for all code using it.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+under their ABI does not mix with code built under the baseline. But a baseline must exist, 
+for sub-ABI compatibility to be decided in the first place.
+
+A baseline provides the means to describe or contain ABI-islands. Where a compiler implementation
```
smithp35, line 95: I'd delete the first sentence. It uses ABI-Islands before introducing the concept.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+departs from the baseline completely (an ABI-island), it would be the responsibility of the
+maintainer of that implementation to ensure their programs are not mixed with programs built for 
+baseline ABI compatibility, or provide adequate warnings at compile time.
```
smithp35, line 98: I'm not sure if I've understood the difference between adding command-line options and an ABI-island.

I think you mean code with extensions to the base-line enabled by command-line, and a completely separate incompatible ABI.

I think in the extension case, Arm cannot provide any statement on the compatibility of the extensions with respect to the baseline specification. In the ABI-island, which could be a known incompatibility with the base-line then users should not mix ABIs, it is QoI whether a toolchain is able to diagnose incompatibility.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+A baseline provides the means to describe or contain ABI-islands. Where a compiler implementation
+departs from the baseline completely (an ABI-island), it would be the responsibility of the
+maintainer of that implementation to ensure their programs are not mixed with programs built for 
+baseline ABI compatibility, or provide adequate warnings at compile time. 
+
+Further, numerous different parties have asked the ABI team whether
+the same atomics mapping is correct. Writing down the known cases helps engineers
+answer these queries without the concurrency expertise required to come up with
+current compatible mappings. A future section of the ABI could document common
```
smithp35, line 103: I don't think the ABI would contain that section, although the design rationale (this document could), will be worth making that explicit.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+.. _ARMARM: https://developer.arm.com/documentation/ddi0487/latest
+.. _PAPER: https://doi.org/10.1109/CGO57630.2024.10444836
+
+Rationale Document for C11 Atomics ABI.
+***************************************
+
+Preamble
+========
+
```
smithp35, line 14: Could you add a Scope section like the BuildAttributes design https://github.com/ARM-software/abi-aa/pull/230/files#diff-f662c6d976647c3fea89e33e36da10a2d57957cb9261b76512a4b35ef0fa3686R157

Just want to capture that this is rationale and not requirements and any ambiguity is resolved in favour of the specfication.

smithp35 (commented):
Thanks very much for splitting out the specification and the rationale. I've given both a read through and have some mostly cosmetic comments.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+document). Precisely, implementations that use loops should use the instructions
+that load or store from memory with the relevant memory order, and the
+appropriate <op> Assembly Sequence inside the loop. Exceptions, where Assembly 
+Sequences exist, are stated (for instance ``fetch_or`` can be implemented using
+``LDSETP`` when the LSE128 extension is enabled).
+
+Special Cases
+-------------
+
+There are special cases in the Mappings presented above, these must be handled
```
stuij, line 1165: `above, these` -> `above. These`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+in order to prevent unexpected outcomes of the compiled program. The special 
+cases are identified below.
```
stuij, line 1167: `below.` -> `below:`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+software if the memory is writeable. Compilers and runtimes should use the
+LSE2/LRCPC3 sequence when available.
+
+
+Declarative statement of Mappings compatibility
+===============================================
+
+To ensure that the above Mappings are ABI-compatible we tested the compilation of
+Concurrent Programs, where each Atomic Operation is compiled to one of the
+aforementioned Mappings. We test if there is a compiled program that exhibits
```
stuij, line 1212: tested/test. We're switching tense. I understand that the sentences express different things, but it still reads as odd.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+an outcome of execution according to the AArch64 Memory Model contained in §B2
+of the Arm Architecture Reference Manual [ARMARM_] that is not an outcome of
+execution of the source program under the ISO C model. In this section we
+define the process by which we test compatibility. 
+
+Definition of ABI-Compatibility for Atomic Operations
+-----------------------------------------------------
+
+*A compiler that implements the above set of Mappings and special cases is ABI-Compatible with
+respect to other compilers that implement the Mappings and special cases.*
```
stuij, line 1222: `implement the` -> `implement these`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+  bounded testing. C/C++ Atomics ABI-compatibility is thus tested for the Mappings
+  above by generating C/C++ Concurrent Programs that permute combinations of
+  Atomic Operations on each Thread of Execution. We bound our test size between
+  2 and 5 Threads of Execution, where each Thread has at least 1 Atomic
+  Operation or Synchronization Operation and at most 5 Atomic Operations or
+  Synchronization Operations. We do not make any statement about the
+  ABI-Compatibility of Concurrent Programs outside these bounds.
+* We test Concurrent Programs with a fixed initial state, loop unroll factor
+  (equal to 1 loop unroll), and function calls or recursion. 
+* The above Mappings are not exhaustive, we recommend that Arm's partners
```
stuij, line 1236: `exhaustive, we` -> `exhaustive. We`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+  submit requests for other Mappings to the ABI team using the `issue tracker page on GitHub <https://github.com/ARM-software/abi-aa/issues>`_.
+* This document makes no statement about the ABI-Compatibility of optimised
+  Concurrent Programs, nor does a statement concerning the performance of
```
stuij, line 1239: `does a statement` -> `does it make a statement`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+compilers. However due to the lack of a baseline specification or widespread
+concurrency expertise, testing such mappings has been a challenge and concurrency bugs have been
+unintentionally introduced into compilers when new mappings are added.
+
+We need a baseline ABI in order to determine if a given sub-ABI respects or departs
+from the baseline. Adding command-line options is a logical consequence of defining such an ABI, 
+and makes it possible to track ABI compatibility of concurrent programs at compile or link-time,
+rather than runtime. It is the responsibility of the sub-ABI maintainer to ensure code built
+under their ABI does not mix with code built under the baseline. But a baseline must exist, 
+for sub-ABI compatibility to be decided in the first place.
```
stuij, line 93: `must exist, for` -> `must exist for`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+A baseline provides the means to describe or contain ABI-islands. Where a compiler implementation
+departs from the baseline completely (an ABI-island), it would be the responsibility of the
+maintainer of that implementation to ensure their programs are not mixed with programs built for 
+baseline ABI compatibility, or provide adequate warnings at compile time. 
+
+Further, numerous different parties have asked the ABI team whether
```
stuij, line 100: remove `different`. numerous parties are automatically different parties.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+queries received by the ABI team, in order to assist implementers and engineers
+with such issues.
+
+Backwards Compatibility and New Architecture Features
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Put another way, a baseline ABI assists in deciding whether new mappings are compatible
+with compiler implementations targeting older versions of the Armv8 architecture.
+Certain instructions (such as Load/Store-Pair instructions [ARMARM_]) have different
+single-copy atomicity guarantees with respect different architecture versions. A baseline
```
stuij, line 113: `respect different` -> `respect to different`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+decides which assembly sequences can be composed correctly (at least as far as testing can decide).
+
+
+Compatibility Between Compilers and Runtimes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The above issues also apply when ensuring object files compiled with different compilers can be mixed. 
+For instance LLVM and GCC code should be interoperable. At the time of writing we identified a number of
+places where this does not apply, both when compiling to target the same architecture version, and mixing
```
stuij, line 122: `and mixing` -> `and when mixing`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+different (compatible) architecture versions. Further, the above is not limited to statically compiled code. We found
+one instance where proposed mappings implemented in a JiT compiler would not be interoperable with respect
```
stuij, line 124: remove `respect to`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+Under ISO C, the above Concurrent Program finishes execution in one of three
+possible outcomes (a reference for this notation is found here [PAPER_])::
+
+  { thread_1:r0=0; y=1; }
+  { thread_1:r0=1; y=1; }
+  { thread_1:r0=1; y=2; }
+
+In this case the value read by the exchange on ``thread_1`` is not used, and a
+compiler is free to remove references to unused data. It is not legal according
+to this ABI for a compliant implementation piler to translate the program into
```
stuij, line 171: `implementation piler` -> `compiler` or `implementation`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+  { thread_1:r0=0; [y]=1; }
+  { thread_1:r0=0; [y]=2; } <-- Forbidden by source model, a bug!
+  { thread_1:r0=1; [y]=1; }
+  { thread_1:r0=1; [y]=2; }
+
+By comparing ``W3`` and the local variable ``r0`` of the original Concurrent
+Program we see there is one additional outcome of executing the compiled
+program that is not an outcome of executing the Concurrent Program. This is due
+to the fact that according to the Arm Architecture Reference Manual [ARMARM_]
```
stuij, line 208: the embedded quite is grammatically still part of the sentence. And so the sentence still needs to be grammatically correct. It isn't in this case.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+By comparing ``W3`` and the local variable ``r0`` of the original Concurrent
+Program we see there is one additional outcome of executing the compiled
+program that is not an outcome of executing the Concurrent Program. This is due
+to the fact that according to the Arm Architecture Reference Manual [ARMARM_] 
+*instructions where the destination register is WZR or XZR, are not regarded as
+doing a read for the purpose of a DMB LD barrier.*
+
+In this case the compiler introduces another outcome of Execution. To fix this
+issue, a compiler is not permitted to rewrite the destination register to be the
+zero register in this case::
```
stuij, line 214: twice `in this case`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+state (where ``x`` and ``y`` are ``0``) produces one of the following outcomes,
+according to the AArch64 Memory Model contained in §B2 of the Arm Architecture
+Reference Manual [ARMARM_]::
+
+  { thread_1:r0=0; [y]=1; }
+  { thread_1:r0=1; [y]=1; }
+  { thread_1:r0=1; [y]=2; }
+
+As such the unexpected outcome has disappeared. There are multiple Mappings
+that exhibit this behaviour, those affected make use of ``SWP`` and ``LD<OP>``
```
stuij, line 238: `behaviour,` -> `behaviour.`, or fix this sentence in another way.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+and the possibility of user error. We believe an ABI is necessary to document
+this corner of code generation.
+
+
+The Mix Testing Process
+-----------------------
+
+ABI compatibility must be testable. Concurrency is not trivial, and the ABI
+presents a simplification of part of the problem that is understandable by
+engineers. We provide novel, yet simple, techniques and tools for
```
stuij, line 257: `novel, yet simple,` -> `simple`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+testing ABI compatibility. These techniques reduce the difficulty of checking
+compatibility from a problem of understanding concurrent executions, to the
+familiar testing domain of comparing program outcomes of tests. This document
+does not preclude other means of testing compatibility however.
```
stuij, line 261: remove `however`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+We test for Compiler bugs, a Compiler Bug is defined as an outcome of a
```
stuij, line 263: `bugs, a` -> `bugs. A`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+an outcome of execution of the source Concurrent Program (under the 
+ISO C memory model). Consider the hypothetical example where a source
+Concurrent Program finishes execution in one of three possible outcomes
+(a reference for this notation is found here [PAPER_])::
+
+  { thread_0:r0=0, thread_1:r0=1 }
+  { thread_0:r0=1, thread_1:r0=0 }
+  { thread_0:r0=1, thread_1:r0=1 }
+
+and one possible compiled program outcome has the following according to the
```
stuij, line 275: has the following what?

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+By comparing ``X3`` and the local variable ``r0`` of the original Concurrent
+Program in this example we see there is one additional outcome of executing the
+compiled program that is not an outcome of executing the source program (under
+the respective models). This suggests the Mappings under question are
+incompatible, and a compiler that implements them exhibits a Compiler Bug. To
+ensure compatibility we therefore test for the absence of such outcomes of the
+compiled programs when mixing all combinations of the above Mappings. We define
+the *Mix Testing* process as follows:
+
+#. Take an arbitrary Concurrent Program, when executed on the C/C++ memory
```
stuij, line 293: `Program, when` -> `Program. When`

and

`model will` -> `model, it will`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+#. Compile each individual source file containing an Atomic Operation 
+   using each Compiler Profile under test that generates Assembly Sequences
+   under a given Mapping.
+#. Combine the Assembly Sequences from above into *multiple* possible Compiled
+   Programs.
+#. Compute the outcomes of each compiled program under the AArch64 Memory Model
+   contained in §B2 of the Arm Architecture Reference Manual [ARMARM_]. Get a
+   *set* of compiled program outcomes *C*.
+#. If any compiled program set of outcomes *c* in *C* exhibits a Compiler Bug
+   (Check that *c* is a subset of *S*) with then the given Mappings are not
```
stuij, line 306: `with then` -> `,`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,308 @@
...
+- The mappings should cover atomic accesses of various sign, size, and type accessible through C11 atomic operations using compiler profiles.
+
+The main trade-offs we have identified or have been made aware of are:
+
+- Performance of different mappings versus compatibility with all architectures.
+- Whether certain compiler operations lead to unexpected behaviours.
+
+As motivated by the use cases expanded upon below:
+
+- The need for a baseline ABI
```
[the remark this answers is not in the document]
sallyarmneale: that is correct. Full stop if the bullet contains a full sentence.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+Preamble
+========
+
+Abstract
+--------
+
+This document describes the C/C++ Atomics Application Binary Interface for the
+Arm 64-bit architecture. This document concerns the valid Mappings from C/C++
+Atomic Operations to sequences of A64 instructions. For matters concerning the
+memory model, please consult §B2 of the Arm Architecture Reference Manual
```
sallyarmneale, line 51: `For further information on the memory model, refer to the Arm......`

Keep sentences as simple as possible.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+
+Synchronization Operation
+   The order that atomic operations are executed by each Thread of Execution
+   may not be the same as the order they are written in the program.
+   Synchronization Operations are statements that constrain the order of
+   accesses made to Shared-Memory Locations by each thread. Synchronization
+   Operations include Thread Fences.
+
+Shared-Memory Location
+   A memory location that can be accessed by any Thread of Execution in the
```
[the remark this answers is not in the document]
Wilco1: Fixed, I've changed it to Thread and simplified definitions.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+   PageBreak
+
+Overview
+========
+
+The C/C++ Atomics ABI for the Arm 64-bit architecture (AABI64) comprises the
+following sub-components.
+
+* The `Mappings from Atomic Operations to Assembly Sequences`_, which defines
+  the Mappings from C/C++ atomic operations to sto one of more Assembly
```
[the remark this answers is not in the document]
Wilco1: Fixed

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+.. |copyright-date| replace:: 2024
+.. |footer| replace:: Copyright © |copyright-date|, Arm Limited and its
+                      affiliates. All rights reserved.
+
+.. _ARMARM: https://developer.arm.com/documentation/ddi0487/latest
+.. _AAELF64: https://github.com/ARM-software/abi-aa/releases
+.. _CPPABI64: https://github.com/ARM-software/abi-aa/releases
+.. _CSTD: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf
+.. _PAPER: https://doi.org/10.1109/CGO57630.2024.10444836
+
```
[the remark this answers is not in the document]
Wilco1: Fixed

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+  +-------------+--------------------------------------------------------------+-----------------------------------------------------------------------------+
+  | CSTD_       | ISO/IEC 9899:2018                                            | International Standard ISO/IEC 9899:2018 – Programming languages C.         |
+  +-------------+--------------------------------------------------------------+-----------------------------------------------------------------------------+
+  | AAELF64_    | ELF for the Arm 64-bit Architecture (AArch64)                | ELF for the Arm 64-bit Architecture (AArch64)                               |
+  +-------------+--------------------------------------------------------------+-----------------------------------------------------------------------------+
+  | PAPER_      | CGO paper                                                    | Compiler Testing with Relaxed Memory Models                                 |
+  +-------------+--------------------------------------------------------------+-----------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
Wilco1: Fixed

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+* The above Mappings are not exhaustive, we recommend that Arm's partners
+  submit requests for other Mappings to the ABI team using the `issue tracker page on GitHub <https://github.com/ARM-software/abi-aa/issues>`_.
+* This document makes no statement about the ABI-Compatibility of optimised
+  Concurrent Programs, nor does a statement concerning the performance of
+  compiled programs under the above Mappings when executed on a given Arm-based
+  machine.
+* This document makes no statement about the ABI-Compatibility of compilers
+  that implement Mappings other than what is stated in this document.
+
+Appendix: Mix Testing
```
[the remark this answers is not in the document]
Wilco1: I've removed this section for now, it can be added to rationale.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+========
+
+Abstract
+--------
+
+This document describes the C/C++ Atomics Application Binary Interface for the
+Arm 64-bit architecture. This document concerns the valid Mappings from C/C++
+Atomic Operations to sequences of A64 instructions. For matters concerning the
+memory model, please consult §B2 of the Arm Architecture Reference Manual
+[ARMARM_]. We focus only on a subset of the C11 atomic operations at the time
```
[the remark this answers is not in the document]
Wilco1: Removed the whole sentence since we're not doing a subset here.

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1252 @@
...
+
+Preamble
+========
+
+Abstract
+--------
+
+This document describes the C/C++ Atomics Application Binary Interface for the
+Arm 64-bit architecture. This document concerns the valid Mappings from C/C++
+Atomic Operations to sequences of A64 instructions. For matters concerning the
```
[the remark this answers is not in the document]
Wilco1: Removed

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,302 @@
...
+an outcome of execution of the source Concurrent Program (under the 
+ISO C memory model). Consider the hypothetical example where a source
+Concurrent Program finishes execution in one of three possible outcomes
+(a reference for this notation is found here [PAPER_])::
+
+  { thread_0:r0=0, thread_1:r0=1 }
+  { thread_0:r0=1, thread_1:r0=0 }
+  { thread_0:r0=1, thread_1:r0=1 }
+
+and one possible compiled program outcome has the following outcomes
```
stuij, line 269: `one possible compiled program outcome has the following outcomes`. double outcomes.

Could this be rephrased to: `one compiled program execution run has the following possible outcomes`

atomicsabi64/atomicsabi64.rst
```diff
@@ -0,0 +1,1087 @@
...
+   Arm considers this specification to be complete, but existing
+   implementations do not meet the requirements for confidence in its release
+   quality. Arm may need to make incompatible changes if issues emerge from its
+   implementation.
+
+**Alpha**
+   The content of this specification is a draft, and Arm considers the
+   likelihood of future incompatible changes to be significant.
+
+All content in this document is at the **Release** quality level.
```
stuij, line 206: This should be `**Alpha** quality` I believe.

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,319 @@
...
+    STR W3,[X4]
+
+  thread_1:
+    MOV W9,#2
+    SWP W9, WZR, [X2]
+    DMB ISHLD
+    LDR W3,[X4]
+
+where ``thread_0:X2`` contains the address of ``x``, ``thread_0:X4`` contains
+the address of ``y``, and ``thread_1:X2`` contains the address of ``y``,
```
stuij, line 198: `and` should move to the last clause:

`the address of ``y``, and ``thread_1:X2`` contains the address of ``y``, ``thread_1:X4`` contains` 

-> `the address of ``y``,  ``thread_1:X2`` contains the address of ``y``, and ``thread_1:X4`` contains`

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,319 @@
...
+and the possibility of user error. We believe an ABI is necessary to document
+this corner of code generation.
+
+
+The Mix Testing Process
+-----------------------
+
+ABI compatibility must be testable. Concurrency is not trivial, and the ABI
+presents a simplification of part of the problem that is understandable by
+engineers. We provide a simple technique for testing ABI compatibility.
```
stuij, line 268: mixing of singular and plural: `a simple technique`. Next sentence: `these techniques`. Which one is it?

design-documents/atomics-ABI.rst
```diff
@@ -0,0 +1,319 @@
...
+
+- We need to choose a baseline ABI (a set of mappings), that is compatible for all versions of the Armv8 architecture.
+- The mappings should cover atomic accesses of various sign, size, and type accessible through C11 atomic operations using compiler profiles.
+
+We have identified the following trade-offs:
+
+- Performance of different mappings versus compatibility with all architectures.
+- Whether certain compiler operations lead to unexpected behaviours.
+
+As motivated by the use cases expanded upon below:
```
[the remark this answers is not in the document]
lukeg101: I changed to `The use cases expanded upon below motivate why we need an atomics abi:`

stuij (approved):
Thanks Luke and Wilco for your patience with the barrage of comments. Looks like all the reviewers are happy with the current state of the docs right now.

If there are no objections, I'll push this tomorrow.

smithp35 (approved):
LGTM thanks very much for all the hard work.

PR #259, opened 2024-04-22
[PAUTHABIELF64] Add R_AARCH64_AUTH_GOT_ADR_PREL_LO21 relocation

smithp35:
With the tiny code model and a signed GOT, an adr instruction is needed to get the address of the GOT entry for input to the authenication. For example:

adr x8, :got_auth: symbol
ldr x0, [x8]
// Authenticate to get unsigned pointer
autia x0, x8

The adr requires a new relocation code where there isn't a direct equivalent in the main ABI as there is not need to take the address of the GOT slot when no authentication is required.

We define R_AARCH64_AUTH_GOT_ADR_PREL21_LO21 for this purpose following the naming convention of R_<CLS>_ADR_PREL_LO21. which is its closest equivalent.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -1069,7 +1069,7 @@ GOT entry as the modifier. The static linker must encode the signing
 schema into the GOT slot. AUTH variant dynamic relocations must be
 used for signed GOT entries.
 
-Example Code to access a signed GOT entry
+Example Code to access a signed GOT entry with the small code model:
 
 .. code-block:: asm
 
@@ -1081,6 +1081,18 @@ Example Code to access a signed GOT entry
 
 In the example the :got_auth: and :got_auth_lo12: operators result in
 AUTH variant GOT generating relocations being used.
+
+Example Code to access a signed GOT entry with the tiny code model:
+
+.. code-block:: asm
+
+  adr x8, :got_auth: symbol
+  ldr x0, [x8]
+  // Authenticate to get unsigned pointer
+  autia x0, x8
+
+Compared to the tiny code model without pointer authentication an
+additonal adr is required to get the address of the GOT entry.
 
 AUTH variant GOT Generating Relocations
 ---------------------------------------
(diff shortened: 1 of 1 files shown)


pauthabielf64/pauthabielf64.rst
```diff
@@ -1092,6 +1092,18 @@ Example Code to access a signed GOT entry
...
+
+.. code-block:: asm
+
+  adr x8, :got_auth: symbol
+  ldr x0, [x8]
+  // Authenticate to get unsigned pointer
+  autia x0, x8
+
+Compared to the tiny code model without pointer authentication an
+additonal adr is required to get the address of the GOT entry.
```
Wilco1, line 1105: I'm not sure I am following here - all PLT sequences as currently defined(1) must compute the address of the GOT entry for the dynamic linker. So the ADR is required even without PAUTH, and it would be worth having the equivalent relocation if it doesn't already exist (given that just having PC-relative LDR won't work).

(1) It is possible to define PLT sequences that compute the address of the GOT entry in a different way.

pauthabielf64/pauthabielf64.rst
```diff
@@ -1167,6 +1179,11 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
   |             |                                        |                                  | value to bits [11:0] of  |
   |             |                                        |                                  | X. No overflow check.    |
   +-------------+----------------------------------------+----------------------------------+--------------------------+
+  | 0x811D      | R\AARCH64\_AUTH\_GOT\_ADR\_PREL\_LO21  | G(ENCD(GDAT(S + A))) - P         | Set the immediate        |
```
Wilco1, line 1182: Remove the +A

pauthabielf64/pauthabielf64.rst
Wilco1: You're right, there doesn't seem a case for user code, so if linkers use internal relocations then there is no need to add one.

asl:
We double checked and aligned all names in the implementation according to the spec (see https://github.com/kovdan01/llvm-project/commit/029ba553bda1520d2c2cae8d2c1fae152972d088)

smithp35:
Thanks for the confirmation.

sallyarmneale (commented):
No changes required

PR #261, opened 2024-05-05
Fix the typo in Operation Definition for R_ARM_REL32

eymay:
R_ARM_REL32 relocation's operation is defined as ((S + A) | T) – P but an extra "|" is left in the current version.


smithp35 (approved):
Thanks very much for spotting the problem and for the fix.

PR #264, opened 2024-06-07
[SME] Add agnostic-ZA interface and routines to save/restore SME state

sdesmalen-arm:
This PR adds a new "agnostic-ZA" interface which is intended to be called from any subroutine without requiring a change to PSTATE.ZA. This PR also adds new SME ABI routines to save/restore state enabled by PSTATE.ZA.

The corresponding ACLE PR can be found [here](https://github.com/ARM-software/acle/pull/336).

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -1749,10 +1749,11 @@ ZA interfaces
 As noted in `ZA states`_, there are three possible ZA states: off,
 dormant, and active.  A subroutine's “ZA interface” specifies the possible
 states of ZA on entry to a subroutine and the possible states of ZA on a
-`normal return`_.  The AAPCS64 defines two types of ZA interface:
+`normal return`_.  The AAPCS64 defines three types of ZA interface:
 
 .. _`private-ZA`:
 .. _`shared-ZA`:
+.. _`compatible-ZA`:
 
 +-------------------+-------------------+---------------------------+
 | Type of interface | ZA state on entry | ZA state on normal return |
@@ -1760,6 +1761,9 @@ states of ZA on entry to a subroutine and the possible states of ZA on a
 | private ZA        | dormant or off    | unchanged or off          |
 +-------------------+-------------------+---------------------------+
 | shared ZA         | active            | active                    |
++-------------------+-------------------+---------------------------+
+| compatible ZA     | active, dormant   | unchanged                 |
+|                   | or off            |                           |
 +-------------------+-------------------+---------------------------+
 
 Every subroutine has exactly one ZA interface.  A subroutine's ZA interface
@@ -1776,8 +1780,12 @@ The shared-ZA interface is so called because it allows the subroutine
 to share ZA contents with its caller.  This can be useful if an SME
 operation is split into several cooperating subroutines.
 
-Subroutines with a `private-ZA`_ interface and subroutines with a `shared-ZA`_
-interface can both (at their option) choose to guarantee that they
+The compatible-ZA interface is intended to be called from any function
+without requiring a change to PSTATE.ZA and is generally used in conjunction
+with the expectation that it `preserves ZA`_.
+
+Subroutines with a `private-ZA`_ interface, `shared-ZA`_ interface or
+`compatible-ZA`_ interface can (at their option) choose to guarantee that they
 `preserve ZA`_.
 
 Parameter passing
@@ -2081,6 +2089,17 @@ support routines:
 ``__arm_get_current_vg``
    Provides a safe way to detect the current value of VG.
 
+``__arm_sme_state_size``
+   Provides a simple way to query the total size required to save the requested
+   state.
+
+``__arm_sme_save``
+   Provides a safe way to save all state enabled by PSTATE.ZA.
+
+``__arm_sme_restore``
+   Provides a safe way to restore all state enabled by PSTATE.ZA from a buffer.
+
+
 ``__arm_sme_state``
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


aapcs64/aapcs64.rst
```diff
@@ -2069,6 +2077,8 @@ support routines:
 
    * the current values of TPIDR2_EL0, PSTATE.SM and PSTATE.ZA.
 
+   * the registers enabled by PSTATE.ZA, if PSTATE.ZA is enabled.
```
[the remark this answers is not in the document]
sdesmalen-arm: Sorry this should not have been committed. I've removed the change.

rsandifo-arm (commented):
It isn't clear to me whether we expect callers of ``__arm_sme_state_size`` to check for zero returns, and skip the save and restore calls for that case.  If so (“case A”), I think this makes things unnecessarily complex for the callers.  If not (“case B”), the buffer should have nonzero size even if PSTATE.ZA==0, so that it can hold at least the `SAVED_ZA` and `SAVED_ZT0` fields.

And for case B, the save and restore functions should be callable even in threads that don't have access to SME.

In a similar vein, how about making ``__arm_sme_restore`` decide what to do based only on the ``SAVED_*`` fields, rather than passing ``OPTIONS`` again?  The caller of the save and restore functions is expected to be the same, so it should know what needs to be preserved.

I wonder how useful the ``OPTIONS`` will be in practice, given that ``__arm_sme_save`` must leave PSTATE.ZA in a dormant or off state.  It seems unlikely that a caller would know enough about its position in the call hierarchy/overall program to know that certain state doesn't need to be preserved, but at the same time not know enough for it to have a specific `__arm_in/out/inout/preserves` annotation.  No objection to keeping ``OPTIONS`` if you prefer though.

sdesmalen-arm:
> It isn't clear to me whether we expect callers of `__arm_sme_state_size` to check for zero returns, and skip the save and restore calls for that case. If so (“case A”), I think this makes things unnecessarily complex for the callers. If not (“case B”), the buffer should have nonzero size even if PSTATE.ZA==0, so that it can hold at least the `SAVED_ZA` and `SAVED_ZT0` fields.
> 
> And for case B, the save and restore functions should be callable even in threads that don't have access to SME.

The initial thinking was "case A", i.e. the intrinsics are only called when there is something worth saving. But now that I've added SAVED_ZA and SAVED_ZT0, it probably makes sense to remove that assumption.

> I wonder how useful the `OPTIONS` will be in practice, given that `__arm_sme_save` must leave PSTATE.ZA in a dormant or off state. It seems unlikely that a caller would know enough about its position in the call hierarchy/overall program to know that certain state doesn't need to be preserved, but at the same time not know enough for it to have a specific `__arm_in/out/inout/preserves` annotation. No objection to keeping `OPTIONS` if you prefer though.

My reasons for adding `OPTIONS` was to give an interface that allows the routines to be used more generically, e.g. to support implementing the following case more easily using these intrinsics:
```
void bar() __arm_inout("zt0");
void foo() __arm_inout("za", "zt0") { bar(); }
```
where only "za" needs preserving across the call, but not "zt0". This could be done using explicit allocas, calling the `__arm_tpidr2_*` routines, but this functionality may become more involved in the future should more state be added to PSTATE.ZA.

> In a similar vein, how about making `__arm_sme_restore` decide what to do based only on the `SAVED_*` fields, rather than passing `OPTIONS` again? The caller of the save and restore functions is expected to be the same, so it should know what needs to be preserved.

The reasoning for that was that one might only want to restore part of the saved state at one point, and restore the other part at another point. The compiler or user writing asm could use these routines to save/restore any state, e.g.
```
void bar();
void use_zt0() __arm_inout("zt0");
void use_za() __arm_inout("za");

void foo() __arm_inout("za", "zt0") {
               // preserve both zt0 and za, could use the __arm_sme_save routine for this.
  bar();
               // reload only zt0
  use_zt0();
               // save only zt0 (za was already saved)
  bar();
               // restore za
  use_za();
               // restore zt0
}
```

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2324,221 @@ value of VG, with the subroutine having the following properties:
...
+    * The full contents of ``TPIDR2_EL0`` are written to ``PTR->TPIDR2_EL0``.
+
+    * The address ``PTR->ZA`` is written to ``PTR->BLK.za_save_buffer``,
+      the streaming vector length in bytes (``SVL.B``) is written to
+      ``PTR->BLK.num_za_save_slices`` and the address ``PTR->BLK`` is
+      written to ``TPIDR2_EL0``, thus setting up a lazy save.
+
+    * The value 1 is written to ``PTR->SAVED_ZA``.
+
+  * If bit 1 of ``OPTIONS`` is 0 and ZT0 is available, then for the addresses
```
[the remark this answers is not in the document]
sdesmalen-arm: Good spot, you're right!

rsandifo-arm:
> > I wonder how useful the `OPTIONS` will be in practice, given that `__arm_sme_save` must leave PSTATE.ZA in a dormant or off state. It seems unlikely that a caller would know enough about its position in the call hierarchy/overall program to know that certain state doesn't need to be preserved, but at the same time not know enough for it to have a specific `__arm_in/out/inout/preserves` annotation. No objection to keeping `OPTIONS` if you prefer though.
> 
> My reasons for adding `OPTIONS` was to give an interface that allows the routines to be used more generically, e.g. to support implementing the following case more easily using these intrinsics:
> 
> ```
> void bar() __arm_inout("zt0");
> void foo() __arm_inout("za", "zt0") { bar(); }
> ```
> 
> where only "za" needs preserving across the call, but not "zt0". This could be done using explicit allocas, calling the `__arm_tpidr2_*` routines, but this functionality may become more involved in the future should more state be added to PSTATE.ZA.

I don't think the new routines work for that case though.  The routines rely on saving ZA lazily, whereas the above needs to save ZA eagerly.  `bar` is entitled to assume that ZA is active rather than dormant on entry.

sdesmalen-arm:
> > > I wonder how useful the `OPTIONS` will be in practice, given that `__arm_sme_save` must leave PSTATE.ZA in a dormant or off state. It seems unlikely that a caller would know enough about its position in the call hierarchy/overall program to know that certain state doesn't need to be preserved, but at the same time not know enough for it to have a specific `__arm_in/out/inout/preserves` annotation. No objection to keeping `OPTIONS` if you prefer though.
> > 
> > 
> > My reasons for adding `OPTIONS` was to give an interface that allows the routines to be used more generically, e.g. to support implementing the following case more easily using these intrinsics:
> > ```
> > void bar() __arm_inout("zt0");
> > void foo() __arm_inout("za", "zt0") { bar(); }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where only "za" needs preserving across the call, but not "zt0". This could be done using explicit allocas, calling the `__arm_tpidr2_*` routines, but this functionality may become more involved in the future should more state be added to PSTATE.ZA.
> 
> I don't think the new routines work for that case though. The routines rely on saving ZA lazily, whereas the above needs to save ZA eagerly. `bar` is entitled to assume that ZA is active rather than dormant on entry.

Yes, that was the reason I initially added the "lazily" as an option to the struct. I figured it should still be possible to set up the lazy-save using these routines and then commit the lazy-save manually before the call to `bar` and then restore afterwards, if ZA needs saving eagerly. Or is there some subtlety that I'm missing?

rsandifo-arm:
> Yes, that was the reason I initially added the "lazily" as an option to the struct. I figured it should still be possible to set up the lazy-save using these routines and then commit the lazy-save manually before the call to `bar` and then restore afterwards, if ZA needs saving eagerly. Or is there some subtlety that I'm missing?

The difference for me is that the new routines are fundamentally about processing the live ZA state (and any future SME state) such that it's safe to call a normal function.  In some cases this will involve turning PSTATE.ZA off.  This is different from what the mixed-sharing example needs: there the intention is explicitly to keep ZA on (and active).

The new routines are also about providing an interface that can handle unknown future state.  Thus the assumption is that extra state must be stored unless explicitly turned off (via ``OPTIONS``).  This is different from the example of moving between different sharing types, since there the compiler (by necessity) knows the state that is live in the caller and the state that is shared with the callee.  The compiler only wants to store the difference between the two, and not any future unknown state.

When the callee shares some ZA state, but shares less state than the caller, the caller has to do an eager save and restore of the remaining state.  I think we should just treat this as being conceptually like any other caller save.  Admittedly the save and restore are quite heavy operations for ZA, but in concept they're not much different from a register save and restore.  Treating them like that also helps with your later example involving calls to separate `__arm_inout("zt0")` and `__arm_inout("za")` functions.

rsandifo-arm (commented):
Sorry for only raising this now, but I wonder if we should reconsider the case in which ZA is dormant on entry to an agnostic-ZA function.  Rather than say that agnostic-ZA functions must preserve the state (and so leave ZA dormant), should we instead say that agnostic-ZA behaves like private-ZA for that case?  No other PSTATE.ZA state can be live if ZA is dormant.  In particular, if ZA is dormant then ZT0 must be dead.

In other words, agnostic-ZA is like private-ZA except that ZA can be active on entry.  If ZA is active on entry, it must be unchanged on return.

(If in future there is any other SME state that needs to be preserved, and that is not controlled by PSTATE.ZA, then that would still need to be saved and restored.)

The dormant-on-entry case isn't too interesting for direct calls between streaming code and agnostic-ZA functions, since the streaming code wouldn't be expected to set up a lazy save in that case.  But perhaps it is more relevant if streaming code calls a normal function and that normal function calls a private-ZA function.  (Ideally, the normal function would be changed to private-ZA, but that might not always be possible for ABI reasons.)

Other than that, it looks good to me.  Some minor comments below.

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+
+* The subroutine does not return a value.
+
+* The subroutine behaves as follows:
+
+  * If ``PTR`` does not point to a valid buffer with the required size, the
+    behavior of calling this subroutine is undefined.
+
+  * If ZA state is 'active' on entry, then it is 'dormant' on normal return.
+    Otherwise the ZA state is unchanged.
```
rsandifo-arm, line 2404:
```suggestion
  * If ZA state is `active <ZA active state>`_ on entry, then it is
    `dormant <ZA dormant state>`_ on normal return.  Otherwise the ZA state
    is unchanged.
```

(untested) with a new internal target for “ZA dormant state”.

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+
+  * For the address ``PTR->VALID`` at an unspecified offset in the buffer,
+    the value 0 is written to ``PTR->VALID`` and the subroutine returns, if
+    either of the following conditions is true:
```
rsandifo-arm, line 2408:
```suggestion
    one of the following conditions is true:
```

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+
+  The exact layout used to calculate the size is unspecified.  The
+  implementations of `__arm_sme_save`_ and `__arm_sme_restore`_ and
+  `__arm_sme_state_size`_ must all assume the same layout.
+
+  The size is guaranteed to be a multiple of 16.
+
+* The subroutine behaves as follows:
+
+  * If the current thread has access to FEAT_SME and PSTATE.ZA is 1,
```
rsandifo-arm, line 2364: Maybe “ZA is active” (perhaps with the same kind of internal link as the comment below) now that the routines don't save anything when ZA is dormant.

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+  PTR
+    a 64-bit data pointer passed in X0 that points to a buffer which
+    is guaranteed to have a size that is equal to or larger than the size
+    returned by `__arm_sme_state_size`_. The pointer must be 16-byte aligned.
+
+* The subroutine does not return a value.
+
+* The subroutine behaves as follows:
+
+  * If ``PTR`` does not point to a valid buffer with the required size, the
```
rsandifo-arm, line 2400: Perhaps we should explicitly require the routine to “abort the program in some platform-defined manner” if ``PTR`` is not 16-byte aligned.  This would allow the low 4 bits to be used for future extensions.  Same for the restore routine.

This would then be “Otherwise, if …”

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+    aligned.
+
+* The subroutine does not return a value.
+
+* The subroutine behaves as follows:
+
+  * If ``PTR`` does not point to a valid buffer with the required size, the
+    behavior of calling this routine is undefined.
+
+  * The ZA state on normal return is the same as the ZA state on entry to the
```
rsandifo-arm, line 2461: This is no longer true if ZA was dormant or off, since the save might have been committed between the calls.  Perhaps we should just delete this bullet point and deal with the various cases below.  E.g. the next bullet point deals with the case where ZA was dormant or off on entry to the save function.

aapcs64/aapcs64.rst
```diff
@@ -2305,6 +2328,188 @@ value of VG, with the subroutine having the following properties:
...
+  * For the address ``PTR->VALID`` at an unspecified offset in the buffer,
+    if the value stored at address ``PTR->VALID`` is 0, then the subroutine
+    does nothing.
+
+  * Otherwise, the subroutine aborts in some platform-specific manner if
+    either of the following conditions is true:
+
+    * The current thread does not have access to SME.
+
+    * ZA state is 'active' on entry.
```
rsandifo-arm, line 2474:
```suggestion
    * ZA state is `active <ZA active state>`_ on entry.
```

aapcs64/aapcs64.rst
sdesmalen-arm: For some reason, these links don't work for me when I generate the PDF (active or dormant), even though I've spelled them correctly.

rsandifo-arm: OK, how about using:

```
   If the `ZA state is active`_ on entry, then it is dormant on return. …
```

with:

```
active
   .. _`ZA active state`:

   .. _`ZA state is active`:

```

Hopefully the active link is a good enough clue regarding “dormant”.

Same for ``__arm_sme_state_size``, in particular with “the ZA state”.

sallyarmneale (commented):
Looks good to me

PR #48, opened 2020-11-05
Import AAELF32 into this repository.

statham-arm:
This is the first document that uses the 'rubric' markup element, so
I've added a style stanza for it in rst2pdf-abi.style. Also updated
links in other documents to link to the new location.

Issue #16, opened 2020-01-30
Changelog tables in ABI


fpetrogalli:
I think we should convert the tables that track the changes in the ABI specifications in sections + paragraphs. At the moment, we see changes even in the rows describing the previous versions just because we need to reformat the table. I think this could be error prone (changes could be missed).

As an example of what I mean for "reformatting", see this PR:

https://github.com/ARM-software/software-standards/pull/15/files#diff-b590b56110000b706ad06b2250159441

stuij:
This is a potential issue with any table. I personally think the table layout for the change log is quite nice in that it gives a nice clear view of the changes, and I'm not sure if the effort is worth the outcome.

fpetrogalli:
> This is a potential issue with any table. 

Yes, but the other tables will rarely change. The changelog table will change every time we do a release of a document after changing it.

stuij:
The main reason we see so many tables change in for this release is because the tables needed classes and therefore whole-table readjustments. In rare occasions I widened some columns so it's easier to add info. But in general the column width should stay static, either for changelog tables or other tables. Any changes to individual rows will of course only affect those rows, and for changelog tables this should only be new release info. And if it's just the table widening, and there are no other changes around the table, most diffing tools, including the github one will show incremental changes in a deeper color, which should make it clear that we're not accidentally changing things we don't want.

fpetrogalli:
@stuij , I think you gave enough arguments to leave things the way they are. Closing this. Thank you.

PR #271, opened 2024-07-01
[AAELF64] Clarify how addends work in MOVZ, MOVK and ADRP.

statham-arm:
This brings AAELF64 into line with AAELF32, which already has a similar clarification for the MOVW+MOVT pair. For the instructions which shift their operand left (ADRP, and the shifted MOVZ and MOVK), if the relocation addend is taken from the input value of the immediate field, it is not treated as shifted.

The rationale is that this allows a sequence of related instructions to consistently compute the same value (symbol + small offset), and cooperate to load that value into the target register, one small chunk at a time. For example, this would load `mySymbol + 0x123`:

  mov  x0, #0x123          ; R_AARCH64_MOVW_UABS_G0_NC(mySymbol)
  movk x0, #0x123, lsl #16 ; R_AARCH64_MOVW_UABS_G1_NC(mySymbol)
  movk x0, #0x123, lsl #32 ; R_AARCH64_MOVW_UABS_G2_NC(mySymbol)
  movk x0, #0x123, lsl #48 ; R_AARCH64_MOVW_UABS_G3(mySymbol)

The existing text made it unclear whether the addends were shifted or not. If they are interpreted as shifted, then nothing useful happens, because the first instruction would load the low 16 bits of `mySymbol+0x123`, and the second would load the next 16 bits of `mySymbol+0x1230000`, and so on. This doesn't reliably get you _any_ useful offset from the symbol, because the relocations are processed independently, so that a carry out of the low 16 bits won't be taken into account in the next 16.

If you do need to compute a large offset from the symbol, you have no option but to use SHT_RELA and specify a full 64-bit addend: there's no way to represent that in an SHT_REL setup. But interpreting the SHT_REL addends in the way specified here, you can at least specify _small_ addends successfully.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -940,6 +940,15 @@ A ``RELA`` format relocation must be used if the initial addend cannot be encode
 
 There is no PC bias to accommodate in the relocation of a place containing an instruction that formulates a PC- relative address. The program counter reflects the address of the currently executing instruction.
 
+There are two special cases for forming the initial addend of REL-type relocations where the immediate field cannot normally hold small signed integers:
+
+* For relocations processing MOVZ and MOVK instructions (including the "MOV (wide immediate)" alias), the initial addend is formed by interpreting the 16-bit literal field of the instruction as a 16-bit signed value in the range -32768 <= A < 32768. The interpretation is the same whether or not the instruction applies a left shift to its immediate: the addend is never treated as shifted.
+
+* For relocations processing the ADRP instruction, the initial addend is similarly formed by interpreting the literal field of the instruction as a 21-bit signed integer, in the range -1048576 <= A < 1048576. The ADRP instruction's implicit left shift of 12 bits is not applied.
+
+These special cases permit a sequence of instructions to each add the same small constant to a symbol's value, and extract separate ranges of bits from the sum, so that the instruction sequence as a whole consistently loads the full result of the addition.
+
+In the case of a sequence using ADRP followed by a 12-bit ADD to set up the low bits of the offset, you can express an offset up to 1048576 in either direction, by writing the full offset in the ADRP's immediate field, and repeating its low 12 bits in the ADD's immediate field. A linker resolving the R_AARCH64_ADD_ABS_LO12_NC relocation on the ADD will not compute the correct overall 64-bit value, but the error will only be in the higher bits, which are discarded by that relocation.
 
 Relocation types
 ^^^^^^^^^^^^^^^^


aaelf64/aaelf64.rst
```diff
@@ -940,6 +940,13 @@ A ``RELA`` format relocation must be used if the initial addend cannot be encode
 
 There is no PC bias to accommodate in the relocation of a place containing an instruction that formulates a PC- relative address. The program counter reflects the address of the currently executing instruction.
 
+There are two special cases for forming the initial addend of REL-type relocations where the immediate field cannot normally hold small signed integers:
+
+* For relocations processing MOVZ and MOVK instructions, the initial   addend is formed by interpreting the 16-bit literal field of the instruction as a 16-bit signed value in the range -32768 <= A < 32768. The interpretation is the same whether or not the instruction applies a left shift to its immediate: the addend is never treated as shifted.
```
smithp35, line 945: Would it be worth including MOV (wide immediate) using the Arm ARM description. For example 
```For relocations processing MOV (wide immediate), MOVZ and MOVK instructions,```

MOV (wide immediate) is an alias for MOVZ so this is included in MOVZ and MOVK, but MOV is used a lot in code examples.

**Review by smithp35 — COMMENTED** [review]
One small comment about the MOV alias but that isn't a strong opinion.

As I understand it the sequence
```
ADRP x0, symbol + addend
ADD x0, x0, :lo12: symbol +addend  
```
Would likely limit the size of the addend to ADRP to same addend [0-4095] that could be expressed in the ADD/LDR.

I would like to think there's a way of relating addend1 and addend2 so that we always get the right offset within the page.
```
ADRP x0, symbol + addend1
ADD x0, x0, :lo12: symbol +addend2
```
For example addend2 is congruent to addend1 (modulo 4096). Would need to prove that though.

If there is such an expression we could add that as a note.

statham-arm:
Nice thought! I think you're absolutely right: you can make ADRP+ADD consistently generate up to a 19-bit signed offset even though the ADD immediate can only hold 12 bits of it, by just setting the ADD immediate to the low 12 bits. The error in computing the full value during processing of the ADD relocation doesn't matter, because it's only in bits 12 and above, which don't affect the output.

aaelf64/aaelf64.rst
statham-arm: I dithered about that myself. Added a parenthesis.

statham-arm:
Oops! The ADRP immediate is 21 bits, not 19. Almost didn't notice that the encoding has two more immediate bits near the top of the word.

rearnsha:
Is it really safe to make this change?  The current text says:

5.7.2   Addends and PC-bias
<https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#id49>
<https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#572addends-and-pc-bias>

A binary file may use REL or RELA relocations or a mixture of the two (but
multiple relocations of the same place must use only one type).

The initial addend for a REL-type relocation is formed according to the
following rules.

   - If the relocation relocates data (Static Data relocations
   <https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#static-data-relocations>)
   the initial value in the place is sign-extended to 64 bits.
   - If the relocation relocates an instruction the immediate field of the
   instruction is extracted, scaled as required by the instruction field
   encoding, and sign-extended to 64 bits.

A RELA format relocation must be used if the initial addend cannot be
encoded in the place.

So the question is, could existing code be using these fields in a
different way (interpreting the bits as scaled by the field encoding
information)?  If they could, then this change would break existing usage
and thus cannot be safely made, even if it would have been more useful in
retrospect.

Wilco1:
> So the question is, could existing code be using these fields in a
different way (interpreting the bits as scaled by the field encoding
information)?  If they could, then this change would break existing usage
and thus cannot be safely made, even if it would have been more useful in
retrospect.

As explained, it is not possible to use a non-zero offset with REL MOVZ/MOVK/ADRP/ADD/LDR relocations since they wouldn't compute the same relocation.

smithp35:
> So the question is, could existing code be using these fields in a
different way (interpreting the bits as scaled by the field encoding
information)?  If they could, then this change would break existing usage
and thus cannot be safely made, even if it would have been more useful in
retrospect.

While we can never be 100% certain, we've only got one known user of SHT_REL for AArch64 which is the Arm legacy assembler armasm, and that switches to RELA for the instructions with scaled addends like ADRP, MOVT/MOVW. Part of the motivation for this change is to correct an incorrect implementation in LLD.

On the use case side, we can't be 100% certain, but I certainly can't think of a way a compiler could make use of the scaled immediates as relocation addends. I can remember from the 32-bit case that we first implemented the MOVT/MOVW in SHT_REL using scaled addends and soon ran into runtime failures which provoked the redefinition to not scale the addends. It may be possible to come up with something in assembly that isn't contrived.

I think this is an area where I'd be comfortable taking the risk.

smithp35 (approved):
Thanks for the updates, I'm approving this from my side. I suggest we leave this open for a few days for others to comment/object. Please can Request changes be used for any blocking feedback.

rearnsha:
I'm sure there are use cases where the bottom 16 bits might be known to be
zero, then you're more likely to want something that's a multiple of 2^16.

smithp35:
It would have to be an expression of the form
`symbol + <immediate>* 2 ^ 16`
Where the symbol is aligned on a` 2 ^ 16` boundary. The only thing that I can think of in that case is where symbol is the section symbol of a section aligned to at least `2 ^ 16`, and the section is larger than `2 ^ 16` bytes.

Given how rare that this use-case is compared to sections that are 4/8 byte aligned I doubt that a potential SHT_REL only toolchain would implement section symbol + offset like that. I'd expect it to use local anonymous symbols as anchors.

smithp35:
Will be merging this PR by end of UK working day 08/07/2024

smithp35:
End of UK working day. Merging this PR.

sallyarmneale (commented):
No changes required.

PR #273, opened 2024-07-17
[aapcs64] Describe the FPMR register and the FP8 types

momchil-velikov:
- Add a description of the FPMR register. The FPMR register
  is used to control operations with FP8 operands. It is expected
  that this register may need to be set more than once per
  function, hence it is considered a temporary register
  (clobbered by function calls).

- Add descriptions of the modal 8-bit floating point types: scalar,
   short-vector, and scalable vector. The respective parameter
   passing rules effectively state that values of these types are
   passed to and returned from functions in FP/vector registers, which
   reflects the typical and anticipated place of such values in instruction
   operands.


aapcs64/aapcs64.rst
```diff
@@ -892,6 +892,8 @@ thread-local storage on platforms where multi-threaded code is
 supported.  The exact location of such information is platform
 specific.
 
+The FPMR is system register that controls behaviors of the FP8 instructions. It is a temporary register. The state of the register must be held in thread-local storage on platforms where multi-threaded code is supported.
```
CarolineConcatto, line 895: I believe you need to manually cut the line for 80 characters.

momchil-velikov: I would very much prefer that was a  part of the "coding standard" for ABI.

aapcs64/aapcs64.rst
```diff
@@ -892,6 +892,8 @@ thread-local storage on platforms where multi-threaded code is
...
+
```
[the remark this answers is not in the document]
momchil-velikov: Done.

aapcs64/aapcs64.rst
```diff
@@ -892,6 +892,11 @@ thread-local storage on platforms where multi-threaded code is
 supported.  The exact location of such information is platform
 specific.
 
+The FPMR is system register that controls behaviors of the FP8 instructions. It
```
rsandifo-arm, line 895:
```suggestion
The FPMR is a system register that controls behaviors of the FP8 instructions. It
```

aapcs64/aapcs64.rst
```diff
@@ -892,6 +892,11 @@ thread-local storage on platforms where multi-threaded code is
...
+is a temporary register. The state of the register must be held in thread-local
+storage on platforms where multi-threaded code is supported. The exact location
+of that storage is platform specific.
```
rsandifo-arm, line 898:
```suggestion
is a temporary register.
```
I don't think we want the part about thread-local storage, for two reasons:

*   The idea of storing state in thread-local storage seems inconsistent with having a temporary/caller-saved register.  (This is contrast to something like FPCR, which is maintained cooperatively by the thread, rather than being caller-saved or callee-saved.)

*   The point of the corresponding decimal floating-point wording is that the DFP state is emulated rather than architected.  That is, DFP relies on something like the FPCR, but since there is no architected support for DFP, there is no architected support for a DFP-like FPCR either.  The DFP emulation code must therefore emulate both the DFP operations themselves and the thread-local nature of the control bits.

momchil-velikov: I'm open to rewording, perhaps not using the overloaded term "thread-local storage", but anyway IMO we would like to state exlicitly that the register contains thread-private data and must be saved/restored durign context switch, whuch  might be obviuos for temporary GPRs, but not for a system register.

aapcs64/aapcs64.rst
momchil-velikov: Done

aapcs64/aapcs64.rst
momchil-velikov: Done

aapcs64/aapcs64.rst
```diff
@@ -892,6 +892,9 @@ thread-local storage on platforms where multi-threaded code is
 supported.  The exact location of such information is platform
 specific.
 
+The FPMR is a system register that controls behaviors of the FP8 instructions.
```
mgabka, line 895: is it worth to tag this section as alpha for now ?

momchil-velikov: ok

aapcs64/aapcs64.rst
```diff
@@ -2570,6 +2575,9 @@ The mapping of C arithmetic types to Fundamental Data Types is shown in `Table 3
   |                                |                                         | significant bits of the type in a big-endian view.  Non-significant    |
   |                                |                                         | bits within the last quad-word are unspecified.                        |
   +--------------------------------+-----------------------------------------+------------------------------------------------------------------------+
+  |  **(Alpha)** ``__mfp8``        | unsigned byte                           | Arm extension. Values are intrepreted as either E5M2 or E4M3,          |
```
mro-github-12345, line 2578: typo
```suggestion
  |  **(Alpha)** ``__mfp8``        | unsigned byte                           | Arm extension. Values are interpreted as either E5M2 or E4M3,          |
```

aapcs64/aapcs64.rst
```diff
@@ -3054,35 +3066,37 @@ document.
...
+  | ``__SVInt64_t``                | ``svint64_t``         | signed double-word                        | VG             |
+  +--------------------------------+-----------------------+-------------------------------------------+----------------+
+  | ``__SVUint64_t``               | ``svuint64_t``        | unsigned double-word                      | VG             |
+  +--------------------------------+-----------------------+-------------------------------------------+----------------+
+  | ``__SVFloat64_t``              | ``svfloat64_t``       | double-precision float                    | VG             |
+  +--------------------------------+-----------------------+-------------------------------------------+----------------+
+  | ``__SVBool_t``                 | ``svbool_t``          | single bit (fully packed into VG bytes)   | VG×8           |
+  +--------------------------------+-----------------------+-------------------------------------------+----------------+
+  | **(Alpha)** ``__SVMfloat8_t``  | ``svmfloat8_t``       | modal 8-bit float                         | VG×8           |
```
CarolineConcatto, line 3098: Not a problem of this patch, but at some point we should add "svcount" if ABI is used to described the types

aapcs64/aapcs64.rst
```diff
@@ -2965,61 +2973,65 @@ The header file ``arm_neon.h`` also defines a number of intrinsic functions that
 
 .. table:: Table 7: Short vector extended types
 
-  +-----------------+-------------------+--------------------------+-----------+
```
CarolineConcatto, line 2968: Should we also add on
 Table 4, C/C++ type variants by data model
Just because I see for __fp16

momchil-velikov: I don't think so. `__mfp8` does not vary by data model.

CarolineConcatto (approved):

paulwalker-arm (approved):

momchil-velikov:
Rebased to resolve conflict and fix CI.

sallyarmneale (commented):
Looks good to me

Issue #196, opened 2023-03-27
[sysvabi64] document requirement for bti c in more detail


nsz-arm:
the text currently has

"An executable or shared library that supports BTI must have a bti c instruction at the start of any entry that might be called indirectly."

but it's not clear if compilers should consider potential linker inserted veneers with indirect call/jump or if the linker should ensure that when a veneer is inserted it does not break bti compatibility.

(gcc+ld.bfd made different choice than llvm+lld)

see discussion at
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106671

MaskRay:
Functions with LR signing gets PACI[AB]SP{,PC}. They have an implicit BTI.
If PACI[AB]SP is absent (leaf functions, or when PAuth is not enabled), Clang adds "bti c" to every candidate function because of the potential long calls
(LLD and GNU ld before https://sourceware.org/bugzilla/show_bug.cgi?id=30076) (https://reviews.llvm.org/D99417).

I assume that the LLD work is planned and Clang will eventually remove the "bti c" (BTW `-fbinutils-version=` exists if compatibility with older GNU ld versions is needed).
Is there more information about the double veneer scheme used by GNU ld. Do we need a new relocation to mark "bti c"?
(If there is concern with a new relocation type, NONE with a custom addend might be utilized.)

smithp35:
We've got an idea of where we want to go with this, I've been wanting to have an implementation in LLD ready before publishing and have not been able to find time to do this.

The change that needs making should make clear the requirements for code-generators and static linkers. The prevailing opinion within Arm is that we would like to enable code-generators to omit BTI if they can prove that the function will never be called indirectly (GCC behaviour). A static linker may therefore not assume that all indirect branch targets have a BTI compatible landing pad.

A "BTI compatible" thunk either doesn't use an indirect branch (chain of direct branches) or they are split up into two parts, the indirect branch, and a "header" that contains a BTI c, and ends with a direct branch. Something like:
```
caller:
  bl thunk_to_foo
  ...
thunk_to_foo:
  adrp x16, foo_bti_header
  add  x16, :lo12: foo_bti_header
  br   x16
  ...
foo_bti_header:
  bti  c
  b    foo
  ...
foo:
```
The "header" has a range limit (+-128Mib), and is essentially an alternative entry point for indirect calls. The presence of this alternative entry point undoes the compiler's hard work in omitting the BTI, but it will only be done if necessary.

[...]

PR #282, opened 2024-09-17
[sysvabi64] Document requirements for tools wrt BTI

smithp35:
Add requirements for when a tool must generate a BTI instruction. This permits tools to elide BTI instructions when they can prove that no indirect branch to that location is possible from local information available to the tool.

Static linkers are not allowed to assume that all direct branch targets have a BTI instruction. If a veneer is required then the static linker must generate additional BTI instructions if needed.

A static linker is allowed to assume that a symbol that is exported to the dynamic symbol table has a BTI instruction.

In practice this will permit compilers to remove BTI instructions from static functions that do not have their address taken and that address escapes the function.

This matches the behavior of the GNU toolchain.

Fixes https://github.com/ARM-software/abi-aa/issues/196

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1690,31 +1690,33 @@ It is desirable to minimize the number of BTI instructions to limit
 the number of indirect branch destinations in the program. The
 following tool requirements determine which tool has the
 responsibility of inserting the BTI instruction, permitting a tool to
-elide the BTI instuction when it can prove that there are no indirect
+elide the BTI instruction when it can prove that there are no indirect
 calls to that location.
 
-A relocatable object producer is required to add a BTI instruction to
-the destination of an indirect branch originating in the same
-relocatable object.
+* A relocatable object producer is required to add a BTI instruction
+  to the destination of an indirect branch originating in the same
+  relocatable object.
 
-A relocatable object producer is required to add a BTI instruction to
-a location when the address of that location is live and escapes out
-of the relocatable object. This includes the locations of all symbols
-that can be exported into the dynamic symbol table by a static linker.
+* A relocatable object producer is required to add a BTI instruction
+ to a location when the address of the location is live, and it
+ escapes to an entity that is permitted to generate an indirect branch
+ that is opaque to the relocatable object producer. This includes the
+ locations of all symbols that can be exported into the dynamic symbol
+ table by a static linker.
 
-A static linker is required to generate `Custom PLTs`_ with BTI
-instructions.
+* A static linker is required to generate `Custom PLTs`_ with BTI
+  instructions.
 
-A static linker that uses indirect branches in veneers is required to
-generate a BTI compatible landing pad if the target of the indirect
-branch is defined within the same link unit and does not have a
-compatible BTI instruction at the destination of the veneer. A BTI
-compatible landing pad consists of a BTI instruction followed by a
-direct branch. For example:
+* A static linker that uses indirect branches in veneers is required
+  to generate a BTI compatible landing pad if the target of the
+  indirect branch is defined within the same link unit and does not
+  have a compatible BTI instruction at the destination of the
+  veneer. A BTI compatible landing pad consists of a BTI instruction
+  followed by a direct branch. For example:
 
 .. code-block:: asm
 
-  // Linker generated veneer using indirect bracnh
+  // Linker generated veneer using indirect branch
   adrp x16, fn
   add  x16, :lo12: fn
   br   x16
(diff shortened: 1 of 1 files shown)


smithp35:
LLD patch to implement BTI aware thunks https://github.com/llvm/llvm-project/pull/108989

sysvabi64/sysvabi64.rst
```diff
@@ -1671,6 +1670,65 @@ include:
...
+elide the BTI instuction when it can prove that there are no indirect
+calls to that location.
+
+A relocatable object producer is required to add a BTI instruction to
+the destination of an indirect branch originating in the same
+relocatable object.
+
+A relocatable object producer is required to add a BTI instruction to
+a location when the address of that location escapes out of the
+relocatable object. This includes the locations of all symbols that
```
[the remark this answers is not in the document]
smithp35: I'm thinking about how to write this without tying myself in knots. My current favourite is
"the address of that location is live and escapes out of the relocatable object." That should cover the case where the address of the is in a register or memory but there's no reader.

smithp35:
Apologies for the force push, needed to rebase to pick up the new github actions.

sysvabi64/sysvabi64.rst
```diff
@@ -1671,6 +1670,65 @@ include:
...
+* Static linkers when generating PLT sequences or veneers.
+
+* Other executables and shared libraries via call from a PLT or a
+  function pointer.
+
+It is desirable to minimize the number of BTI instructions to limit
+the number of indirect branch destinations in the program. The
+following tool requirements determine which tool has the
+responsibility of inserting the BTI instruction, permitting a tool to
+elide the BTI instuction when it can prove that there are no indirect
```
vhscampos, line 1693: Typo "instuction"

sysvabi64/sysvabi64.rst
```diff
@@ -1671,6 +1670,65 @@ include:
...
+calls to that location.
+
+A relocatable object producer is required to add a BTI instruction to
```
vhscampos, line 1696: Consider putting each requirement in a list item.

sysvabi64/sysvabi64.rst
```diff
@@ -1671,6 +1670,65 @@ include:
...
+A static linker that uses indirect branches in veneers is required to
+generate a BTI compatible landing pad if the target of the indirect
+branch is defined within the same link unit and does not have a
+compatible BTI instruction at the destination of the veneer. A BTI
+compatible landing pad consists of a BTI instruction followed by a
+direct branch. For example:
+
+.. code-block:: asm
+
+  // Linker generated veneer using indirect bracnh
```
vhscampos, line 1717: Typo: "bracnh"

sysvabi64/sysvabi64.rst
```diff
@@ -1671,6 +1670,65 @@ include:
...
+a location when the address of that location is live and escapes out
+of the relocatable object. This includes the locations of all symbols
+that can be exported into the dynamic symbol table by a static linker.
+
+A static linker is required to generate `Custom PLTs`_ with BTI
+instructions.
+
+A static linker that uses indirect branches in veneers is required to
+generate a BTI compatible landing pad if the target of the indirect
+branch is defined within the same link unit and does not have a
```
vhscampos, line 1710: Can you please clarify why "if ... defined within the same link unit" is necessary? What happens in the case where the target is defined in another link unit?

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: I've gone with a more abstract escapes to an opaque entity. The original text refers to the boundary between the compiler/assembler and linker, but it can be more general than that. For example inline assembly that the compiler can't see into could indirect branch if the escaped address were on the input list.

sysvabi64/sysvabi64.rst
smithp35: ACK

smithp35 (commented):
Thanks for the comments. I'll upload a new patch.

vhscampos:
Ok by me. Thanks. But please wait for another approval.

statham-arm (approved):
I'm also happy with this now.

sallyarmneale (commented):
Looks good to me

PR #285, opened 2024-09-24
Upgrade actions in ci.yml

tzussman:
CI runs are failing because actions/upload-artifact@v2 has been deprecated: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Upgrade upload-artifact to v4 and upgrade the rest of the actions while at it too in order to avoid deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,9 +7,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


stuij:
Thanks so much for this pull request! And sorry, I missed it somehow. I pushed a similar patch already.

PR #286, opened 2024-09-25
Actions version bump

stuij:
our GitHub action checks were failing because of update-artefact got depreciated.

While updating, I took the liberty to also update the other script versions:
- update checkout to v4, v2 is using a node that will soon be depreciated
- upload-artifact to v4, v2 has become depreciated
- update setup-python to v5, just to stay up-to-date


smithp35 (commented):
LGTM. Thanks for fixing.

PR #290, opened 2024-10-11
Include SME attributes in the name mangling of types

kmclaughlin-arm:
This change extends the name mangling of types to include the SME streaming
and ZA interface. This will avoid naming conflicts which can currently arise such
as in the following example:

```
  void foo(void (*f)()) { f(); }
  void foo(void (*f)() __arm_streaming) { f(); }
```

Without this change, both functions 'foo' above will mangle to the same
name, despite the function pointers being different.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -256,6 +256,7 @@ changes to the content of the document for that release.
 |            | September 2024     | - Add soft-float PCS variant.                                    |
 |            |                    | - Add the __arm_get_current_vg SME support routine.              |
 |            |                    | - Clarify use of `it` when preserving z and p registers.         |
+|            |                    | - Update C++ mangling to include SME attributes in type names    |
 +------------+--------------------+------------------------------------------------------------------+
 
 References
@@ -3110,6 +3111,34 @@ instead.
 
 The SVE tuple types are mangled using their ``arm_sve.h`` names
 (``svBASExN_t``).
+
+Types which have an SME streaming or ZA interface should include an additional suffix as described in the table below:
+
++-------------------------+-----------------------------+
+| Type of interface       | Suffix                      |
++=========================+=============================+
+| Non-streaming (default) | None                        |
++-------------------------+-----------------------------+
+| Streaming               | sm                          |
++-------------------------+-----------------------------+
+| Streaming-compatible    | sc                          |
++-------------------------+-----------------------------+
+| Private-ZA (default)    | None                        |
++-------------------------+-----------------------------+
+| Shared-ZA               | sz                          |
++-------------------------+-----------------------------+
+
+A streaming interface suffix should precede any ZA interface suffix. For example:
+
+.. code:: c
+
+void f(svint8_t (*fn)() __arm_inout("za") __arm_streaming) { fn(); }
+
+is mangled as
+
+.. code:: c
+
+_Z1fPFu10__SVInt8_tsmszvE
 
 .. raw:: pdf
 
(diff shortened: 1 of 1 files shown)


kmclaughlin-arm:
@rsandifo-arm

aapcs64/aapcs64.rst
```diff
@@ -3111,6 +3112,34 @@ instead.
 The SVE tuple types are mangled using their ``arm_sve.h`` names
 (``svBASExN_t``).
 
+Types which have an SME streaming or ZA interface should include an additional suffix as described in the table below:
```
rsandifo-arm, line 3115: Adding suffixes like this would conflict with the existing mangling scheme for types.  E.g. `sm` would normally indicate an extra `short` argument followed by an extra `unsigned long` argument ([spec](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling)):

```
$ c++filt
_Z1fPFu10__SVInt8_tsmE
f(__SVInt8_t (*)(short, unsigned long))
```

I wondered whether we could use a vendor-specific qualifier (`U…`), but function argument types are understandably mangled without a qualifier.  (For example, there is no difference between `f(const int)` and `f(int)`.)

For `__attribute__((arm_sve_vector_bits(N)))`, we used a fake template type `__SVE_VLS<…, N>` to attach the vector length N to the underlying standard SVE vector type.  Perhaps we could do the same sort of thing here: have a fake template type that wraps the unadorned function type and adds information about the streaming mode and shared state?  This would probably mean encoding that information as C++ expressions.

rsandifo-arm: Sorry, I hopelessly confused myself while writing and rewriting the comment above, so gave the wrong reason.  The problem isn't that function argument types have no qualifiers, but that function types themselves have only cv-qualifiers.  The spec hard-codes markup for transactional safety (`Tx`), but AFAIK there's no general mechanism for adding vendor extensions at the function type level.

The conclusion is the same though :-)

kmclaughlin-arm:
@sdesmalen-arm
@efriedma-quic
@aemerson

aapcs64/aapcs64.rst
```diff
@@ -3111,6 +3112,65 @@ instead.
 The SVE tuple types are mangled using their ``arm_sve.h`` names
 (``svBASExN_t``).
 
+Function types which have SME streaming, ZA interface or ZT0 interface attributes must include these attributes in the name mangling of the type.
```
rsandifo-arm, line 3115: Like @efriedma-quic says, the intention was to include this mangling if and only if the mangling would normally include the return type of the function.  We should probably say that explicitly.

aapcs64/aapcs64.rst
```diff
@@ -3111,6 +3112,65 @@ instead.
...
+  +------------------------+--------------------------+
+
+* za_state is an integer representing the ZA state of the function:
+
+  +------------------------+--------------------------+
+  | Interface Type         | Value                    |
+  +========================+==========================+
+  | No ZA State (default)  | 0                        |
+  +------------------------+--------------------------+
+  | Shared ZA              | 1                        |
```
rsandifo-arm, line 3148: The reason for mangling the ZA information is to deal with the fact that the sharing attributes create distinct function types for overloading purposes.  I think this means that we need to represent all five possibilities:

* normal
* `__arm_in("za")`
* `__arm_inout("za")`
* `__arm_out("za")`
* `__arm_preserves("za")`

Similarly for `zt0`.

Since that classification is only described in the ACLE, we should probably describe the mangling there too.  Sorry for only realising that now.

kmclaughlin-arm: Thanks @rsandifo-arm, I've updated this to represent each of the shared ZA/ZT0 attributes separately and moved these changes to the ACLE (https://github.com/ARM-software/acle/pull/358).

Issue #293, opened 2024-10-25
[pauthabielf64] Define static AUTH TLSDESC relocations


kovdan01:
In pauthabielf64, dynamic AUTH TLSDESC relocation `R_AARCH64_AUTH_TLSDESC` is defined https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#auth-variant-got-generating-relocations.

However, the section with AUTH variant GOT-generating relocations does not contain definitions for static TLSDESC relocs: https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#auth-variant-got-generating-relocations.

Particularly, the following relocation types are missing and should be defined:

1. `R_AARCH64_AUTH_TLSDESC_ADR_PAGE21`
2. `R_AARCH64_AUTH_TLSDESC_LD64_LO12`
3. `R_AARCH64_AUTH_TLSDESC_ADD_LO12`

kovdan01:
Tagging @smithp35 @asl

kovdan01:
@smithp35 I've submitted a PR to llvm-project repo adding the relocs with the following codes (the latest code for previously defined GOT-generating relocs was 0x811d):

```
R_AARCH64_AUTH_TLSDESC_ADR_PAGE21 0x811e
R_AARCH64_AUTH_TLSDESC_LD64_LO12  0x811f
R_AARCH64_AUTH_TLSDESC_ADD_LO12   0x8120
```

Please let me know if these code are OK.

smithp35:
My apologies, I'll be out of office for the the week of 28th October to the 1st November so this is a quick response at the weekend. I'll get back to this when I'm back in the office in a week's time.

The relocation codes are OK for now [1] as they are in the private space reserved for experiments https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5714private-and-platform-specific-relocations 

The GOT signing relocations are currently in the private space. As I understand it there's a high chance that these will be needed. When they stabilise we'll need to move them to the space reserved for PAuthABI https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5716pauthabi-relocations 

What I think needs doing:
* There are more non-auth TLSDESC relocations defined in AAELF64. Is it worth defining AUTH equivalents for these (assuming they are relevant to PAuth; I've not checked)?
* Make sure we're happy with the names. From a glance they look fine.
[...]

PR #295, opened 2024-11-05
[pauthabi64] Add Auth variants of TLSDESC generating GOT relocations

smithp35:
Add static Auth equivalents of the existing relocations:
* R_AARCH64_TLSDESC_ADR_PAGE21
* R_AARCH64_TLSDESC_LD64_LO12
* R_AARCH64_TLSDESC_ADD_LO12

These are sufficient to implement the small code-model for TLSDESC using the PAuthABI. Support for the tiny and large code-model is not implemented for the PAuthABI, or for the normal ABI in both clang and GCC so we defer from specifiying Auth variants of all possible relocations.

I've followed the convention that non-zero addends are not permitted for GOT generating relocations. As each unique S+A combination needs its own pair of GOT slots indexed by S+A and no static linker supports that.

Fixes https://github.com/ARM-software/abi-aa/issues/293

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -243,6 +243,9 @@ changes to the content of the document for that release.
   |            |                             | - Add R_AARCH64_AUTH_GOT_ADR_PREL_LO21 relocation.               |
   |            |                             | - Remove addend in GDAT relocation operation.                    |
   +------------+-----------------------------+------------------------------------------------------------------+
+  | 2024Q1     | 5\ :sup:`th` November 2024  | - Add static Auth TLSDESC generating relocations                 |
+  |            |                             | - Correct comment entry for R_AARCH64_AUTH_LD64_GOTOFF_LO15      |
+  +------------+-----------------------------+------------------------------------------------------------------+
 
 References
 ----------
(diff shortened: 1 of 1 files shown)


pauthabielf64/pauthabielf64.rst
```diff
@@ -1103,80 +1106,94 @@ AUTH variant GOT Generating Relocations
...
+  +-------------+----------------------------------------+----------------------------------------+--------------------------+
+  | 0x811D      | R\_AARCH64\_AUTH\_GOT\_ADR\_PREL\_LO21 | G(ENCD(GDAT(S))) - P                   | Set the immediate        |
+  |             |                                        |                                        | value to bits[20:0] of X;|
+  |             |                                        |                                        | check that -2 :sup:`20`  |
+  |             |                                        |                                        | <= 2 :sup: `20`          |
+  +-------------+----------------------------------------+----------------------------------------+--------------------------+
+  | 0x811E      | R\_AARCH64\_AUTH\_TLSDESC\_ADR\_PAGE21 | Page(G(ENCD(GTLSDESC(S)))) - Page(P)   | Set an ADRP immediate    |
+  |             |                                        |                                        | field to bits [32:12] of |
+  |             |                                        |                                        | X; check –2\ :sup:`20`   |
```
stuij, line 1186: The range for ADRP is ±4GB [1], so I think that the check should be between -2^32 and 2^32, instead of 20.

[1] https://developer.arm.com/documentation/100076/0100/A64-Instruction-Set-Reference/A64-General-Instructions/ADRP

smithp35: ACK. It looks like I used the text from `R_<CLS>_TLSDESC_ ADR_PREL21` in the table rather than the similarly named R_<CLS>_TLSDESC_ ADR_PAGE21

Will fix.

smithp35: I've updated the PR to use -2^32 and 2^32

stuij (approved):
LGTM

sallyarmneale (commented):
Looks good to me.

PR #303, opened 2024-12-13
Markup: fix two accidental triple-backticks.

statham-arm:
I noticed one of these in passing in RTABI32: Github's Markdown renderer showed a spurious backtick at the end of a sensible identifier in fixed-pitch text. Looking at the source, that's because the fixed-pitch section was closed by three backticks instead of the usual two.

Grepping the rest of the repository turned up one more case of the same thing, in DBGOVL32. This patch fixes both.


smithp35 (approved):
LGTM. Let me know if you don't have enough permissions to merge and I can do that.

statham-arm:
Apparently I don't – I can't see the usual merge button anywhere. So yes please!

PR #311, opened 2025-02-20
[sysvabi64] Add chapter on Thread Local Storage

smithp35:
The thread local storage chapter contains:
* A description of Thread Local Storage based on addenda32
* The key design decisions of AArch64 TLS such as tls variant, tls dialect, TCB size.
* The ABI required code sequence for TLSDESC that must be emitted exactly, as GNU ld requires it to be.
* Sequences for the different code-models.
* Relaxations for GD->IE, GD->LE and IE->LE.
* Synchronization requirements for Lazy TLSDESC. With advice not to support it due to overhead of synchronization.


sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1854,601 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+  KiB, 16 MiB, 4GiB or 16EiB, depending on the addressing mode. The
+  default is 16 MiB for all addressing modes.
+
+The static and dynamic linker must agree on the size of the padding
+between the TCB and the executables TLS Block. Using ``TCB`` as the
+size of the TCB (16 bytes), ``PAD`` as the size of the padding bytes,
+and ``PT_TLS`` as the program header with type PT_TLS. ``PAD`` must be
+the smallest positive integer that satisfies the following congruence:
+
+``TP + TCB + PAD ≡ PT_TLS.p_vaddr (modulo PT_TLS.p_align)``
```
MaskRay, line 2004: `TP+TCB+PAD` on the left could be confusing, as TCB is placed before TP. Perhaps mention the requirement of TP first (= 0 (modulo p_align)), then describe `PAD` and this formula.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1854,601 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+Given that ``TP ≡ 0 (modulo PT_TLS.p_align)``. An expression
+for `PAD` is ``PAD = (PT_TLS.p_vaddr - TCB) mod PT_TLS.p_align``.
+
+A significant number of dynamic linkers use a different calculation
```
MaskRay, line 2009: I believe that glibc Variant I handles `p_vaddr!=0 (mod p_align)` correctly. The bug (https://sourceware.org/bugzilla/show_bug.cgi?id=24606
) is for Variant II (x86 etc).

I have fixed FreeBSD rtld's Variant II in https://reviews.freebsd.org/D31538 . Its Variant I may or may not have the bug.

musl has been good since 1.1.23

Therefore, it's probably not "a significant number" but yeah `p_vaddr=0 (mod p_align)` is good for maximum compatibility

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1854,601 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+    // offset of var from tp in xn
+
+Optimization to load a 64-bit var directly into a core register.
+
+.. code-block:: asm
+
+    add  xn, tp, :tprel_hi12:var, lsl #12 // R_AARCH64_TLSLE_ADD_TPREL_HI12 var
+    ldr  xn, [xn, #:tprel_lo12_nc:var]    // R_AARCH64_TLSLE_LDST64_TPREL_LO12_NC var
+
+Static link time TLS Relaxations
```
MaskRay, line 2191: Perhaps call this Optimization to be consistent with x86/ppc and "Relocation optimization" (ADRP) and leave the term "relocation relaxation"  for RISC-V style section shrinking.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1854,601 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+thread pointer by subtracting the address of the thread pointer from
+the address of the TLS variable. In practice an implementation of the
+dynamic TLS resolver contains many platform specific details outside
+of the scope of the ABI. An example of how a dynamic resolver might be
+implemented can be found in the Dynamic Specialization section of
+TLSDESC_.
+
+Undefined Weak Symbols
+
+An undefined weak symbol has the value 0. As the resolver function
```
MaskRay, line 2379: This might be the glibc behavior, but musl doesn't have the special `__dl_tlsdesc_undefweak`. I think it's better to allow flexibility and require a particular behavior on undefined weak TLS.

smithp35: This part is just an example of what can be done. I've written at the top of the section
> The TLS resolver functions are not standardized by this ABI as they are internal to the dynamic linker

and

> These examples are for illustrative purposes only

I'll see if there's anything I can do to state that there is no requirement to implement a specific resolver function.

sysvabi64/sysvabi64.rst
smithp35: I'll see if I can word it better. I've found it difficult to try and explain the formula intuitively.

smithp35 (commented):
Thanks very much for the review.

I've updated based on this and some comments I received internally.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+*  Thread local storage (variables instanced once per thread, shared between
+   all accessing function activations).
+
+*  (Most local) Automatic data (stack variables, instanced once per function
+   activation, per thread).
+
+Rules governing thread local storage on AArch64
+-----------------------------------------------
+
+*  How to denote TLS in source programs.
```
stuij: For these `How..` bullet points,  I'd replace the period at the end with a `:`, to prime readers for the format.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+   SCO-ELF_).
+
+*  How a loader or run-time system creates instances of TLS per-thread at
+   execution time.
+
+   This is part of ABI for the platform or execution environment.
+
+This document and AAELF64_ are concerned with:
+
+*  How to relocate, statically and dynamically, with respect to symbols
```
stuij, line 2118: In `How to relocate..` I'm expecting the subject at some point. Especially with the subclause and the parentheses it becomes a bit of a confusing sentence that kinda peters out.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+In the most general form, a program is constructed dynamically from an
+executable and a number of shared libraries. Each component,
+(executable or shared library) can be mapped into multiple
+processes. Additionally a shared library can be loaded dynamically by
+a program, rather than being part of the initial process image
+constructed when the program is first loaded.
+
+For the purpose of addressing TLS, components, referred to as modules,
+of an application are identified using indexes. The module index for
```
stuij: `components, referred to as modules, of an application` -> `components of an application, referred to as modules,`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+to the TLS block for module N.
+
+Index 0 of the Dynamic Thread Vector DTV[0] is reserved for use by the
+platform. It typically contains the thread's generation counter which
+can be used to update or reallocate the DTV when TLS variables in
+dynamic modules loaded by ``dlopen`` are first used. When a dynamic
+TLS variable is accessed the thread's generation count is compared
+with the global generation count which can be used to trigger updates
+of the DTV. The details are platform specific.
+
```
stuij: At this point it might be useful to give a small high-level example along the lines of `for example when dynamic library x gets loaded, we add TLS block, add DTV entry..`, just to drive off confusion.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+   4. Local Exec, can be used in the executable for TLS variables
+      defined in the executables static TLS block.
+
+SystemV AArch64 TLS addressing
+------------------------------
+
+AArch64 TLS SystemV design choices
+
+* AArch64 uses variant 1 TLS as described in ELFTLS_.
```
stuij, line 2228: Perhaps mention ELFTLS when doing the TLS introduction as a `for more in-depth info` resource.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+* The thread pointer (TP) is always accessible via the ``TPIDR_EL0``
+  system register. This can be accessed via inlining a ``mrs``
```
stuij: `a ``mrs`` instruction` -> `an ``mrs`` instruction`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+The static and dynamic linker must agree on the size of the padding
+between the TCB and the executables TLS Block. Using ``TCBsize`` as the
+size of the TCB (16 bytes), ``PADsize`` as the size of the padding bytes,
+and ``PT_TLS`` as the program header with type PT_TLS.
+
+The Thread Pointer ``TP`` and also the address of the start of the
+``TCB``, must satisfy the requirement.
+
+``TP ≡ 0 (modulo PT_TLS.p_align)``.
```
stuij: I had to search a bit to discover that this is mathematical jargon. I asked a few people who I suspected would be more mathematical inclined than me, and they also came up empty.

If this is also linker jargon, than fine. Otherwise perhaps rephrase to make the precise meaning a bit more accessible?

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+General Dynamic is the most general form of accessing TLS. It supports
+static and dynamic TLS.
+
+To permit static linker relaxation. The TLSDESC code sequences must be
+emitted exactly as specified, with no other instruction breaking up
+the sequence, with exactly the same registers used.
+
+The code sequences below return the offset of the TLS variable from
+``tp`` in ``x0``.  To get the address of the TLS variable requires
+additional code to add ``x0`` to be added to ``tp``, this is not part
```
stuij: I presume `to add x0 to be added to tp` 
should be `to add x0 to tp`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+    // offset of var from tp in x0
+
+Local Dynamic
+^^^^^^^^^^^^^
+
+Local Dynamic is a special case of general dynamic where the compiler
+knows that the TLS variable is defined in the same module as the code
+that is accessing the variable. In this case the offset of the TLS
+variable from the start of the module's TLS block is a static link
+time constant. Instead of dynamically calculating the offset of the
```
stuij: `constant. Instead` -> `constant, instead`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+Recall that the Thread Pointer ``TP`` points to the start of the ``TCB``.
+
+The static and dynamic linker must agree on the size of the padding
+between the TCB and the executables TLS Block. Using ``TCBsize`` as the
+size of the TCB (16 bytes), ``PADsize`` as the size of the padding bytes,
+and ``PT_TLS`` as the program header with type PT_TLS.
+
+The Thread Pointer ``TP`` and also the address of the start of the
+``TCB``, must satisfy the requirement.
```
stuij: `must satisfy the requirement.` -> `must satisfy the requirement:`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+only dialect supported by clang.
+
+Code sequences for accessing TLS variables
+------------------------------------------
+
+The code sequences below assume the default TLS block size of 16 MiB,
+this permits the Local Exec model to use of a pair of add instructions
+with a combined 24-bit immediate field. Larger TLS sizes can be
+supported by using a ``movz`` and one or more ``movk`` instructions to
+construct an offset from the thread-pointer in a register.
```
stuij: `thread-pointer` -> `thread pointer` (x 6)

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+   3. Initial Exec, can be used for TLS variables defined in the
+      static TLS block.
+
+   4. Local Exec, can be used in the executable for TLS variables
+      defined in the executables static TLS block.
+
+SystemV AArch64 TLS addressing
+------------------------------
+
+AArch64 TLS SystemV design choices
```
stuij, line 2225: `choices` -> `choices:`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+To permit static linker relaxation. The TLSDESC code sequences must be
+emitted exactly as specified, with no other instruction breaking up
+the sequence, with exactly the same registers used.
+
+The code sequences below return the offset of the TLS variable from
+``tp`` in ``x0``.  To get the address of the TLS variable requires
+additional code to add ``x0`` to be added to ``tp``, this is not part
+of the ABI required TLSDESC code sequence.
+
+Small Code Model
```
stuij: In this section and the sections below I'd put semicolons after all the model headers: `Small Code Model:`, or use some kind of header format.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+Local Exec
+^^^^^^^^^^
+
+Local Exec is used for accesses to the executable's TLS block. The
+executable always has the TLS module index of 1 so the offsets of the
+TLS variables from the thread pointer are static link time
+constants. The code sequences are the same for all code models.
+
+The instruction sequences below are not ABI. Using the instructions
```
stuij: `are not ABI` doesn't sound wholly English?

Perhaps: `are not ABI specifications, but using..`?

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+.. code-block:: asm
+
+    add  xn, tp, :tprel_hi12:var, lsl #12 // R_AARCH64_TLSLE_ADD_TPREL_HI12 var
+    ldr  xn, [xn, #:tprel_lo12_nc:var]    // R_AARCH64_TLSLE_LDST64_TPREL_LO12_NC var
+
+Static link time TLS Relaxations
+--------------------------------
+
+Relaxation is a term used by the TLS literature such as ELFTLS_ to
```
stuij: Maybe move this general info up a bit. By now we have used the term twice in this patch.

Say the first 2 paragraphs until `using the constrained model.`

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+Example Resolver Functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+These examples are for illustrative purposes only. There is no
+requirement for any of the following resolver functions to be
+implemented.  Due to the restrictions on calling convention, the
+resolver routines must be written in assembly language.
+
+Static TLS Specialization
```
stuij: For these as well I would add a colon.

sysvabi64/sysvabi64.rst
```diff
@@ -1844,6 +1857,635 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+static linker is permitted to relax each instruction individually,
+using the relocation directive to identify the instruction. The
+destination register must be preserved.
+
+.. code-block:: asm
+
+    movz    xn, :tprel_g1:var // R_AARCH64_TLSLE_MOVW_TPREL_G1 var
+    movk    xn, :tprel_g0:var // R_AARCH64_TLSLE_MOVW_TPREL_G0_NC var
+
+TLS Descriptors
```
stuij: I'd move these two more general sections above `Code sequences for accessing TLS variables`.

sysvabi64/sysvabi64.rst
smithp35: ACK.

sysvabi64/sysvabi64.rst
smithp35: I took the format from the 32-bit ABI https://github.com/ARM-software/abi-aa/blob/main/addenda32/addenda32.rst#41introduction-to-thread-local-storage which has a similar list of bullet points followed by `It is the last two bullet points that are the subject of this ABI.`

What I'm trying to say is that the ABI only covers a small part of what is needed to support TLS. I may have lost that part. I'll try and reword.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: I'll see what I can do. This part of TLS is, strictly speaking, outside the scope of the ABI, it is implementation details of a dynamic loader.

I think there's three parts:
* dlopen, dlclose increment/decrement the global generation number independently of whether the module uses TLS.
* tls_get_addr() compares the thread's generation counter against the global generation number and will resize/reallocate the dynamic thread vector if it is different.
* Initially the dynamic thread vector entry is unallocated. On first use tls_get_addr will allocate and initialize the TLS.

sysvabi64/sysvabi64.rst
smithp35: ACK. I'll mention that the introduction in the ABI is only sufficient to describe the terms used like Thread Control Block. A general introduction can be found in ELFTLS_

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: I originally had this written out in words ``TP congruent-to 0 (modulo PT_TLS.p_align)`` . This is more easily searched for, but it does get somewhat verbose to write multiple times.

I think I'll start with the expression for pad-size, rather than the derivation. I can also introduce the simpler max(TCBsize, PT_TLS.p_align) as well which can be understood intuitively, it is just a pity it also requires PT_TLS.p_vaddr to be 0 modulo PT_TLS.p_align.

The derivation can come later for the people that want to understand where it came from.

sysvabi64/sysvabi64.rst
smithp35: Yes, thanks for spotting.

sysvabi64/sysvabi64.rst
smithp35: I'll use a semicolon. I think I'm already at the lowest level of header.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: I've used are "not required by the ABI, but using".

sysvabi64/sysvabi64.rst
smithp35: I'll take a look at this tomorrow.

sysvabi64/sysvabi64.rst
smithp35: ACK

smithp35 (commented):
Thanks very much for the comments. I'll hopefully have a new patch tomorrow.

sysvabi64/sysvabi64.rst
smithp35: ACK.

sysvabi64/sysvabi64.rst
smithp35: ACK

smithp35 (commented):
Thanks for the review comments. I've made the following updates:
* Simple NFC text changes.
* Better description of deferred TLS and generation count.
* Reworded the padding size derivation.
* Moved paragraphs around to make the flow a bit easier.

Should be visible as 4 separate commits

sysvabi64/sysvabi64.rst
```diff
@@ -22,7 +23,15 @@
 .. _SCO-ELF: http://www.sco.com/developers/gabi
 .. _SYM-VER: http://www.akkadia.org/drepper/symbol-versioning
 .. _SYSVABI: https://github.com/ARM-software/abi-aa/releases
-.. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
+.. _ELFTLS: https://www.uclibc.org/docs/tls.pdf
+.. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-ARM.txt
+.. _TLSDESCRES: https://github.com/ARM-software/abi-aa/tree/main/design-documents/tlsdesc-resolvers.txt
```
yury-khrustalev, line 28: nit: file extension should probably be `.rst`

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+*  (Most global) Program-own data (static and extern variables, instanced
+   once per program/process).
+
+*  Thread local storage (variables instanced once per thread, shared between
+   all accessing function activations).
+
+*  (Most local) Automatic data (stack variables, instanced once per function
+   activation, per thread).
+
+Rules governing thread local storage on AArch64
```
yury-khrustalev: This section probably should be named "Scope" and the part about `thread_local` and `__thread` should probably elsewhere.

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+	   the global_generation_number */
+        update_thread_dtv();
+
+      if (dtv[module_id] == unallocated)
+        allocate_tls(dtv, module_id);
+
+      return dtv[module_id][offset];
+    }
+
+The calculation in __tls_get_addr is the most general and it can be
```
yury-khrustalev: nit:
```
s/__tls_get_addr/``__tls_get_addr``
```

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+platform. It is typically used to store the thread's generation
+counter. In an implementation that supports deferred allocation of
+TLS, a global generation number is incremented whenever the number of
+dynamic modules changes due to ``dlopen`` or ``dlclose``. In the
+``__tls_get_addr(module_id, offset)`` function, if the thread's
+generation count is less than the global generation number, the
+thread's DTV is updated, and the TLS for the ``module_id`` is
+allocated if it is not present.
+
+In pseudo code
```
yury-khrustalev: I think this pseudo code listing is not necessary since we already have the description of the operation above. If both are retained, the code and the description may diverge over time.

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+        update_thread_dtv();
+
+      if (dtv[module_id] == unallocated)
+        allocate_tls(dtv, module_id);
+
+      return dtv[module_id][offset];
+    }
+
+The calculation in __tls_get_addr is the most general and it can be
+applied to both static and dynamic TLS. There are four defined models
```
yury-khrustalev: > There are four defined models

This probably should start new paragraph and probably a new section

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+of accessing TLS that trade off generality for performance. In order
+of descending generality:
```
yury-khrustalev: > descending generality

nit: maybe "descending" is better?

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+   2. Local Dynamic, can be used anywhere where the definition of the
+      TLS variable and the access are from the same module.
+
+   3. Initial Exec, can be used for TLS variables defined in the
+      static TLS block.
+
+   4. Local Exec, can be used in the executable for TLS variables
+      defined in the executables static TLS block.
+
+SystemV AArch64 TLS addressing
```
yury-khrustalev: > SystemV AArch64 TLS addressing

The title of this section is very similar to the previous one. I think it makes it a bit difficult to navigate this chapter.

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+* AArch64 uses variant 1 TLS as described in ELFTLS_.
+
+* There are two dialects of TLS supported by the relocations defined
+  in AAELF64_, the traditional dialect described by ELFTLS_ and the
+  descriptor dialect described by TLSDESC_. This document describes
+  only the descriptor dialect as this is the default dialect for GCC
+  and the only dialect supported by clang.
+
+* The thread pointer (TP) is always accessible via the ``TPIDR_EL0``
```
yury-khrustalev: > The thread pointer (TP)

nit:

```
s/TP/``TP``
```

as `TP` is used like inline code below.

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+The ``TP``, and hence the start of the ``TCB`` must be aligned to a
+``PT_TLS.p_align`` boundary. This can be expressed as ``TP ≡ 0 (modulo
+PT_TLS.p_align)`` where ``≡`` means congruent to.
+
+The static and dynamic linker must agree on the size of the padding
+(``PADsize``) between the TCB and the executable's TLS Block. Using
+``TCBsize`` as the size of the TCB (16 bytes), the following expression can be used to calcluate ``PADsize`` from the ``PT_TLS`` program header.
+
+``PADsize = (PT_TLS.p_vaddr - TCBsize) mod PT_TLS.p_align``.
```
yury-khrustalev: style: add a `.. code-block::` instead

yury-khrustalev: Same for other formulae in this section. I think it will make reading easier.

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+TLS Descriptors
+---------------
+
+AArch64 uses the TLS Descriptor dialect for the general dynamic model.
+The TLS Descriptor dialect permits a dynamic linker to use the
+location and properties of the TLS symbol to select an optimal
+resolver function.
+
+The static relocations with a prefix of ``R_AARCH64_TLSDESC_``
+targeting TLS symbol ``var``, instruct the static linker to create a
```
yury-khrustalev: nit: superfluous comma after `var`?

sysvabi64/sysvabi64.rst
```diff
@@ -1890,6 +1906,577 @@ See `MemTagABIELF64`_ and `PAuthABIELF64`_ for details of reserved tags.
...
+
+Small Code model;
+
+The static linker is permitted to relax the instructions below to
+Local Exec individually using the relocation directive. The
+instructions do not have to be contiguous.
+
+.. code-block:: asm
+
+    adrp xn, :gottprel: var            // R_AARCH64_TLSIE_ADR_GOTTPREL_PAGE21 var
```
yury-khrustalev: nit: superfluous space before var

yury-khrustalev (commented):
I've added a few minor comments. Otherwise this LGTM and should be approved and merged after the small fixes as per the comments. I think this chapter is very useful and provides good level of detail suitable for this ABI documentation.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: Will have a think to see where best to split out the source parts.

sysvabi64/sysvabi64.rst
smithp35: OK, I wanted to include the pseudo code in case the description wasn't good enough, but if it is then I can remove it. There are other sources to find pseudo code.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: Will have a think to see if I can find a better one.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: ACK

smithp35 (commented):
Thanks for the comments, I'll prepare a new patch, most likely going to be Wednesday.

sysvabi64/sysvabi64.rst
smithp35: ACK

sysvabi64/sysvabi64.rst
smithp35: Will have a think. Perhaps `In descending order of generality:`

sysvabi64/sysvabi64.rst
smithp35: I've decided to remove the source parts as they are out of scope, and everyone knows what they are anyway.

smithp35:
Made changes to address review comments and rebase.

yury-khrustalev:
LGTM. I think this PR is ready to merge.

yury-khrustalev (approved):
LGTM

smithp35:
Thanks for the review. Now merging.

PR #312, opened 2025-02-24
Promote the FP8 ABI status from Alpha to Beta

CarolineConcatto:
FP8 ABI have been implemented in upstream GCC and Clang/LLVM. They're ready to be treated as Beta.


jthackray (approved):
LGTM

stuij:
I think there's enough consensus on moving FP8 to beta from both the LLVM and GCC side. If I hear no objections within 2 days, I'll merge this.

stuij (approved):

sallyarmneale (commented):
All good.

PR #313, opened 2025-02-25
[AAELF64] TLS GOT generating relocations must have 0 addend.

smithp35:
Bring TLS GOT generating relocations in line with non GOT generating relocations in https://github.com/ARM-software/abi-aa/pull/272.

The ABI rule is that static linkers should generate a GOT entry for each unique tuple of (S,A). However static linkers such as GNU ld and lld only generate a unique entry per unique S, and handle A inconsistently. With GNU ld ignoring A and lld adding it after. The only consistent behaviour between implementations is when A is 0.


DanielKristofKiss (approved):
LGTM

smithp35:
Rebase to account for whitespace differences [NFC]

PR #314, opened 2025-02-27
[AAPCS64] Clarify how __bf16 affects HFAs

ostannard:
Currently, it's not completely clear whether a struct containing a mixture of different 16-bit floating point types is a Homogeneous Aggregate or not:

  struct foo {
    __fp16 a;
    __bf16 b;
  };

I think that the best interpretation of the current text is that this is an HFA, because there is only one entry in the "Fundamental Data Types" table, which is the only thing considered in the definition of a Homogeneous Aggregate.

However, the table in the "Mapping of C & C++ built-in data types" section used the formats in a column which otherwise contains Fundamental Data Types, which could be read to imply that the above struct is _not_ an HFA. I don't think this is correct, so I've moved the formats to the notes column, making it clear that the two source types map to the same Fundamental Data Type.

Current compiler behaviour:
* Clang considers this struct to be an HFA, matches the behaviour described in this ABI patch.
* GCC does not consider it to be an HFA, but it also does not think that a struct only containing __bf16 is an HFA either. That's unambiguously incorrect both before and after this patch, so it's just a special case of a broader GCC bug.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -755,7 +755,7 @@ A member of an aggregate that is a Fundamental Data Type may be subdivided into 
 Homogeneous Aggregates
 ^^^^^^^^^^^^^^^^^^^^^^
 
-A Homogeneous Aggregate is a composite type where all of the Fundamental Data Types of the members that compose the type are the same. The test for homogeneity is applied after data layout is completed and without regard to access control or other source language restrictions. Note that for short-vector types the fundamental types are 64-bit vector and 128-bit vector; the type of the elements in the short vector does not form part of the test for homogeneity.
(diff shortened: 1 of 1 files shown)


smithp35 (approved):
LGTM.

I think this is the best option given the existing text. While it does seem strange to consider a mix of `__fp16` and `__bf16` as homogenous, in terms of making good use of the registers in the calling convention it makes sense.

Would like to see if there are any objections on the GNU side?

smithp35:
Agreed to merge this PR with Arm's GNU team. While it is not ideal to have __bf16 and __fp16 mapping to the same underlying natural type, changing it risks breaking too much existing code.

rsandifo-arm:
For the record, I've filed https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120007 to track this on the GCC side.

PR #315, opened 2025-03-04
[LSFE] Correct change 9f262a0c7 to preserve S1 (val) in example code

jthackray:
Amend change 9f262a0c7 to preserve S1 (`val`) in the example code, which was overwritten if the final write failed. Also format the change history correctly.


Wilco1:
Strictly speaking there is nothing that says that 'val' is either valid or returned after the operation. So either version works fine.

sallyarmneale (commented):
All good.

PR #320, opened 2025-04-04
[sysvabi64] Update documentation on GNU ifunc resolvers

yury-khrustalev:
Support for additional AT_HWCAP3,4 fields is being added to Glibc

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1414,30 +1414,57 @@ below.
 GNU C Library IFUNC interface
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The GNU C Library (glibc) ``sys/ifunc.h`` header provides the
-necessary type definitions for defining IFUNC resolvers. The resolver
-is passed at least one, and at most 2 parameters dependent on the
-glibc IFUNC ABI version. A flag is set in the first parameter if a
-second parameter is available. The parameters provide access to the
-HWCAP_ flags.
+The prototype of a GNU indirect function resolver is:
 
 .. code-block:: c
 
-  /* A second argument is passed to the ifunc resolver.  */
-  #define _IFUNC_ARG_HWCAP        (1ULL << 62)
+   ElfW(Addr) ifunc_resolver (uint64_t, const uint64_t *);
 
-  /* Function prototype for an IFUNC resolver.
-  ElfW(Addr) ifunc_resolver (uint64_t, const __ifunc_arg_t *);
+The resolver is passed at least one, and at most 2 parameters dependent
+on the GNU C Library (glibc) IFUNC ABI version. The bits in the first parameter
+will match the ``AT_HWCAP`` settings except for the ``_IFUNC_ARG_HWCAP`` bit
+that has special meaning: if this bit is set, then the second parameter is
+passed to the resolver function.
 
-  /* Second argument to an ifunc resolver.  */
-  struct __ifunc_arg_t
-  {
-    unsigned long _size; /* Size of the struct, so it can grow.  */
-    unsigned long _hwcap;
-    unsigned long _hwcap2;
-  };
+.. code-block:: c
 
-  typedef struct __ifunc_arg_t __ifunc_arg_t;
+   /* A second parameter is passed to the ifunc resolver.  */
+   #define _IFUNC_ARG_HWCAP (1ULL << 62)
+
+When the second parameter is passed, it will be a const pointer to a buffer
+that allows to access all HWCAP elements. The buffer may contain the following
+fields:
+
+``_size``
+   A ``uint64_t`` value at byte offset 0 containing the size of the buffer
+   in bytes, this field is always present
+
+``_hwcap``
+   A ``uint64_t`` value at byte offset 8 matching the AT_HWCAP settings,
+   this field is always present
+
+``_hwcap2``
+   A ``uint64_t`` value at byte offset 16 matching the AT_HWCAP2 settings,
+   this field is always present
+
+``_hwcap3``
+   A ``uint64_t`` value at byte offset 24 matching the AT_HWCAP3 settings
+
+``_hwcap4``
+   A ``uint64_t`` value at byte offset 32 matching the AT_HWCAP4 settings
+
+IFUNC resolver functions must use the value of the ``_size`` field to check
+how many HWCAP fields are available.
+
+The glibc header ``sys/ifunc.h`` provides the necessary type definitions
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -1435,10 +1435,17 @@ HWCAP_ flags.
     unsigned long _size; /* Size of the struct, so it can grow.  */
     unsigned long _hwcap;
     unsigned long _hwcap2;
+    unsigned long _hwcap3;
+    unsigned long _hwcap4;
   };
 
   typedef struct __ifunc_arg_t __ifunc_arg_t;
 
+The second paramter is a pointer to a ``__ifunc_arg_t`` object and its
```
enh-google, line 1444: parameter

enh-google, line 1444: (also: mention that this is only true if `_IFUNC_ARG_HWCAP` is set in the _first_ parameter?)

enh-google (approved):

sysvabi64/sysvabi64.rst
yury-khrustalev: Fixed, thank you

sysvabi64/sysvabi64.rst
```diff
@@ -1417,9 +1417,7 @@ GNU C Library IFUNC interface
 The GNU C Library (glibc) ``sys/ifunc.h`` header provides the
 necessary type definitions for defining IFUNC resolvers. The resolver
 is passed at least one, and at most 2 parameters dependent on the
-glibc IFUNC ABI version. A flag is set in the first parameter if a
-second parameter is available. The parameters provide access to the
-HWCAP_ flags.
+glibc IFUNC ABI version.
```
enh-google, line 1420: probably clearer to move this sentence to the start of the next paragraph, after the code?

sysvabi64/sysvabi64.rst
```diff
@@ -1435,10 +1433,20 @@ HWCAP_ flags.
...
   };
 
   typedef struct __ifunc_arg_t __ifunc_arg_t;
 
+The ``_IFUNC_ARG_HWCAP`` bit will be set in the first parameter when the
+second parameter is available. When this is the case, the second parameter
+is a pointer to a ``__ifunc_arg_t`` object and its ``_size`` field is set
+to ``sizeof(__ifunc_arg_t)``. IFUNC resolver functions can use the value
+of the ``_size`` field to check if additional ``_hwcap`` fields are available.
+When the ``_IFUNC_ARG_HWCAP`` bit is not set, only one parameter is passed
```
enh-google, line 1447: (this makes it sound like the first parameter is _only_ _IFUNC_ARG_HWCAP in the "modern" case, whereas in actual fact ifuncs that only need AT_HWCAP can always just check the bits in the first argument and ignore the second entirely.)

enh-google (commented):
(i think all of these are "not a regression", and this change is already strictly better than what's committed, but since we're here...)

yury-khrustalev:
I will come back to this PR but first I'd like this discussion in the public list to come to some conclusion: https://inbox.sourceware.org/libc-alpha/Z_UCLjazSkUPjOVs@arm.com/

sysvabi64/sysvabi64.rst
```diff
@@ -1417,9 +1417,7 @@ GNU C Library IFUNC interface
 The GNU C Library (glibc) ``sys/ifunc.h`` header provides the
 necessary type definitions for defining IFUNC resolvers. The resolver
 is passed at least one, and at most 2 parameters dependent on the
-glibc IFUNC ABI version. A flag is set in the first parameter if a
-second parameter is available. The parameters provide access to the
-HWCAP_ flags.
+glibc IFUNC ABI version.
 
 .. code-block:: c
```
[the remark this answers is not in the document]
yury-khrustalev: I've pushed new version of this patch where this section is re-written, so hopefully it is more clear now. If you think that expanding comments in the code listings is useful, I think we could do that, but I suppose that text should cover it anyway.

yury-khrustalev:
Revised Glibc patch (v4): https://inbox.sourceware.org/libc-alpha/20250425113522.3310684-1-yury.khrustalev@arm.com/

yury-khrustalev:
Revised Glibc patch (v5): https://inbox.sourceware.org/libc-alpha/20250428113220.2660956-1-yury.khrustalev@arm.com/

sysvabi64/sysvabi64.rst
```diff
@@ -1414,30 +1414,57 @@ below.
...
+
+   /* A second parameter is passed to the ifunc resolver.  */
+   #define _IFUNC_ARG_HWCAP (1ULL << 62)
+
+When the second parameter is passed, it will be a const pointer to a buffer
+that allows to access all HWCAP elements. The buffer may contain the following
+fields:
+
+``_size``
+   an ``uint64_t`` value at byte offset 0 containing the size of the buffer
```
smithp35, line 1439: In the category of English is an awful language, whilie u is a vowel and if we spelled out unsigned integer, `an` would be correct, when it is an abbreviation `an uint64_t` doesn't read well. It seems like `A uint64_t` is more common and reads a lot better.

We could potentially avoid that by just starting with ``uint64_t``

smithp35 (commented):
Updated text looks good to me. I think we may want to avoid `an uint64_t` (cosmetic).

sysvabi64/sysvabi64.rst
enh-google: or just 'unsigned long', since that's what the corresponding code uses...

yury-khrustalev: @enh-google what are your thoughts on using `uint64_t` in the libc header instead of `unsigned long`? We really use those value as 64 bit integers and `unsigned long` is not always 64 bit on some platforms.

enh-google: but this is specifically only for arm64 :-)

(my personal preference in new api is always for the more specific types, but i suspect what happened here was that they copied the type from getauxval(), which is where these values come from. so left as `unsigned long` it's potentially reusable for other architectures too, where `uint64_t` would make it LP64-only. i have no idea whether glibc has any plans to use this for other architectures. android only supports arm64, riscv64, and x86-64, and they're all sufficiently different from one another that the sharing wouldn't really make sense.)

yury-khrustalev: I've fixed the cosmetic suggestion and kept `uint64_t` because I think it's more correct to describe it this way. We use bits above 32 in both `hwcap` and `hwcap2` already. I'll keep `unsigned long` in the header though.

DanielKristofKiss (approved):

smithp35:
PR has been approved. Merging into main.

Issue #321, opened 2025-04-07
Clarify "base" level BTI requirements for a toolchain


smithp35:
https://github.com/ARM-software/abi-aa/pull/282 added some text alongside the GNU property description of BTI to define what a code-generator could assume about the static linker.

While this may have been the most convenient place to put the information. This is strictly speaking not just restricted to the sysvabi. Bare-metal and non-sysv platforms may use BTI. 

Thoughts:
The https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#577call-and-jump-relocations should mention that static linkers cannot assume a BTI at the destination. Add a reference to the sysvabi text for more details.

Being "indirectly callable" could be considered part of the Procedure Call standard too, on requirements for callees when BTI is enabled.

PR #322, opened 2025-04-08
[AAELF64] BTI and veneers for bare-metal contexts.

smithp35:
Range extension veneers in a bare-metal context must not assume a BTI instruction is at the destination.

There is documentation in sysvabi64 that states this, but strictly speaking this need not apply to a bare-metal context. Document that linker generated veneers cannot assume a BTI for definitions in the same link unit.

Bare-metal systems do not support PLT or iPLT (ifunc) so the requirements surrounding code-generation can stay in the sysvabi64.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1354,6 +1354,7 @@ In some systems indirect calls may also use veneers in order to support dynamic 
 
 On platforms that do not support dynamic pre-emption of symbols, an unresolved weak reference to a symbol relocated by ``R_<CLS>_CALL26`` shall be treated as a jump to the next instruction (the call becomes a no-op). The behaviour of ``R_<CLS>_JUMP26`` and ``R_<CLS>_PLT32`` in these conditions is not specified by this standard.
 
+When a static linker uses an indirect branch in a veneer, it may not assume that a destination in the same link unit has a BTI instruction, or BTI compatible instruction, at the destination of the indirect branch. See `SYSVABI64`_ for additional requirements on code-generators and PLT sequences.
 
 Group relocations
 ^^^^^^^^^^^^^^^^^
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1365,7 +1365,7 @@ In some systems indirect calls may also use veneers in order to support dynamic 
 
 On platforms that do not support dynamic pre-emption of symbols, an unresolved weak reference to a symbol relocated by ``R_<CLS>_CALL26`` shall be treated as a jump to the next instruction (the call becomes a no-op). The behaviour of ``R_<CLS>_JUMP26`` and ``R_<CLS>_PLT32`` in these conditions is not specified by this standard.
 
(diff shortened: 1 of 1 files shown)


smithp35:
Addresses the indirect veneer part of https://github.com/ARM-software/abi-aa/issues/321

aaelf64/aaelf64.rst
```diff
@@ -1354,6 +1354,7 @@ In some systems indirect calls may also use veneers in order to support dynamic
 
 On platforms that do not support dynamic pre-emption of symbols, an unresolved weak reference to a symbol relocated by ``R_<CLS>_CALL26`` shall be treated as a jump to the next instruction (the call becomes a no-op). The behaviour of ``R_<CLS>_JUMP26`` and ``R_<CLS>_PLT32`` in these conditions is not specified by this standard.
 
+When a static linker uses an indirect branch in a veneer, it may not assume that a destination in the same link unit has a BTI instruction, or BTI compatible instruction, at the destination of the indirect branch. See `SYSVABI64`_ for additional requirements on code-generators and PLT sequences.
```
statham-arm, line 1357: The sentiment LGTM. Some thoughts about clarification:

1. possibly not all readers know what a BTI is, so does this need to be prefixed with some kind of condition "When BTI branch protection is in use, ..." and a link of some kind so they can at least find out that it's a CPU feature they aren't using?
2. this only says what linkers _can't_ do, and not what they should do instead. Do you need to spell out in full "... and so if the linker needs to add a long-branch thunk using an indirect call instruction, it must insert a landing pad containing a BTI which branches to the destination directly"?

But maybe I'm being too cautious, and the typical reader of this document who isn't using BTI will at least recognise the name as "some CPU feature I don't need to worry about", and one who is will be able to figure out for themself what they'd have to do in the linker to satisfy this requirement?

smithp35: Thanks for the comments. I've had another go.

I put BTI in the terms and definitions section at the top, and have given more of a suggestion about what linker's should do.

aaelf64/aaelf64.rst
```diff
@@ -1354,6 +1365,7 @@ In some systems indirect calls may also use veneers in order to support dynamic
 
 On platforms that do not support dynamic pre-emption of symbols, an unresolved weak reference to a symbol relocated by ``R_<CLS>_CALL26`` shall be treated as a jump to the next instruction (the call becomes a no-op). The behaviour of ``R_<CLS>_JUMP26`` and ``R_<CLS>_PLT32`` in these conditions is not specified by this standard.
 
+In a link-unit that is intended to be used when BTI guarded pages are enable, veneers created by the static linker that use an indirect branch must target a BTI instruction or a BTI compatible instruction. If the destination of the veneer is in the same link unit and does not have a BTI or BTI compatible instruction, the static linker must generate an additional veneer that has a BTI instruction followed by a transfer of control to the destination that does not use an indirect branch. See `SYSVABI64`_ for additional requirements on code-generators and PLT sequences.
```
statham-arm, line 1368: nit: "are enable" → "are enabled"

statham-arm (approved):
Thanks, the changes LGTM (with one tiny grammar nit).

aaelf64/aaelf64.rst
smithp35: ACK, thanks for spotting.

PR #324, opened 2025-04-08
[SYSVABI] Add example that ifunc resolvers are indirectly called

smithp35:
An ifunc resolver function needs a BTI as it is called indirectly via the dynamic loader when resolving the R_AARCH64_IRELATIVE relocation. When static linking the relocation is also resolved via an indirect branch to the resolver from a local resolver function like _dl_relocate_static_pie in glibc.


smithp35:
FWIW both clang and gcc add a BTI to the ifunc resolver already. This is just to make sure it is covered in the documentation.

DanielKristofKiss (approved):
LGTM

PR #325, opened 2025-04-23
Some tweaks to the SME lazy save scheme

rsandifo-arm:
This patch does four things:

- Explicitly document that ZT0 is a temporary register.

- Clarify that __arm_sme_save and __arm_sme_restore treat both ZA
  and ZT0 specially.

- Add a general note about how platforms should handle asynchronous control
  flow (such as POSIX signals).  The specific details are out of scope
  for the AAPCS64, which is why this wasn't mentioned before.  However,
  for the sake of an example, we can at least sketch the approach that
  linux will use.

- Change the notes around setjmp and longjmp to recommend that both
  routines turn ZA off.  Just turning ZA off in longjmp isn't enough
  when jumping out of a linux-style signal handler, since longjmp
  would not be able to see any ZA save buffer that was in use when the
  signal was raised, nor the ZA contents that should be saved to it.

  The previous recommendation to turn ZA off only in longjmp was chosen
  by analogy with exception unwinding; it seemed to make sense in both
  cases to commit a lazy save once it was known that some frames would
  be unwound.  However, the two cases are not analogous for asynchronous
  signals, since exception unwinders can examine each intermediate frame
  and see the point at which a signal handler was entered, whereas
  longjmp (in general) cannot.

  It may not be strictly necessary for longjmp to commit a lazy save
  if setjmp does; it might be enough for it to force PSTATE.ZA and
  TPIDR2_EL0 to zero.  However, that would be a novel state transition
  and it would be optimising for what should be a vanishingly rare case
  (the one where there is an in-progress SME routine between a setjmp
  and a longjmp).  It seems safer to use the standard transition instead.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -264,6 +264,12 @@ changes to the content of the document for that release.
 |            |                    | - Add agnostic-ZA interface and routines to save/restore SME     |
 |            |                    |   state.                                                         |
 +------------+--------------------+------------------------------------------------------------------+
+|            |                    | - Explicitly say that ZT0 is a temporary register.               |
+|            |                    | - Add a note about the interaction between the SME lazy save     |
+|            |                    |   scheme and asynchronous transfers of control.                  |
+|            |                    | - Recommend that ``setjmp`` as well as ``longjmp`` call          |
+|            |                    |   ``__arm_za_disable``.                                          |
++------------+--------------------+------------------------------------------------------------------+
 
 References
 ^^^^^^^^^^
@@ -414,7 +420,7 @@ Global register
 Program state
    The state of the program’s memory, including values in machine registers.
 
-Scratch register, temporary register, caller-saved register
+Scratch register, _`temporary register`, caller-saved register
    A register used to hold an intermediate value during a calculation (usually, such values are not named in the program source and have a limited lifetime). If a function needs to preserve the value held in such a register over a call to another function, then the calling function must save and restore the value.
 
 Callee-saved register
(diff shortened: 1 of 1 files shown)


smithp35 (commented):
Spotted a typo in the change history.

yury-khrustalev:
Acked-by: @yury-khrustalev

aapcs64/aapcs64.rst
```diff
@@ -1729,6 +1738,69 @@ it induces ``memset`` to modify ``BLK`` before ``memset`` has returned:
...
+* On a normal return from the handler, restore the saved SME state
+  and resume execution.
+
+* Make ``setjmp`` and ``longjmp`` turn ZA off; see `setjmp and longjmp`_
+  for details.
+
+* When unwinding out of a handler, use the saved TPIDR2_EL0
+  information to see whether ZA was previously dormant; that is,
+  if TPIDR2_EL0 pointed to a TPIDR2 block that has a nonnull ZA save
+  buffer.  If so, copy the saved ZA contents to this buffer.
```
sdesmalen-arm, line 1781: Should this also mention that the unwinder must commit the lazy-save and turn ZA off?

smithp35 (approved):
LGTM.

Agreed to approve with stakeholders in Arm's libraries and kernel team. Will merge at end of the UK working day.

aapcs64/aapcs64.rst
sdesmalen-arm: In that case, would be useful to amend it with something like `, followed by any operations to implement the existing requirements for unwinding exceptions, as described in 'Exceptions'`, or to slightly rephrase the sentence below ("See 'Exceptions' for more information) to make it clear that the above steps are additive to the existing requirements?

The reason I ask, is because that wasn't immediately clear to me from reading the text.

sdesmalen-arm (approved):
Thanks for the rewording @rsandifo-arm. LGTM!

smithp35:
PR has been approved. Merging into main.

Issue #176, opened 2022-10-27
[aaelf64] No static TLS 64-bit relocation with expression DTPREL(S+A)


smithp35:
Both clang and GCC leave out the `DW_AT_location` for TLS variables ostensibly due to lack of a relocation to describe the location in the debug data.

For x86_64 `DW_AT_location` is emitted as:
`DW_AT_location (DW_OP_const8u 0x0, DW_OP_form_tls_address)`
With the relocation:
`R_X86_64_DTPOFF64` to the symbol definition.

In AArch64 the closest equivalent is `R_AArch64_TLS_DTPREL` although it is marked as a dynamic relocation (normally used for the GOT entry created by the global-dynamic in the traditional dialect). If we marked this relocation as static and dynamic (and made sure that static linkers could handle it in a static context), then we could support `DW_AT_location`

asl:
Tagging @andreil99

smithp35:
Raised https://github.com/ARM-software/abi-aa/pull/330 . This follows x86_64 and PPC by using R_AARCH64_TLS_DTPREL for the relocation. An alternative implementation is to create a new static local dynamic relocation equivalent to R_ARM_TLS_LDO32. The calculation would be the same in both cases.

smithp35:
Thanks for the reminder. Just catching up from the Christmas break. I'll raise this again internally to see if we can get a consensus on the relocation number.

I may need to separate out the syntax for the data directive, but that is "implementation detail" and not ABI.

PR #330, opened 2025-05-20
[AAELF64] Allow R_AARCH64_TLS_DTPREL to be used statically.

smithp35:
Permit the R_AARCH64_TLS_DTPREL dynamic relocation to be used statically so that debug information can describe the location of TLS variables.

All code sequences to access TLS data use immediates, hence all existing static TLS relocations are instruction relocations.

To describe the location of a TLS variable in debug information requires a static data relocation. The necessary expression needed is DTPREL(S + A) which is performed by the traditional dialect dynamic relocation R_AARCH64_TLS_DTPREL.

There is prior art in the x86_64 and PPC64 ABIs to use the equivalent relocation R_X86_64_DTPOFF64 and R_PPC_DTPREL64 respectively for relocating debug information statically.

The 32-bit Arm ABI uses the R_ARM_TLS_LDO32 static local dynamic relocation. The advantage of using local dynamic is that these only apply to a TLS variable local to the module so no additional note is needed. An alternative design adds a new relocation code to local dynamic.

fixes: https://github.com/ARM-software/abi-aa/issues/176

a static relocation with the necessary

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -291,6 +291,8 @@ changes to the content of the document for that release.
   | 2025Q2        | 9\ :sup:`th`       | - In `Call and Jump relocations`_ added |
   |               | April 2025         |   static linker requirements on veneers |
   |               |                    |   when BTI guarded pages are used.      |
+  |               |                    | - R_AARCH64_TLS_DTPREL can be used as a |
+  |               |                    |   static relocation as well as dynamic  |
   +---------------+--------------------+-----------------------------------------+
 
 References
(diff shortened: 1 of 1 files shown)


smithp35:
I've chosen to use %operator(expr) for data directives following the conversation in https://github.com/llvm/llvm-project/issues/132570 based on https://maskray.me/blog/2025-03-16-relocation-generation-in-assemblers

sysvabi64/sysvabi64.rst
```diff
@@ -649,6 +649,23 @@ syntax is of the form ``#:<operator>:<symbol name>``
 
    PageBreak oneColumn
 
+Assembler language addressing mode conventions for data
+-------------------------------------------------------
+
+The assembler operators that apply to instructions use the syntax
+``#:<operator>:<symbol name>``. This syntax is not easy to apply to
+data directives due to a parsing ambiguity with labels.
```
jrtc27, line 665: Is that true if you make `.directive #foo` legal, but redundant, unless there's an operator, at which point it disambiguates? Not saying it's nicer syntax, but it would avoid having the odd syntax split between instructions and data if so.

smithp35: Apologies for the delay in responding. I have been taking the statement in https://maskray.me/blog/2025-03-16-relocation-generation-in-assemblers on trust

> Applying this syntax to data directives, however, could create parsing ambiguity. In both GNU Assembler and LLVM, .word :plt:fun would be interpreted as .word: plt: fun, treating .word and plt as labels, rather than achieving the intended meaning.

From the GNU as manual https://sourceware.org/binutils/docs/as/Statements.html
> A label is a symbol immediately followed by a colon (:). Whitespace before a label or after a colon is permitted, but you may not have whitespace between a label’s symbol and its colon. See [Labels](https://sourceware.org/binutils/docs/as/Labels.html).

At least from the perspective of the syntax `.word :plt:fun` is not legal when written as `.word: plt: fun` however MaskRay may be referring to the implementations, and changing them to support .word :plt:fun may be non trivial.

We'd need to do some further investigation to clarify.

PR #331, opened 2025-06-02
[aapcs64] Document ABI stance on FFR

acoplan-arm:
The ABI doesn't currently say anything about the first fault register (FFR). The de facto treatment is that there is no guarantee about the contents of the FFR across a public interface boundary, i.e. it is a temporary register (in the language of the PCS).  Document this.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -269,6 +269,7 @@ changes to the content of the document for that release.
 |            |                    |   scheme and asynchronous transfers of control.                  |
 |            |                    | - Recommend that ``setjmp`` as well as ``longjmp`` call          |
 |            |                    |   ``__arm_za_disable``.                                          |
+|            |                    | - Explicitly say that the FFR is a temporary register.           |
 +------------+--------------------+------------------------------------------------------------------+
 
 References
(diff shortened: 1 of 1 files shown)


smithp35:
Thanks for the patch. Would you be able to add an entry in the change history?

aapcs64/aapcs64.rst
```diff
@@ -960,6 +961,15 @@ the subroutine must ensure that p4-p15 are preserved across the
 call. In other cases it need not preserve any scalable predicate
 register contents.
 
+First Fault Register (FFR)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The special-purpose First Fault Register (FFR) has the same size and
+format as the scalable predicate registers, and is available if and only
```
stuij, line 968: `...and is available if and only they are.`: I think you dropped an if?

I'm guessing this should read: `..and is available if and only if they are.`

I would personally prefer: `..and is only available if they are.`, to make the sentence a bit more clear/readable.

acoplan-arm: Ah, I did drop an if indeed, thanks for spotting that! I was trying to make this consistent with the section on Scalable Predicate Registers above, which says:
> These registers are available if and only if the scalable vector registers are available

Is it OK if I add the missing "if" to keep this consistent with the above, or would you still prefer a re-word?

stuij: Yea, perhaps word it the same. I've reread it to see if I'm being pedantic about it. I probably am, but by the time we've arrived at the last `they` I'm getting slightly confused about what `they` was again.

I'd just spell it out: `.. are available if and only if the scalable vector registers are available.`

acoplan-arm: Thanks, I've reworded this as suggested in the latest version.

stuij (approved):

smithp35:
PR approved by two people, merging.

PR #332, opened 2025-06-05
[FMV][AArch64] Document the interface for Function Multi-versioning.

labrinea:
Documents the Application Binary Interface requirements for
Function Multi-versioning (FMV) on System V platforms.

A global variable `__aarch64_cpu_features` provided by
the runtime library contains information about the available
CPU features. The variable must be defined as DSO-local with
its symbol visibility set to `STV_HIDDEN`.

A function `__init_cpu_features_resolver` provided by
the runtime library is used for initializing the variable
`__aarch64_cpu_features`. The function must be defined as
DSO-local with its symbol visibility set to `STV_HIDDEN`.

Accessing `__aarch64_cpu_features` is reserved for the
compiler generated code or the runtime library. If the
variable is only accessed for FMV, then it may be placed
in the Relocation Read Only (RELRO) program segment to
prevent it from being modified. Otherwise, if the variable
is used besides FMV the runtime library must ensure that
it is initialized via a constructor function and cannot
be placed in the `RELRO` segment.

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1598,16 +1598,19 @@ Function Multi-versioning
 Function Multi-versioning (FMV) is an Arm C Language Extension that
 lets the compiler generate multiple function versions and auto-dispatch
 between them. Each of the function versions is specialized for a set
-of architecture extensions. The most suitable version is selected at
-load time. This requires runtime information about the CPU features
+of architecture extensions. The most suitable version is selected
+dynamically. This requires runtime information about the CPU features
 available on the host. FMV is supported on GNU/Linux, Android, and
 many of the BSD operating systems.
 
 On System V platforms Function Multi-versioning is implemented using
 GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
-uses a global variable ``__aarch64_cpu_features`` that contains a
-description of the detected CPU features to select the best function
-version:
+may rely on a runtime library for selecting the most suitable function
+version, but it is not a requirement.
+
+If a runtime library is used, it must ensure that a global variable
+``__aarch64_cpu_features`` containing a description of the detected
+CPU features is populated before the application code reaches main.
 
 .. code-block:: c
 
@@ -1716,11 +1719,22 @@ The variable may contain the following fields:
 A special value ``FEAT_INIT = (1U << 63);`` is used to signify
 initialization completion.
 
+Implementing FMV using ``__aarch64_cpu_features`` is not required.
+Accessing ``__aarch64_cpu_features`` from outside a FMV resolver
+function is not well defined.
+
 Initialization of the ``__aarch64_cpu_features`` is performed by
-the runtime. Each FMV resolver function will call the
-``__init_cpu_features_resolver`` function that must be provided
-by the runtime environment. This runtime function must ensure
-that ``__aarch64_cpu_features`` is populated.
+the runtime library. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function at load time. This runtime
+function must ensure that ``__aarch64_cpu_features`` is populated.
+
+Systems without support for IFUNC may call the
+``__init_cpu_features_resolver`` function from an
+``__attribute__((constructor))`` function provided by the runtime.
+
+Systems with support for IFUNC should place the global variable
+``__aarch64_cpu_features`` in Relocation Read Only (RELRO) segment
+to prevent it from being modified after the FMV resolvers have run.
 
 The ``__init_cpu_features_resolver`` function has the following
 prototype:
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1716,12 +1716,13 @@ The variable may contain the following fields:
     +-------------------+----------+
 
 Implementing FMV using ``__aarch64_cpu_features`` is not required.
-Accessing this variable from outside a FMV resolver function is not
-well defined. If the variable is only accessed by the FMV resolvers,
-then it may be placed in the `Relocation Read Only (RELRO)`_ program
-segment to prevent it from being modified after the FMV resolvers
-have run. The variable must be defined as DSO-local with its symbol
-visibility set to ``STV_HIDDEN``.
+Accessing ``__aarch64_cpu_features`` is reserved for the compiler
+generated code or the runtime library. If the variable is only
+accessed by the FMV resolvers, then it may be placed in the
+`Relocation Read Only (RELRO)`_ program segment to prevent it from
+being modified after the FMV resolvers have run. The variable must
+be defined as DSO-local with its symbol visibility set to
+``STV_HIDDEN``.
 
 .. note::
 
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1715,6 +1715,11 @@ The variable may contain the following fields:
     | FEAT_MOPS         | 1U << 60 |
     +-------------------+----------+
 
+A special value ``FEAT_EXT = (1U << 63);`` is reserved to indicate
+presence of additional feature fields exceeding
+``__aarch64_cpu_features``.
+
+
 Implementing FMV using ``__aarch64_cpu_features`` is not required.
 Accessing this variable from outside a FMV resolver function is
 not well defined. The variable may be placed in the
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1715,11 +1715,6 @@ The variable may contain the following fields:
     | FEAT_MOPS         | 1U << 60 |
     +-------------------+----------+
 
-A special value ``FEAT_EXT = (1U << 63);`` is reserved to indicate
-presence of additional feature fields exceeding
-``__aarch64_cpu_features``.
-
-
 Implementing FMV using ``__aarch64_cpu_features`` is not required.
 Accessing this variable from outside a FMV resolver function is
 not well defined. The variable may be placed in the
@@ -1727,6 +1722,17 @@ not well defined. The variable may be placed in the
 from being modified after the FMV resolvers have run. The
 variable must be defined as DSO-local with its symbol visibility
 set to ``STV_HIDDEN``.
+
+.. note::
+
+Both the SME support routines (see AAPCS64_ for more information)
+and the compiler built-in function ``__builtin_cpu_supports`` rely
+on the ``__aarch64_cpu_features`` variable for detecting CPU features.
+Therefore the runtime library must ensure that the variable is
+initialized prior to their usage. FMV support is not required for
+using the SME support routines or the ``__builtin_cpu_supports``
+function. The ``__aarch64_cpu_features`` variable should not be
+placed in ``RELRO`` if no FMV resolver has run.
 
 The ``__init_cpu_features_resolver`` function has the following
 prototype:
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1715,14 +1715,13 @@ The variable may contain the following fields:
     | FEAT_MOPS         | 1U << 60 |
     +-------------------+----------+
 
-A special value ``FEAT_INIT = (1U << 63);`` is used to signify
-initialization completion.
-
 Implementing FMV using ``__aarch64_cpu_features`` is not required.
 Accessing this variable from outside a FMV resolver function is
 not well defined. The variable may be placed in the
 `Relocation Read Only (RELRO)`_ program segment to prevent it
-from being modified after the FMV resolvers have run.
+from being modified after the FMV resolvers have run. The
+variable must be defined as DSO-local with its symbol visibility
+set to ``STV_HIDDEN``.
 
 The ``__init_cpu_features_resolver`` function has the following
 prototype:
@@ -1734,7 +1733,7 @@ prototype:
 The above interface expects the same parameters as a GNU Indirect
 Function resolver. See `GNU C Library IFUNC interface`_. Other
 platforms may use a different interface with the runtime library.
-However, all implementations must provide a DSO local definition
+However, all implementations must provide a DSO-local definition
 of the function by setting the symbol visibility to ``STV_HIDDEN``.
 
 Initialization and Termination Functions
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1599,18 +1599,17 @@ Function Multi-versioning (FMV) is an Arm C Language Extension that
 lets the compiler generate multiple function versions and auto-dispatch
 between them. Each of the function versions is specialized for a set
 of architecture extensions. The most suitable version is selected
-dynamically. This requires runtime information about the CPU features
+at load time. This requires runtime information about the CPU features
 available on the host. FMV is supported on GNU/Linux, Android, and
 many of the BSD operating systems.
 
 On System V platforms Function Multi-versioning is implemented using
 GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
-may rely on a runtime library for selecting the most suitable function
-version, but it is not a requirement.
-
-If a runtime library is used, it must ensure that a global variable
-``__aarch64_cpu_features`` containing a description of the detected
-CPU features is populated before the application code reaches main.
+may rely on the presence of a global variable ``__aarch64_cpu_features``
+provided by the runtime library. It contains information about the
+available CPU features. The runtime library must also provide a
+function ``__init_cpu_features_resolver`` that the IFUNC resolver
+can call to initialize ``__aarch64_cpu_features``.
 
 .. code-block:: c
 
@@ -1720,21 +1719,10 @@ A special value ``FEAT_INIT = (1U << 63);`` is used to signify
 initialization completion.
 
 Implementing FMV using ``__aarch64_cpu_features`` is not required.
-Accessing ``__aarch64_cpu_features`` from outside a FMV resolver
-function is not well defined.
-
-Initialization of the ``__aarch64_cpu_features`` is performed by
-the runtime library. Each FMV resolver function will call the
-``__init_cpu_features_resolver`` function at load time. This runtime
-function must ensure that ``__aarch64_cpu_features`` is populated.
-
-Systems without support for IFUNC may call the
-``__init_cpu_features_resolver`` function from an
-``__attribute__((constructor))`` function provided by the runtime.
-
-Systems with support for IFUNC should place the global variable
-``__aarch64_cpu_features`` in Relocation Read Only (RELRO) segment
-to prevent it from being modified after the FMV resolvers have run.
+Accessing this variable from outside a FMV resolver function is
+not well defined. The variable may be placed in the
+`Relocation Read Only (RELRO)`_ program segment to prevent it
+from being modified after the FMV resolvers have run.
 
 The ``__init_cpu_features_resolver`` function has the following
 prototype:
@@ -1744,10 +1732,10 @@ prototype:
    void __init_cpu_features_resolver (uint64_t, const uint64_t *);
 
 The above interface expects the same parameters as a GNU Indirect
-Function resolver. Other platforms may use a different interface
-with the runtime library. However, all implementations must provide
-a DSO local definition of the function by setting the symbol
-visibility to ``STV_HIDDEN``.
+Function resolver. See `GNU C Library IFUNC interface`_. Other
+platforms may use a different interface with the runtime library.
+However, all implementations must provide a DSO local definition
+of the function by setting the symbol visibility to ``STV_HIDDEN``.
 
 Initialization and Termination Functions
 ----------------------------------------
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1592,39 +1592,26 @@ IFUNC requirements for dynamic linkers
 To resolve an ``R_AARCH64_IRELATIVE`` relocation the dynamic linker
 performs the calculation described in AAELF64_ Dynamic Relocations.
 
-Function Multi-Versioning
+Function Multi-versioning
 -------------------------
 
-Function Multi-Versioning (FMV) is an Arm C Language Extension that
+Function Multi-versioning (FMV) is an Arm C Language Extension that
 lets the compiler generate multiple function versions and auto-dispatch
 between them. Each of the function versions is specialized for a set
-of architecture extensions. The most suitable version is selected
-dynamically. This requires runtime information about the CPU features
-available on the host. FMV is supported on GNU/Linux, Android, Windows
-and many of the BSD operating systems, including Darwin.
+of architecture extensions. The most suitable version is selected at
+load time. This requires runtime information about the CPU features
+available on the host. FMV is supported on GNU/Linux, Android, and
+many of the BSD operating systems.
 
-The initialization of the runtime is platform dependent. It may be
-performed at startup by calling a constructor (Windows), or it may be
-performed by calling ``__init_cpu_features_resolver`` (otherwise).
-On GNU/Linux this function is called at load time and has the following
-prototype:
-
-.. code-block:: c
-
-   void __init_cpu_features_resolver (uint64_t, const uint64_t *);
-
-The above interface expects the same parameters as a GNU Indirect
-Function resolver. Other platforms may use a different interface
-with the runtime library. However, all implementations must provide
-a DSO local definition of the function by setting the symbol
-visibility to ``STV_HIDDEN``.
-
-The runtime initialization must set a global variable which contains
-the bits corresponding to the CPU features that have been detected:
+On System V platforms Function Multi-versioning is implemented using
+GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
+uses a global variable ``__aarch64_cpu_features`` that contains a
+description of the detected CPU features to select the best function
+version:
 
 .. code-block:: c
 
-   uint64_t __aarch64_cpu_features
+   uint64_t __aarch64_cpu_features = 0;
 
 The variable may contain the following fields:
 
@@ -1725,6 +1712,28 @@ The variable may contain the following fields:
     +-------------------+----------+
     | FEAT_MOPS         | 1U << 60 |
     +-------------------+----------+
+
+A special value ``FEAT_INIT = (1U << 63);`` is used to signify
+initialization completion.
+
+Initialization of the ``__aarch64_cpu_features`` is performed by
+the runtime. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function that must be provided
+by the runtime environment. This runtime function must ensure
+that ``__aarch64_cpu_features`` is populated.
+
+The ``__init_cpu_features_resolver`` function has the following
+prototype:
+
+.. code-block:: c
+
+   void __init_cpu_features_resolver (uint64_t, const uint64_t *);
+
+The above interface expects the same parameters as a GNU Indirect
+Function resolver. Other platforms may use a different interface
+with the runtime library. However, all implementations must provide
+a DSO local definition of the function by setting the symbol
+visibility to ``STV_HIDDEN``.
 
 Initialization and Termination Functions
 ----------------------------------------
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1725,14 +1725,14 @@ set to ``STV_HIDDEN``.
 
 .. note::
 
-Both the SME support routines (see AAPCS64_ for more information)
-and the compiler built-in function ``__builtin_cpu_supports`` rely
-on the ``__aarch64_cpu_features`` variable for detecting CPU features.
-Therefore the runtime library must ensure that the variable is
-initialized prior to their usage. FMV support is not required for
-using the SME support routines or the ``__builtin_cpu_supports``
-function. The ``__aarch64_cpu_features`` variable should not be
-placed in ``RELRO`` if no FMV resolver has run.
+   Both the SME support routines (see AAPCS64_ for more information)
+   and the compiler built-in function ``__builtin_cpu_supports`` rely
+   on the ``__aarch64_cpu_features`` variable for detecting CPU features.
+   Therefore the runtime library must ensure that the variable is
+   initialized prior to their usage. FMV support is not required for
+   using the SME support routines or the ``__builtin_cpu_supports``
+   function. The ``__aarch64_cpu_features`` variable should not be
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,140 @@ IFUNC requirements for dynamic linkers
...
 
+Function Multi-Versioning
+-------------------------
+
+Function Multi-Versioning (FMV) is an Arm C Language Extension that
+lets the compiler generate multiple function versions and auto-dispatch
+between them. Each of the function versions is specialized for a set
+of architecture extensions. The most suitable version is selected
+dynamically. This requires runtime information about the CPU features
+available on the host. FMV is supported on GNU/Linux, Android, Windows
```
smithp35: I think the SystemV document should concentrate on ELF SystemV platforms. I think we should concentrate on the requirements for an ELF runtime to support FMV.

Windows and Mac have their own platform ABI and would be best documented alongside that by the platform owner.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,140 @@ IFUNC requirements for dynamic linkers
...
+and many of the BSD operating systems, including Darwin.
+
+The initialization of the runtime is platform dependent. It may be
```
smithp35: I think we need to focus on what the minimum requirements on the runtime are to support FMV, and not put too much of the details of the current implementation. For example the constructor part is specific to Windows and a conforming SystemV runtime could omit the constructor and still work. With the current code-generation, a SystemV runtime must support ifuncs though and can't use the constructor as a replacement for them.

My suggestion for what to put here:
```
On System V platforms Function Multi-versioning is implemented using `GNU Indirect Functions`_ (IFUNC). The compiler generated IFUNC resolver uses a global variable ``__aarch64_cpu_features`` that contains a description of the detected CPU features to select the best function version``. Initialization of the ``__aarch64_cpu_features`` is performed by the runtime. Each FMV resolver function will call the ``__init_cpu_features_resolver`` function that must be provided by the runtime environment. This runtime function must ensure that ``__aarch64_cpu_features`` is populated.

 that selects between function versions calls a function provided by the runtime called ``__init_cpu_features_resolver`` that ensures a global variable containing a description of the detected CPU features. The ``__init_cpu_features_resolver`` function has the following prototype:
```
I think I've worded that so that the runtime could in theory populate ``__aarch64_cpu_features`` prior to IFUNC resolution and have ``__init_cpu_features_resolver`` be a no-op.

As an aside:
In theory each FMV ifunc resolver could have been self contained (`__init_cpu_features` inlined into every resolver and no caching of the result in `__aarch64_cpu_features`), which would have meant that all a runtime needs to support FMV is ifunc support.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,140 @@ IFUNC requirements for dynamic linkers
...
+the bits corresponding to the CPU features that have been detected:
+
+.. code-block:: c
+
+   uint64_t __aarch64_cpu_features
+
+The variable may contain the following fields:
+
+.. table:: CPU features detected
+
```
smithp35, line 1621: Just an observation on the implentation. I see that ``__init_cpu_features_resolver`` assumes that `__aarch64_cpu_features` has been calculated if it is non-zero.

Is there a CPU that would have none of these features set after running ``__init_cpu_features_resolver`` i.e. that function would return 0? It may be worth reserving a value that is always set so you can tell the difference between an explicit 0 and the variable is unitialized.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,140 @@ IFUNC requirements for dynamic linkers
...
+with the runtime library. However, all implementations must provide
+a DSO local definition of the function by setting the symbol
+visibility to ``STV_HIDDEN``.
+
+The runtime initialization must set a global variable which contains
+the bits corresponding to the CPU features that have been detected:
+
+.. code-block:: c
+
+   uint64_t __aarch64_cpu_features
```
smithp35: Could be worth `uint64_t __aarch64_cpu_features = 0;` as the initial value should be 0 or the runtime will think it has already been initialised.

labrinea:
ping

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,149 @@ IFUNC requirements for dynamic linkers
...
+
+On System V platforms Function Multi-versioning is implemented using
+GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
+uses a global variable ``__aarch64_cpu_features`` that contains a
+description of the detected CPU features to select the best function
+version:
+
+.. code-block:: c
+
+   uint64_t __aarch64_cpu_features = 0;
```
DanielKristofKiss, line 1616: As we are already at the bit 60 probably we could declare it as array or such to make it future proof.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,149 @@ IFUNC requirements for dynamic linkers
...
+    +-------------------+----------+
+
+A special value ``FEAT_INIT = (1U << 63);`` is used to signify
+initialization completion.
+
+Initialization of the ``__aarch64_cpu_features`` is performed by
+the runtime. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function that must be provided
+by the runtime environment. This runtime function must ensure
+that ``__aarch64_cpu_features`` is populated.
```
DanielKristofKiss: if app code would use the `__aarch64_cpu_features` we could a requirement to be sure `__init_cpu_features_resolver` is callable before main (e.g. from a __attribute__((constructor)) function )

DanielKristofKiss: maybe even require the `__aarch64_cpu_features` is initialized by the time the app reaches `main`

sysvabi64/sysvabi64.rst
labrinea: For the time being there are a few gaps in the enum we can reuse. Precisely 15 entries. Also an array would not reflect what is currently implemented in the runtime libraries used by clang and gcc. I think we want to document the current ABI?

sysvabi64/sysvabi64.rst
smithp35: That's an interesting case that I think we need to be explicit about. If we do permit app developers to use it then we'd probably want some function call to access it. Mostly so that calling the accessor function would ensure that the definition and static constructor (or alternative mechanism) were dragged in from the library.

If we reserve its use for FMV then we shouldn't need a static constructor (or maybe we use the static constructor to clear the variable so no-one else can use it!).

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,149 @@ IFUNC requirements for dynamic linkers
...
+    +-------------------+----------+
+    | FEAT_SME          | 1U << 43 |
+    +-------------------+----------+
+    | FEAT_MEMTAG2      | 1U << 45 |
+    +-------------------+----------+
+    | FEAT_SB           | 1U << 47 |
+    +-------------------+----------+
+    | FEAT_SSBS2        | 1U << 50 |
+    +-------------------+----------+
+    | FEAT_BTI          | 1U << 51 |
```
jacobbramley, line 1703: Is it really the case that there's a bit for BTI, but not for PAuth?

sysvabi64/sysvabi64.rst
jacobbramley: What would happen with late-loaded libraries (e.g. dlopen)? I'm somewhat ignorant of the FMV implementation details, but won't dlopen need access to these bits?

There might also security considerations to think about: if an attacker can influence which version of a function is loaded, they can potentially pick one without security mitigations, for example by falsely claiming that PAuth/BTI aren't present, or that DIT is. If this all happens before main, that's unlikely to matter. Could something similar to RELRO be effective here?

smithp35: As I understand it, each shared-library with ifunc resolvers from FMV will have a definition of `__aarch64_cpu_features` as it cannot know that the executable will define it. As written the definition is pre-emptible so the shared-library that has been dlopen'ed will use the existing previaling definition if it exists, otherwise it will use its own. If the definition isn't pre-emptible then each link-unit (executable, shared-object) will have its own copy.

This will all happen before main, apart from dlopen'ed shared-libraries. If the only way to initialise `__aarch64_cpu_features` were via an ifunc [1] then `__arch64_cpu_features` could be defined in a `.data.rel.ro` section to make it RELRO.

If we choose not to make `__aarch64_cpu_features` accessible to application code, there could be a case for getting rid of it and not caching it. Yes it would mean a longer start up time for programs that have lots of resolvers.

[1] This wouldn't work for static constructors as `__arch64_cpu_features` would be marked read-only before it could be written to. On platforms that support ifuncs there would always be the possibility of including an "identity" ifunc that exists just to initialise `__aarch64_cpu_features`.

labrinea: We discussed this offline with some of the stakeholders. I believe in this document we want to cover the sysv ABI, but the constructor call on start up for runtime initialization is specific to Windows. Perhaps it's better to document this elsewhere, is what @smithp35 suggested.

Now, in regards to the global variable, it was a design choice predating my involvement in the development of FMV. Do we think it is still necessary to cache the detected features in a global, or can the resolver function keep this information locally? IIUC @Wilco1 suggested that HWCAPs detection is cheap on Linux systems and that we could get rid of the global. @DanielKristofKiss perhaps the global is necessary for Windows which doesn't have IFUNCs? If not, is it too late to amend the design? 

Eitherway, this patch documents the current design. Do we want to document the current behavior?

Wilco1: Note currently the generated ifuncs refer to the global:

```
        bl      __init_cpu_features_resolver
        adrp    x0, __aarch64_cpu_features
        ldr     x0, [x0, #:lo12:__aarch64_cpu_features]
```
In principle the ifunc resolvers could directly work on HWCAPs. However the translation from HWCAP bits into FEAT bits is cheap enough that it is reasonable to compute the FEAT bits on demand rather than cache them.

To remove the global `__init_cpu_features_resolver` would need to return the feature bits. The easiest option is just to return it as `__uint128`. If we ever wanted more than 128 bits, we could add `__init_cpu_features_resolver2` for the next 128 bits.

jroelofs: The features bitset is not cheap to compute on Darwin platforms, as there we need a syscall.

smithp35: For the purposes of the ELF subset of the ABI, I think we're really describing requirements of the runtime on ELF platforms. We don't have a rtabi64.rst (yet), if we did the documentation here would likely go there. I certainly don't want to seem like an implementation must use the runtime provided functions/data. Only that an implementation is permitted to use them, and rely on their presence on the runtime library [1]

To me, because it is possible to implement FMV without the runtime, user code cannot assume the presence of the global variable so it should not be well defined to access it outside of FMV. I don't think we can stop people doing it (Hyrum's law), but if their code breaks due to a future implementation change then it is on them.

I expect Linux, Darwin, and Windows to have different technologies available to them for implementing FMV. It is possible that clang, chooses to have the same generated code for Darwin and Linux so we end up requiring the global variable even if it is not strictly necessary or beneficial for Linux. If different code-generation strategies are used it could be possible to avoid using the runtime.

What I think we need:
* Decide on whether we need make it a requirement that to support FMV, the function and variable must be provided by the runtime? I think that the answer is likely yes to support existing objects.
* Decide on whether accessing the variable, runtime function outside of FMV is supported? My vote is no, any access outside of FMV resolvers is not supported and not guaranteed to work in future releases.
* Decide if the runtime should place the global variable in RELRO on platforms that support it? This would involve specifying a section name with a `.data.rel.ro` prefix. This would prevent the variables from being modified after the ifunc resolvers have run. The possible downside would mean that an implementation of `__init_cpu_features_resolver` must use the cached value if it exists, as an attempt to write to it later (for example with a dlopen'ed shared library with FMV) would fail.

[1] I am making the assumption that we need to support existing objects that already use the runtime library and variable.

Wilco1: I don't believe we are limited by existing usage of the variable - this code is in libgcc.a so objects only get pulled in when referenced. If we place a new resolver (that may cache internally if desired) in a different object, old resolvers using the variable continue to work even if all new code uses the new resolver. Then you can remove the old resolver in a few years.

So the answers could be no, no, implementation defined (if using a variable for caching).

smithp35: I think the way we could approach the answer to "Decide on whether we need make it a requirement that to support FMV, the function and variable must be provided by the runtime?"  is that we change the sentence below
```
The compiler generated IFUNC resolver uses a global variable ``__aarch64_cpu_features`` that contains a
description of the detected CPU features to select the best function version:
```
To replace `uses` with `may use` and may call a function `__init_cpu_features_resolver` from the ifunc resolver. 
```
The compiler generated IFUNC resolver may use a global variable ``__aarch64_cpu_features`` that is initialized by a function `__init_cpu_features_resolver` that is called from the ifunc resolver function. To provide support for these objects a runtime library should implement these as follows:
```

We can also say something like:
```
Implementing FMV using ``__aarch64_cpu_features`` is not required. Accessing ``__aarch64_cpu_features`` from outside a FMV resolver function is not well defined.
```

Agreed that making ``__aarch64_cpu_features`` .data.rel.ro is implementation defined.

labrinea: > I don't believe we are limited by existing usage of the variable - this code is in libgcc.a so objects only get pulled in when referenced. If we place a new resolver (that may cache internally if desired) in a different object, old resolvers using the variable continue to work even if all new code uses the new resolver. Then you can remove the old resolver in a few years.
> So the answers could be no, no, implementation defined (if using a variable for caching).

I am a bit confused by this. In LLVM at least, the resolver function is generated by the compiler in the translation unit which contains the default function version. As you mentioned above currently the generated ifuncs refer to the global and rely on the bitfields for choosing the right version. If none of this is a requirement why are we even documenting it in the ABI?

smithp35: It is a requirement to provide the variable and function as described, because existing objects depend on the runtime library providing it.

A future implementation could do everything in the resolver function and not need the runtime library. That would still be compliant.

A future toolchain could decide to drop the runtime library, although it would then lose compatibility with old objects. I expect that won't happen as the runtime library part is small.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+Implementing FMV using ``__aarch64_cpu_features`` is not required.
+Accessing ``__aarch64_cpu_features`` from outside a FMV resolver
+function is not well defined.
+
+Initialization of the ``__aarch64_cpu_features`` is performed by
+the runtime library. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function at load time. This runtime
+function must ensure that ``__aarch64_cpu_features`` is populated.
+
+Systems without support for IFUNC may call the
```
smithp35: If compiler's are always going to use IFUNCs for a SystemV target such as `--target=aarch64-linux-gnu` or `--target=aarch64-linux-android` then a SystemV platform that doesn't support IFUNCs the dynamic loader is just going to reject the binary at dynamic load time as it won't know what to do with the IFUNC relocations.

I think we should take this part out unless we have support in SystemV platforms for an alternative strategy that doesn't use IFUNCs.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+lets the compiler generate multiple function versions and auto-dispatch
+between them. Each of the function versions is specialized for a set
+of architecture extensions. The most suitable version is selected
+dynamically. This requires runtime information about the CPU features
+available on the host. FMV is supported on GNU/Linux, Android, and
+many of the BSD operating systems.
+
+On System V platforms Function Multi-versioning is implemented using
+GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
+may rely on a runtime library for selecting the most suitable function
```
smithp35: The runtime library does not select the most suitable function, that is done by the ifunc resolver. The runtime library provides the data needed to make that choice.

Perhaps
"
The compiler generated IFUNC resolver may rely on the presence of a global variable ``__aarch64_cpu_features`` provided by the runtime library. It contains information about the available CPU features. The runtime library must also a provide a function ``__init_cpu_features_resolver`` that the IFUNC resolver can call to initialize ``__aarch64_cpu_features``. 
"
I don't think we need the `but it is not a requirement` as we've used `may`.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+version, but it is not a requirement.
+
+If a runtime library is used, it must ensure that a global variable
```
smithp35: I think we should take this part out if all SystemV platforms use IFUNCS. It is already covered by the paragraph above. For ifuncs to work ``__aarch64_cpu_features`` needs to be populated before it is used by the IFUNC resolver and that is before main and before static constructors are run.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+A special value ``FEAT_INIT = (1U << 63);`` is used to signify
+initialization completion.
+
+Implementing FMV using ``__aarch64_cpu_features`` is not required.
+Accessing ``__aarch64_cpu_features`` from outside a FMV resolver
+function is not well defined.
+
+Initialization of the ``__aarch64_cpu_features`` is performed by
+the runtime library. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function at load time. This runtime
```
smithp35: If we adopt my suggestion above then we've already covered this.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+
+The ``__init_cpu_features_resolver`` function has the following
+prototype:
+
+.. code-block:: c
+
+   void __init_cpu_features_resolver (uint64_t, const uint64_t *);
+
+The above interface expects the same parameters as a GNU Indirect
+Function resolver. Other platforms may use a different interface
```
smithp35: Can you add a reference to the section where that is defined with `GNU C Library IFUNC interface`_ ?

smithp35 (commented):
I've made a few suggestions.

If the only implementation for systemv is IFUNCs then I think it is important not to mention the static constructor as it runs too late to be used by the resolvers.

Out of interest what do we do when FMV is asked for when the target is `aarch64-linux-musl` ?

musl doesn't support ifuncs so it would reject any program that uses ifuncs at load time.

I guess our options are:
* Fail to compile when FMV is used with a musl environment.
* Always output the base variant on musl, essentially no FMV.
* Use a form of FMV that doesn't use ifuncs. I guess that has to be some kind of additional code that runs every time the function is called. In that case we would need to add the part about static construction in.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,163 @@ IFUNC requirements for dynamic linkers
...
+Initialization of the ``__aarch64_cpu_features`` is performed by
+the runtime library. Each FMV resolver function will call the
+``__init_cpu_features_resolver`` function at load time. This runtime
+function must ensure that ``__aarch64_cpu_features`` is populated.
+
+Systems without support for IFUNC may call the
+``__init_cpu_features_resolver`` function from an
+``__attribute__((constructor))`` function provided by the runtime.
+
+Systems with support for IFUNC should place the global variable
```
smithp35: Just a thought. If ``__aarch64_cpu_features`` is made RELRO then if called as a constructor then ``__init_cpu_features_resolver`` must not write to it. 

Ideally if ``__aarch64_cpu_features`` is RELRO then we shouldn't also use the constructor, and if we do then all ``__init_cpu_features_resolver`` can do is check that ``__aarch64_cpu_features`` has already been initialised and return.

labrinea: That's exactly what the code does. The constructor reads the cached value and early exits without modifying it. That said, do we still want to encourage placing the global in RELRO in this document, or shall I take this out too?

smithp35: I don't think we need to take it out. When I think about it, if an implementation tried to make the variable RELRO and wrote to it in the constructor all their programs using FMV would instantly break so warning them that it might be a problem is not really necessary. It would be a different matter if it were a subtle problem that might occur intermittently.

smithp35 (approved):
I'm happy with the text. Thanks for all the updates!

I'll hold off merging to give some time for other reviewers to comment.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,151 @@ IFUNC requirements for dynamic linkers
...
+    +-------------------+----------+
+    | FEAT_SME2         | 1U << 58 |
+    +-------------------+----------+
+    | FEAT_RCPC3        | 1U << 59 |
+    +-------------------+----------+
+    | FEAT_MOPS         | 1U << 60 |
+    +-------------------+----------+
+
+A special value ``FEAT_INIT = (1U << 63);`` is used to signify
+initialization completion.
```
Wilco1: Is this strictly needed or even useful? AArch64 mandates FEAT_FP so just checking the field is non-zero would be sufficient, saving a bit (which could be used to extend).

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,151 @@ IFUNC requirements for dynamic linkers
...
+between them. Each of the function versions is specialized for a set
+of architecture extensions. The most suitable version is selected
+at load time. This requires runtime information about the CPU features
+available on the host. FMV is supported on GNU/Linux, Android, and
+many of the BSD operating systems.
+
+On System V platforms Function Multi-versioning is implemented using
+GNU Indirect Functions (IFUNC). The compiler generated IFUNC resolver
+may rely on the presence of a global variable ``__aarch64_cpu_features``
+provided by the runtime library. It contains information about the
```
Wilco1, line 1609: It's important we state that this variable must be a hidden DSO local variable. It cannot be shared (because each DSO may use a different version of the resolver code with different bits defined - it can even have different sizes).

Wilco1 (changes requested):
x

sysvabi64/sysvabi64.rst
labrinea: I am not sure, but that's how it's been used so far in LLVM. Bit 62 indicates extension to a second 64-bit word. See https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/cpu_model/AArch64CPUFeatures.inc

sysvabi64/sysvabi64.rst
smithp35: I think it could work with a non-DSO local variable (assuming stable bit positions) and new bits added with a separate sized symbol (rather than a larger size array using the same symbol name).

However I think there would be drawbacks as the features advertised would be those of the first initializer to run, and as that is controlled by dynamic relocation order, it is likely to be a shared-library dependency, possibly even the system libc, which is likely to be older than the toolchain being used.

Wilco1: We can't use a separate symbol if the extension bits implies a larger array. As I suggested above, we could support several resolver implementations using unique names, and in that case you wouldn't need extension bits.

And it would be a major issue if you could disable security features by ensuring an old DSO gets loaded first...

smithp35: I'm happy for @labrinea to make it DSO local, it does give more scope for future changes to the resolver implementation. The benefits outweigh the restriction to the implementation.

sysvabi64/sysvabi64.rst
labrinea: Shall I document bit 62, or remove the mention of bit 63 too?

smithp35: I think it is worth putting 62 in at least as a reserved field.

I think it is your choice with bit 63. A couple of low priority reasons why we might want to keep it as is. 

I can't see any code that actively checks just bit 63, although every current initialisation function is going to write to it. Any code that could generate code that tests a repurposed bit 63 would need a matching compiler-rt/libgcc that doesn't always set it. That would make me a bit nervous to reuse it immediately, so could be worth marking it as reserved.

The sysvabi does require FP and SIMD (https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#51hardware-requirements). There have been a handful of people that have gone off grid to build their own no-fp/soft-fp platform. They may have to tweak the FMV installation to make it work, but we could say that we don't care about this case.

labrinea: I think it'd be odd to reserve bit 62 for extension and then use bit 63 as a regular detection bit. I'll document both. Unless if it's not too late we could use bit 63 for extension. Thoughts?

labrinea: I believe `FEAT_INIT` is being used implicitly when we read the cached value of `__aarch64_cpu_features` https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/cpu_model/aarch64/fmv/getauxval.inc. If nothing gets detected (as you said this isn't likely since FP/SIMD is always present) the code will early exit because `FEAT_INIT` will be set in the cached value.

ilinpv: In my opinion, replacing FEAT_INIT by FEAT_FP as features initialized flag is already not backward-compatible change so it should be fine to change FEAT_EXT bit to 63. I believe it is also worth documenting FEAT_EXT and how the extension to a second 64-bit word works.

labrinea:
After further digging we found that the `__aarch64_cpu_features` global is used by `__builtin_cpu_supports` and the runtime library for SME. These two rely on constructors which run after the resolvers have run, meaning RELRO cannot be used in conjunction with them. Shall I word all of this info with a note, or just remove the mention of RELRO entirely?

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,161 @@ IFUNC requirements for dynamic linkers
...
```
labrinea, line 1609: > a global variable ``__aarch64_cpu_features`` provided by the runtime library

I believe this isn't right, the compiler generates this variable

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,161 @@ IFUNC requirements for dynamic linkers
...
+    | FEAT_MOPS         | 1U << 60 |
+    +-------------------+----------+
+
+Implementing FMV using ``__aarch64_cpu_features`` is not required.
+Accessing this variable from outside a FMV resolver function is
+not well defined. The variable may be placed in the
+`Relocation Read Only (RELRO)`_ program segment to prevent it
+from being modified after the FMV resolvers have run. The
+variable must be defined as DSO-local with its symbol visibility
+set to ``STV_HIDDEN``.
```
labrinea: > The variable must be defined as DSO-local with its symbol visibility set to ``STV_HIDDEN``

If the compiler generates this then do we still need to document it here?

sysvabi64/sysvabi64.rst
Wilco1: How do you mean? It's defined by the runtime.

sysvabi64/sysvabi64.rst
```diff
@@ -1728,6 +1723,17 @@ from being modified after the FMV resolvers have run. The
 variable must be defined as DSO-local with its symbol visibility
 set to ``STV_HIDDEN``.
 
+.. note::
+
+   Both the SME support routines (see AAPCS64_ for more information)
+   and the compiler built-in function ``__builtin_cpu_supports`` rely
```
Wilco1: There is no requirement for SME functions to use `__aarch64_cpu_features` - in fact GCC's implementation does not. Similarly `__builtin_cpu_supports could use a different variable. So it's better to just state that the variable may be used by runtime libraries or other features besides FMV which implies it needs a constructor init.

sysvabi64/sysvabi64.rst
```diff
@@ -1728,6 +1723,17 @@ from being modified after the FMV resolvers have run. The
...
+   on the ``__aarch64_cpu_features`` variable for detecting CPU features.
+   Therefore the runtime library must ensure that the variable is
+   initialized prior to their usage. FMV support is not required for
+   using the SME support routines or the ``__builtin_cpu_supports``
+   function. The ``__aarch64_cpu_features`` variable should not be
+   placed in ``RELRO`` if no FMV resolver has run.
```
Wilco1: I think this part is redundant. Even if you used FMV and thus have an ifunc, it does not guarantee that the ifunc resolver always runs early on every system. And given other runtime uses requires a constructor init, we can't use RELRO since that is done well before constructors.

sysvabi64/sysvabi64.rst
Wilco1: Yes I believe so.

sysvabi64/sysvabi64.rst
labrinea: Sorry I got mistaken, ignore me

$ llvm-objdump -t ./release/lib/clang/21/lib/aarch64-unknown-linux-gnu/libclang_rt.builtins.a
```
libclang_rt.builtins.a(aarch64.c.o):       file format elf64-littleaarch64
SYMBOL TABLE:
0000000000000008 g     O .bss	0000000000000008 .hidden __aarch64_cpu_features
```

sysvabi64/sysvabi64.rst
labrinea: So if I understand correctly we can place ``__aarch64_cpu_features`` in RELRO as long as it is only used by the FMV resolvers. If the variable is used by the runtime library or by compiler generated code besides FMV then it cannot be in RELRO. Is that right?

Wilco1: Correct - but that choice must be made when defining the symbol, it's not something linkers can decide.

sysvabi64/sysvabi64.rst
```diff
@@ -1592,6 +1592,157 @@ IFUNC requirements for dynamic linkers
...
+    +-------------------+----------+
+    | FEAT_SME2         | 1U << 58 |
+    +-------------------+----------+
+    | FEAT_RCPC3        | 1U << 59 |
+    +-------------------+----------+
+    | FEAT_MOPS         | 1U << 60 |
+    +-------------------+----------+
+
+Implementing FMV using ``__aarch64_cpu_features`` is not required.
+Accessing this variable from outside a FMV resolver function is not
```
smithp35: The `Accessing this variable from outside a FMV resolver function is not well defined.` is in contradiction with the note below. I think we could reword to say:
```
Accessing ``__aarch64_cpu_features`` is reserved for the compiler generated code or the runtime library.
```

smithp35 (commented):
I only have one small comment since the last update. Otherwise LGTM.

PR #333, opened 2025-06-11
[AAELF64] AArch64 Veneer Types Recognized by Binary Analysis Tools.

ilinpv:
The patch introduces definitions for standard veneers on AArch64 to improve recognition by binary analysis tools.

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -1086,8 +1086,8 @@ For AArch64 the linker defined ``_GLOBAL_OFFSET_TABLE_`` symbol should
 be the address of the first global offset table entry in the ``.got``
 section.
 
-See `Sample code sequences for code models`_ for code model for
-examples of how to access the ``.got``.
+See `Sample code sequences for code models`_ for examples of how to
+access the ``.got``.
 
 Function Addresses
 ^^^^^^^^^^^^^^^^^^


aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+
+Note that some of the MOVK instructions may be omitted if their corresponding 16-bit segments of the address are zero and do not need to be explicitly set.
+
+__AArch64Abs[XO]LongThunk_ Short-Range Veneers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When the target is within range of a direct branch, the linker may generate a simple short-range branch veneer under the same name::
+
+  __AArch64Abs[XO]LongThunk_<target>:
+    B <target>
```
maksfb: Is LLD the only linker that generates this type of veneer? I don't know if the inclusion of `AbsLong` in the name was intentional.

aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A 64-bit absolute target address is loaded from a literal pool into a register, and an indirect branch is used to transfer control to the target::
+
+  <caller>:
+    B/BL __AArch64AbsLongThunk_<target>
+
+  __AArch64AbsLongThunk_<target>:
+    LDR X16, =<target>
+    BR  X16
```
maksfb: Is there an expectation that the literal pool follows the instruction sequence?

MaskRay:
Thanks! The description looks good to me.

aaelf64/aaelf64.rst
smithp35: Yes. In theory a linker could put it in a separate section within 1 MiB away, however that would require some special case code to ensure it stayed in range for not much gain. So in practice every veneer implementation that I've seen has the literal immediately afterwards.

Although not an issue with the above implementation, a veneer with an odd number of instructions may need 4-bytes of padding to ensure 8-byte alignment of the literal pool (for targets that have strict alignment enabled).

aaelf64/aaelf64.rst
smithp35: The name is a side-effect of LLDs implementation. There is a single veneer that chooses whether it is short or long depending on the distance.

I think there could be a patch to LLD that changes the name when the choice is made, although to date it's not been important enough to implement.

At the moment ld.bfd doesn't implement this shorter form of veneer. I don't know about mold.

aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+  <caller>:
+    B/BL __AArch64AbsXOLongThunk_<target>
+
+  __AArch64AbsXOLongThunk_<target>:
+    MOVZ X16, #:abs_g0_nc:<target>
+    MOVK X16, #:abs_g1_nc:<target>, LSL #16
+    MOVK X16, #:abs_g2_nc:<target>, LSL #32
+    MOVK X16, #:abs_g3:<target>,    LSL #48
+    BR   X16
+
```
smithp35: There is the ld.bfd veneer which is similar to the `__AArch64AbsLongThunk_` except it is also position independent.

Instead of loading the address directly, it loads the offset from the PC
```
LDR X16, =.Loffset_to_target
ADR X17, #0 // X17 = current PC
ADD X16, X16, X17
BR  X16
.Loffset_to_target:
.xword target - (. - 12) // R_AARCH64_PREL64 target + 12 
```

I don't know what the naming convention is for GNU ld. I think they may have a stubs section without individual names for each veneer.

aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+  <caller>:
+    B/BL __AArch64ADRPThunk_<target>
+
+  __AArch64ADRPThunk_<target>:
+    ADRP X16, <target>
+    ADD  X16, X16, :lo12:<target>
+    BR   X16
+
+Note that ``<target>`` may be an entry in the PLT.
```
smithp35: That's the case for all of the veneers described here. If there is a B or BL to the PLT entry for target and the PLT is > 128 MiB away from the B/BL then there will be a veneer generated.

Although not part of the veneer code. When there is an indirect branch to the PLT entry care is needed to add a BTI to the PLT entry if we're generating a program that is setting `AARCH64_FEATURE_1_BTI`.

aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+
+Recognized by alternative names: ``__<target>_veneer``, ``$thunk[0-9]+``, ``[0-9a-f]+-tramp-[0-9]+``
+
+__AArch64BTIThunk_ BTI Landing Pad Veneers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
smithp35: May be able to handle some of this via reference to https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#971tool-requirements-for-generating-bti-instructions

aaelf64/aaelf64.rst
ilinpv: This one is observed in LLD. At this point, in addition to LLD, I’ve gathered veneer names produced by [ld.bfd](https://github.com/bminor/binutils-gdb/blob/master/bfd/elfnn-aarch64.c#L2355), [mold](https://github.com/rui314/mold/blob/main/src/passes.cc#L1987) ( I guess @rui314 know more about mold thunks ) and [go linker](https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/arm64/asm.go;l=1341)

aaelf64/aaelf64.rst
```diff
@@ -1970,6 +1970,98 @@ Custom PLTs
...
+    B/BL __AArch64AbsXOLongThunk_<target>
+
+  __AArch64AbsXOLongThunk_<target>:
+    MOVZ X16, #:abs_g0_nc:<target>
+    MOVK X16, #:abs_g1_nc:<target>, LSL #16
+    MOVK X16, #:abs_g2_nc:<target>, LSL #32
+    MOVK X16, #:abs_g3:<target>,    LSL #48
+    BR   X16
+
+Note that some of the MOVK instructions may be omitted if their corresponding 16-bit segments of the address are zero and do not need to be explicitly set.
```
Wilco1: You can omit MOVK if the value is 0x0000 if MOVZ is used or 0xffff if MOVN is used. It's unlikely linkers optimize anything but the top 16 bits given that this veneer will only needed if the distance is over 4GB.

aaelf64/aaelf64.rst
Wilco1: Agreed - and if lazy binding is disabled linkers can shortcut the PLT by directly loading from the GOT.

This is also better than using a literal load in the .text section and needing 2 variants for each veneer...

It appears we need a bit of ABI design on linker veneers...

aaelf64/aaelf64.rst
Wilco1: It would be feasible to first emit a number of veneers and only then the data.

However it's not obvious why we need both this and the execute only variant. There are an infinite number of possible veneer sequences, but for each range we need only one. We can always emit only execute veneers.

Note given none of the code models currently support a .text size larger than 2GB, so you'd need assembler trickery with absolute symbols to force these >4GB veneers.

aaelf64/aaelf64.rst
Wilco1: If the GOT is within 4GB, doing ADRP/LDR/BR from the GOT would be a better option since it avoids placing literals in .text. And if not, it makes sense to make this veneer execute-only too and use MOVZ/MOVK for the offset.

aaelf64/aaelf64.rst
smithp35: Yes, at the stage that veneers are generated we'll know if there's a .got.plt entry for the target so we should be able to load from .plt.got when -znow is used. Off the top of my head:
```
ADRP x16, :got: target@got.plt
LDR x16:got_lo12: target@got.plt
BR x16
```

aaelf64/aaelf64.rst
smithp35: One possible limitation for some linkers is that veneers are often added quite late and the GOT may be fixed at this point. I think it should be possible to add to the GOT for ld.lld.

I would expect that for most programs if the GOT were within 4 GiB so would the destination function so the ADRP, ADD could be used.

aaelf64/aaelf64.rst
smithp35: I've occasionally seen people with embedded systems (not set up MMU yet) have a boot loader in low memory that jumps to a high address.

I suspect that if we default to using the ADRP thunk (4 GiB) universally, either using it within range and falling back to the longer sequence if out of range, or for lld at least defaulting --pic-veneer to True which will force use of the ADRP thunk.

smithp35 (commented):
There's definitely some room for improvement in thunks. Would need a small amount of implementation work in linkers.

aaelf64/aaelf64.rst
ilinpv: Seems we have such [GOT veneer](https://cs.opensource.google/go/go/+/master:src/cmd/link/internal/arm64/asm.go;l=1409) generated in go linker. However, I'm not sure what relocation is generated for that kind of veneer and can we rely on it in BOLT. Looking into the code it seems they have jump relocation referred to that veneer.

aaelf64/aaelf64.rst
rui314: I just implemented a long range extension thunk to mold. I chose this code sequence: https://github.com/rui314/mold/blob/47f6c2839c15cd0e982956c760428617ea35a0e9/src/arch-arm64.cc#L604-L614

PR #334, opened 2025-07-03
[AADWARF32] Allocate a register number for CPSR

statham-arm:
The document currently says of CPSR, along with the VFP (and FPA) control registers, "It is considered unlikely that these will be needed for producing a stack back-trace in a debugger."

However, CPSR _can_ be required for producing a correct stack backtrace. This occurs due to conditional return sequences, for example

```
function:
  PUSH   {r4, r5, r6, lr}
  SUB    sp, sp, #64
  // ... do stuff ...
  CMP    this, that       // we will return early if they are equal
  ADDEQ  sp, sp, #64
  POPEQ  {r4, r5, r6, pc}
  // ... now, if we didn't return, continue using our stack frame
```

In between ADDEQ and POPEQ, the state of the stack depends on the flags in CPSR. If the Z flag is set, then the ADDEQ has happened, and the POPEQ is about to; if Z is clear, neither one has happened. In this example the function has no frame pointer, so the CFA is defined as an offset from sp, and _what_ offset depends on whether we just added 64 to sp.

This style of conditional return has always been possible, since it depends only on the earliest features of the Arm instruction set. The accepted idiom in my experience has always been to write stack frame information that is valid for one case but not the other. Introducing a DWARF register number for CPSR makes it possible to write stack frame information that is valid in both cases. For example, you could write an expression along these lines, which uses the 4 bits at the top of CPSR to decide which bit of the constant `bitmask` to test, and then each Arm condition code is representable as a different bitmask:

```
  DW_OP_breg_13                    // fetch sp
  DW_OP_constu #bitmask            // bit mask for the particular condition
  DW_OP_bregx #CPSR                // fetch CPSR
  DW_OP_const1u #28
  DW_OP_shr                        // make (CPSR >> 28), just the NZCV bits
  DW_OP_shr                        // shift bit mask right by the NZCV value
  DW_OP_const1u #1
  DW_OP_and                        // AND with 1 to isolate low bit
  DW_OP_bra #label-offset-else     // if nonzero, branch over the next constant
  DW_OP_constu #offset2            // load one possible value to add to sp
  DW_OP_skip #label-offset-end     // and skip over the other constant
label-offset-else:
  DW_OP_constu #offset1            // load the other value to add to sp
label-offset-end:
  DW_OP_minus                      // add either offset2 or offset1 to sp
```

Of course, debuggers will take time to catch up. But a more interesting use case for being able to precisely describe stack situations like this is automatic static checkers for handwritten assembly language with handwritten call frame directives, which verify the semantics of the instructions against the call frame updates next to them. A debugger might almost never stop at the difficult location above, but a static checker will traverse it every time, and needs a way to avoid getting confused.

### DIFF
--- a/aadwarf32/aadwarf32.rst
+++ b/aadwarf32/aadwarf32.rst
@@ -435,7 +435,9 @@ LEB128 integers. Numbers 0-127 encode in 1 byte, 128-16383 in 2 bytes.
   +----------------+------------------------+-------------------------------------+
   | 133            | SPSR_SVC               | SVC-mode SPSR                       |
   +----------------+------------------------+-------------------------------------+
-  | 134–142        | None                   | Reserved for future allocation      |
+  | 134            | CPSR                   | Current CPSR, including the flags   |
+  +----------------+------------------------+-------------------------------------+
+  | 135–142        | None                   | Reserved for future allocation      |
   +----------------+------------------------+-------------------------------------+
   | 143            | RA_AUTH_CODE           | `Return Address Authentication      |
   |                |                        | Code`_                              |
@@ -520,8 +522,8 @@ LEB128 integers. Numbers 0-127 encode in 1 byte, 128-16383 in 2 bytes.
      more precisely specified scheme using 2-byte register numbers. The new
      numbering scheme should also be used for VFP-v2.
 
-The CPSR, VFP and FPA control registers are not allocated a numbering above.
-It is considered unlikely that these will be needed for producing a stack
+The VFP and FPA control registers are not allocated a numbering above. It is
+considered unlikely that these will be needed for producing a stack
 back-trace in a debugger.
 
 VFP-v3 and Neon register descriptions


walkerkd (commented):
Last comment on the example code should probably be:
 "subtract either offset2 or offset1 from sp"

smithp35:
Adding the register seems reasonable. We're not expecting to be adding large numbers of new registers in AArch32.

Although out of scope for AADWARF32, I'm guessing you will you need additional cfi directive(s) to construct the DWARF expression?

walkerkd:
Since DWARF 3 there have been the CFI instructions DW_CFA_expression, DW_CFA_val_expression and DW_CFA_def_cfa_expression which can take DWARF expression like the one mentioned in the original comment.

However the Linux .eh_frame specification currently only currently defines support for DW_CFA_expression (it uses the DWARF 2 specification as the base standard with some DWARF 3 extensions) : https://refspecs.linuxbase.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/dwarfext.html#DWARFEHENCODING

statham-arm:
I think that's fine, because I don't know of any reason why `.eh_frame` would need to use an expression based on CPSR for determining the stack layout.

It only comes up in `.debug_frame` because a debugger might be made to stop between any two instructions (especially if you're debugging handwritten assembly, which is the class of code most likely to do tricks like this conditional return sequence in the first place).

But in `.eh_frame`, when you're unwinding the stack due to a thrown exception, you'd surely only need to unwind through a function at places where an exception might be thrown, which means function calls. And an ABI-compliant function call clobbers CPSR (or at least can't be trusted not to). So if you need CPSR to know your own stack layout _before_ a function call, then on return from the call, you've forgotten it yourself, and won't even be able to clean up your own stack frame at return time.

statham-arm:
Ah, yes, I see what you mean. Should I add some text along the lines of "CPSR is only for use in expressions, when the stack layout depends on it, and you aren't allowed to write a CFI rule to say where it's stored, because it never is"?

Mind you, this is surely also very like r0–r3: those too aren't supposed to have CFI rules saying where the caller's version lives, but you can use them in CFI anyway, e.g. you could (if you really wanted to) make the CFA based on r3.

rearnsha:
It's correct that r0-r3 (and IP) can't be used in unwinding (with the AAPCS conventions); the same applies to d0-d7 and d16-d31.
Of course, you could envisage ABI variants where these might become call-saved and they can certainly all be used in expressions (eg for 'virtual' variables).  On that basis I think there's some scope for adding the CPSR.
But that raises further questions, such as should the FPSR also be made available (because FP comparisons don't directly write the CPSR)?  What about other status bits?

rearnsha:
```
        vcmpe.f32       s0, s1                   // <- debugger is here
        vmrs    APSR_nzcv, FPSCR
        addeq  sp, sp, #64
        popeq  {r4, r5, r6, pc}
```
So at the point of the comparison, only the FPSR contains the correct value.

rearnsha:
And when I reach
```
popeq  {r4, r5, r6, pc}
```
The CPSR only tells me what it's going to be in one instruction time.  But I don't need dwarf information to tell me that, I could work it out from the instruction and the current register set itself.

So what's your point exactly?
Dwarf expressions are really intended for things like
```
 for (int i = 0; i < 100; i++)
   a[i] = 3;
```
which could be compiled to (assuming `a` is in `r1`)
```
  mov r0, #3
  add r2, r1, #400  // termination value
L0:
  str r0, [r1], #4
  cmp r1, r2
  blt L0
```
now the compiler can describe `i` as `100 - ((r2 - r1) / 4)` to allow it to reconstruct the user variable after optimization.  Absent such a description the debugger can only say that it doesn't know what value `i` has.

Issue #317, opened 2025-03-31
New static relocation type: R_AARCH64_INST32


pcc:
R_AARCH64_INST32 is a static relocation type that will be used to implement [deactivation symbols](https://discourse.llvm.org/t/rfc-deactivation-symbols/85556). The proposed semantics shall be as follows:

If S is a defined symbol, write bits [31:0] of X at byte-aligned place P. Check that -2^31 <= X < 2^32. Otherwise, P is unmodified.

A proposed implementation in LLD is available [here](https://github.com/llvm/llvm-project/pull/133534).

pcc:
Also proposing that the new relocation type would always write a little-endian value regardless of the target endianness because A64 instructions are always little-endian.

smithp35:
Thanks for raising. I've only skimmed through the proposals on the LLVM Discourse at the moment, will need to read it through in more detail to make sure I understand the details; I think the https://discourse.llvm.org/t/rfc-deactivation-symbols/85556 is the best place for me to comment on the non AArch64 specific parts.

My initial reaction, will post on Discourse after reading in more detail:
* If the relocation is an instruction rather than data relocation, the little-endianness should drop out from that.
* Ideally this relocation would apply only to instructions without static relocations. Otherwise there would need to be restrictions on code-generators about ordering of relocations (currently a linker can't assume that instructions are in order of ascending r_offset, even if in practice they are). ELF also has some strange rules for composition of relocations at the same r_offset (although I've not seen any linker implement) https://www.sco.com/developers/gabi/latest/ch4.reloc.html "multiple consecutive relocations"
* To avoid name clashes with user defined symbols, for compiler generated use cases we should standardise the symbol name.
* The relocation is more powerful/flexible than it needs to be for the use case. For example if it is to be used to replace an instruction with a NOP, then at least for AArch64 we don't need to use the symbol value. It could just be symbol defined then write NOP, otherwise ignore. Just thinking that someone is bound to (ab)use this in assembly and run into trouble if they do something the tools aren't expecting.

pcc:
> If the relocation is an instruction rather than data relocation, the little-endianness should drop out from that.

Ack

[...]

PR #336, opened 2025-07-14
[AAELF64] Document R_AARCH64_FUNCINIT64

### LINKED ISSUE #329: New dynamic relocation type: R_AARCH64_FUNCINIT64
R_AARCH64_FUNCINIT64 is a dynamic relocation type for relocating word-sized data in the output file using the return value of a function. An R_AARCH64_FUNCINIT64 shall be relocated as an R_AARCH64_IRELATIVE with the target symbol address if the target symbol is non-preemptible, and it shall be a usage error to relocate an R_AARCH64_FUNCINIT64 with a preemptible or STT_GNU_IFUNC target symbol.

The initial use case for this relocation type shall be for emitting global variable field initializers for [structure protection](https://discourse.llvm.org/t/rfc-structure-protection-a-family-of-uaf-mitigation-techniques/85555). With structure protection, the relocation value computation is tied to the compiler implementation in such a way that it would not be reasonable to define a relocation type for it (for example, it may involve computing a hash using a compiler-determined algorithm), hence the need for the computation to be implemented as code in the binary.

pcc:
Fixes #329.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1818,6 +1818,8 @@ The dynamic relocations for those execution environments that support only a lim
   +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
   | 1032       | 188        | R\_<CLS>\_IRELATIVE         | Indirect(Delta(S) + A)             | See note below.                           |
   +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
+  | 1033       | \-         | R\_AARCH64\_FUNCINIT64      | Indirect(S + A)                    | See note below.                           |
+  +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
   | 1041       | \-         | R\_<CLS>\_AUTH\_RELATIVE    | SIGN(DELTA(S) + A, SCHEMA(\*P))    | See note below.                           |
   +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
   | 1042       | \-         | R\_AARCH64\_AUTH\_GLOB\_DAT | SIGN((S + A), SCHEMA(\*P))         | See note below.                           |
@@ -1852,6 +1854,8 @@ The need for copy relocations can be avoided if a compiler generates all code re
 ``R_<CLS>_RELATIVE`` represents a relative adjustment to the place based on the load address of the object relative to its original link address. All symbols defined in the same segment will have the same relative adjustment. If ``S`` is the null symbol (ELF symbol index 0) then the adjustment is based on the segment defining the place. On systems where all segments are mapped contiguously the adjustment will be the same for each relocation, thus adjustment never needs to resolve the symbol. This relocation represents an optimization; a static linker can use it to replace ``R_<CLS>_GLOB_DAT`` when the symbol is known at static link time to always resolve to the current link unit.
 
 ``R_<CLS>_IRELATIVE`` represents a dynamic selection of the place’s resolved value. The means by which this relocation is generated is platform specific, as are the conditions that must hold when resolving takes place.
+
+``R_AARCH64_FUNCINIT64`` represents a dynamic selection of the place’s resolved value, and generates a ``R_AARCH64_IRELATIVE`` relocation in the binary. The operand to ``R_AARCH64_IRELATIVE`` is an addend relative to the binary load address, so the symbol operand to ``R_AARCH64_FUNCINIT64`` must have a statically known address.
 
 Relocations ``R_AARCH64_TLS_DTPREL``, ``R_AARCH64_TLS_DTPMOD`` and ``R_AARCH64_TLS_TPREL`` were previously documented as ``R_AARCH64_TLS_DTPREL64``, ``R_AARCH64_TLS_DTPMOD64`` and ``R_AARCH64_TLS_TPREL64`` respectively.  The old names can be supported if needed for backwards compatibility.
 


aaelf64/aaelf64.rst
```diff
@@ -1818,6 +1818,8 @@ The dynamic relocations for those execution environments that support only a lim
   +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
   | 1032       | 188        | R\_<CLS>\_IRELATIVE         | Indirect(Delta(S) + A)             | See note below.                           |
   +------------+------------+-----------------------------+------------------------------------+-------------------------------------------+
+  | 1033       | \-         | R\_AARCH64\_FUNCINIT64      | Indirect(S + A)                    | See note below.                           |
```
smithp35, line 1821: The ABI would consider R_AARCH64_FUNCINIT64 a static data relocation as it is handled by a static linker. The R_AARCH64_IRELATIVE relocation that the static linker produces is the dynamic relocation.

I think we'd likely want to make a new "operator" rather than use Indirect(S+A), something like `FuncInit(S)` that we add to https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5733relocation-operations that would contain text similar to what you've put in the See note below. 

I think we've got a couple of alternatives for where to put the relocation. The first choice is https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#575static-data-relocations . The second is to create a new table like relocations for the pauth-abi-extension  
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5712relocations-for-pauth-abi-extension something like "Relocations for the structure protection extension" which could also include `R_AARCH64_PATCHINST`. 

There are a couple of benefits to separating out into a "Relocations for the structure protection extension", the first is that other linker's can know when to expect to see this relocation. For example if my toolchain/platform doesn't support structure protection extension I don't need to implement these relocations. The second is that we can easily mark the extension as ALPHA which gives us more freedom to make potentially breaking changes.

The downside is that R_AARCH64_PATCHINST is more generally useful outside the extension, however I think it could be moved out to the general instruction relocations if/when binutils implements it.

aaelf64/aaelf64.rst
```diff
@@ -1853,6 +1855,8 @@ The need for copy relocations can be avoided if a compiler generates all code re
 
 ``R_<CLS>_IRELATIVE`` represents a dynamic selection of the place’s resolved value. The means by which this relocation is generated is platform specific, as are the conditions that must hold when resolving takes place.
 
+``R_AARCH64_FUNCINIT64`` represents a dynamic selection of the place’s resolved value, and generates a ``R_AARCH64_IRELATIVE`` relocation in the binary. The operand to ``R_AARCH64_IRELATIVE`` is an addend relative to the binary load address, so the symbol operand to ``R_AARCH64_FUNCINIT64`` must have a statically known address.
```
smithp35, line 1858: As mentioned in the previous comment I think it would be better to use a new operator which we could some or all of this text to.

From the LLD implementation the `R_AARCH64_FUNCINIT64` target symbol must not have STB_LOCAL or STB_GNU_INDIRECT binding in addition to having a statically known address.

aaelf64/aaelf64.rst
pcc: > The ABI would consider R_AARCH64_FUNCINIT64 a static data relocation as it is handled by a static linker. The R_AARCH64_IRELATIVE relocation that the static linker produces is the dynamic relocation.

That makes sense. I was imagining that a toolchain could theoretically support FUNCINIT64 with a symbol operand in .rela.dyn and in that case it could also act as a dynamic relocation. This is not very different from e.g. ABS64 which can act as both, but ABS64 is in the static relocation range so we could have FUNCINIT64 be there as well.

> I think we'd likely want to make a new "operator" rather than use Indirect(S+A), something like FuncInit(S) that we add to https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5733relocation-operations that would contain text similar to what you've put in the See note below.

The definition of `Indirect` is:

> ``Indirect(expr)`` represents the result of calling expr as a function. The result is the return value from the function that is returned in ``r0``. The arguments passed to the function are defined by the platform ABI.

So `FuncInit` would have the same semantics as `Indirect` but the argument must be a static link-time constant modulo load bias. All existing uses of `Indirect` already meet the criteria (issues raised in #337 aside, which are likely theoretical), so maybe we could change `Indirect` to require this.

> I think we've got a couple of alternatives for where to put the relocation. The first choice is https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#575static-data-relocations . The second is to create a new table like relocations for the pauth-abi-extension
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#5712relocations-for-pauth-abi-extension something like "Relocations for the structure protection extension" which could also include R_AARCH64_PATCHINST.
>
> There are a couple of benefits to separating out into a "Relocations for the structure protection extension", the first is that other linker's can know when to expect to see this relocation. For example if my toolchain/platform doesn't support structure protection extension I don't need to implement these relocations. The second is that we can easily mark the extension as ALPHA which gives us more freedom to make potentially breaking changes.

That's fine with me. The extensions here are relatively simple so I think they only deserve a new section in this file and not a full blown spec like memtagabielf64.

If you would like to upload a PR tomorrow as mentioned in #317 we can make it cover both and let's continue the discussion there. Otherwise I'll update this one.

> The downside is that R_AARCH64_PATCHINST is more generally useful outside the extension, however I think it could be moved out to the general instruction relocations if/when binutils implements it.

That makes sense.

As a side note, one thing that I've never really liked about the separate tables is that it is easy to accidentally allocate a conflicting relocation type number by adding 1 to the number at the end of one of the tables without checking the others. I needed to be careful to make sure that my new relocation type numbers did not conflict with any of the PAuth ones. Eventually we will close the gap between the last main table relocation and the first PAuth relocation, so leaving gaps is just delaying the problem. I probably would prefer this document to be organized as a large table with all the relocations in numerical order, possibly with each entry annotated to indicate which extension they belong to.

aaelf64/aaelf64.rst
pcc: > From the LLD implementation the R_AARCH64_FUNCINIT64 target symbol must not have STB_LOCAL or STB_GNU_INDIRECT binding in addition to having a statically known address.

With the LLD implementation there are other factors such as whether -Bsymbolic and -shared are passed, which control whether the symbol has a statically known address (modulo load bias). It may be best to simply state the restriction, which implies that it can't be global (in some circumstances) or an ifunc.

PR #337, opened 2025-07-14
[AAELF64] Simplify expressions used for relative relocations

pcc:
The function Delta(S) returns the current binary's load bias (for the null symbol as well as for locally-defined S), and is only used in the definition of RELATIVE family relocations, which are not expected to have a non-null symbol operand.

Theoretically, a non-null symbol operand could be used to refer to a symbol in another binary, which would then cause Delta(S) to evaluate to the load bias of that binary. But I am unaware of any dynamic loader implementing this (checked glibc/musl/bionic/FreeBSD/NetBSD/OpenBSD), and its utility seems limited.

The description of R_<CLS>_RELATIVE appears to contemplate a different case in which different segments in the same binary have different load biases. It is unclear to me how this would work in practice. If the idea is to accommodate individual segments being loaded at independent addresses, I don't think this would be enough; ELF is not generally designed to accommodate this so numerous other changes to relocation processing would need to be made in order for this to work.

Therefore, simplify the definition of Delta as well as the users, and bring it in line with existing practice, by removing the argument and having it always produce the current binary's load bias, and adjust the R_<CLS>_RELATIVE description to match.


smithp35:
I expect that the symbol relative case is a transplant from the 32-bit ABI, in particular the BPABI https://github.com/ARM-software/abi-aa/blob/main/bpabi32/bpabi32.rst

There was a pseudo ELF target for Symbian DLLs (Symbian had an ELF to E32 conversion tool) and the SBREL position independent  which was defined for an OS that did not have a MMU, although I don't think that this got used.

In open-source and open-source linkers, particularly for the sysvabi64 then I think the simplification can be made. I'd like to check to see if I can find any information on proprietary toolchains that may have built of the ABI before taking things out. 

An alternative could be to find a way of preserving the original content, but moving it out of line, perhaps to an appendix.

pcc:
Interesting. I had thought that this might have had something to do with a no-MMU mode inherited from 32-bit.

Another possibility could be to say that the behavior is unspecified if S is non-null. But I would prefer to just remove it if we can't easily find any evidence of proprietary toolchains implementing it. We can always make another change if we later discover a toolchain implementing it.

pauthabielf64/pauthabielf64.rst
```diff
@@ -744,7 +744,7 @@ relocation.
   +====================+==============================+====================================+
   | 0x244 (580)        | R\_AARCH64\_AUTH\_ABS64      | SIGN(S + A, SCHEMA(\*P))           |
   +--------------------+------------------------------+------------------------------------+
-  | 0x411 (1041)       | R\_AARCH64\_AUTH\_RELATIVE   | SIGN(DELTA(S) + A, SCHEMA(\*P))    |
+  | 0x411 (1041)       | R\_AARCH64\_AUTH\_RELATIVE   | SIGN(DELTA + A, SCHEMA(\*P))       |
```
tom-cosgrove-arm: On line 1821 of aaelf64/aaelf64.rst (above) `DELTA(S)` is rewritten as `Delta`, going for Title Case. Is that not wanted here too?

smithp35: Thanks for spotting.

I think that's probably my fault as I added those relocations from PAuthABI where I mistakenly used DELTA rather than Delta. This also affects the PAuthABI document.

I'll submit a PR to standardise all the DELTA instances to Delta.

smithp35: https://github.com/ARM-software/abi-aa/pull/372 to change `DELTA` to `Delta` across the repository.

smithp35 (approved):
Change LGTM, sorry for the delay, but it will need to be rebased prior to merging. @pcc are you happy to do that? If not then I can make a new PR with the changes.

pauthabielf64/pauthabielf64.rst
```diff
@@ -1270,7 +1270,7 @@ Relocation Operation
   +==============+=============================+=================================+===================================================================+
   | 0x244 (580)  | R\_AARCH64\_AUTH\_ABS64     | SIGN((S + A), SCHEMA(\*P))      | SIGN((LDG(S) + A), SCHEMA(\*P))                                   |
   +--------------+-----------------------------+---------------------------------+-------------------------------------------------------------------+
-  | 0x413 (1043) | R\_AARCH64\_AUTH\_RELATIVE  | SIGN(Delta(S) + A, SCHEMA(\*P)) | SIGN((LDG(Delta(S) + A + ADDEND(\*P)) - ADDEND(\*P), SCHEMA(\*P)) |
+  | 0x413 (1043) | R\_AARCH64\_AUTH\_RELATIVE  | SIGN(Delta + A, SCHEMA(\*P))    | SIGN((LDG(Delta + A + ADDEND(\*P)) - ADDEND(\*P), SCHEMA(\*P))    |
```
pcc, line 1273: I noticed that this relocation number (and the one below) are wrong, that can probably be fixed in a followup.

pcc (commented):
> Change LGTM, sorry for the delay, but it will need to be rebased prior to merging. @pcc are you happy to do that? If not then I can make a new PR with the changes.

Done

pauthabielf64/pauthabielf64.rst
smithp35: Thank you. I'll fix that with a follow up patch.

smithp35: https://github.com/ARM-software/abi-aa/pull/374

PR #340, opened 2025-07-16
[AAELF64] Document structure protection relocations

smithp35:
The structure protection extension [1], introduces two new relocations:

R_AARCH64_PATCHINST which replaces an instruction with the result of the relocation if the referenced symbol is defined. The relocation is intended to implement deactivation symbols [2], which can be used to "undo" structure protection when it isn't safe to apply.

R_AARCH64_FUNCINIT64 generates a R_AARCH64_IRELATIVE relocation at the place. This is similar to R_AARCH64_ABS64 referencing a symbol of binding STB_GNU_UNIQUE, with the main difference being that R_AARCH64_FUNCINIT64 references non STB_GNU_UNIQUE symbols. The relocation is used to initialize function pointers in static structs at load time by calling the equivalent of an ifunc resolver function that returns the signed address.

[1] https://discourse.llvm.org/t/rfc-structure-protection-a-family-of-uaf-mitigation-techniques/85555
[2] https://discourse.llvm.org/t/rfc-deactivation-symbols/85556

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1824,8 +1824,8 @@ relocations`_, `Program Linkage Table (PLT) Sequences and Usage
 Models`_ and `Relocation optimization`_ may ignore
 ``R_AARCH64_PATCHINST``.
 
-The intended use case for ``R_AARCH64_PATCHINST`` to be used to
-replace an instruction with a ``NOP``. Uses of the relocation for
+The intended use case for ``R_AARCH64_PATCHINST`` is to replace
+an instruction with a ``NOP``. Uses of the relocation for
 other instructions is limited to what can be constructed with ``S +
 A``. Responsibility for using ``R_AARCH64_PATCHINST`` outside of the
 Structure Protection Extension is out of scope of the ABI.
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1792,7 +1792,7 @@ relocations. The Structure Protection Extension is described in
 
 The structure protection relocations use the following additionl operator:
 
-- ``FUNCINIT(S)`` The place is relocated at run-time with a ``R_AARCH64_IRELATIVE`` relocation with no referenced symbol and the value of S in the addend field.
+- ``FUNCINIT(S + A)`` The place is relocated at run-time with a ``R_AARCH64_IRELATIVE`` relocation with no referenced symbol and the value of S + A in the addend field.
 
 .. class:: structure-protection-instruction-relocations
 
@@ -1824,6 +1824,12 @@ relocations`_, `Program Linkage Table (PLT) Sequences and Usage
 Models`_ and `Relocation optimization`_ may ignore
 ``R_AARCH64_PATCHINST``.
 
+The intended use case for ``R_AARCH64_PATCHINST`` to be used to
+replace an instruction with a ``NOP``. Uses of the relocation for
+other instructions is limited to what can be constructed with ``S +
+A``. Responsibility for using ``R_AARCH64_PATCHINST`` outside of the
+Structure Protection Extension is out of scope of the ABI.
+
 .. class:: structure-protection-data-relocations
 
 .. table:: Structure Protection Data Relocations
@@ -1831,12 +1837,13 @@ Models`_ and `Relocation optimization`_ may ignore
   +------------+------------+----------------------------------------+--------------------------------------+----------------------+
   | ELF64 Code | ELF32 Code | Name                                   | Operation                            | Comment              |
   +============+============+========================================+======================================+======================+
-  |    317     | \-         | R\_AARCH64\_FUNCINIT64                 | FUNCINIT(S)                          | See below            |
+  |    317     | \-         | R\_AARCH64\_FUNCINIT64                 | FUNCINIT(S + A)                      | See below            |
   +------------+------------+----------------------------------------+--------------------------------------+----------------------+
 
 The ``R_AARCH64_FUNCINIT64`` referenced symbol must be a function that
-does not have ``STB_GNU_INDIRECT`` or ``STB_LOCAL`` binding. The
-referenced symbol must have an address that is known at static link time.
+does not have a type of ``STT_GNU_IFUNC``. The referenced symbol must
+be non-pre-emptible and have an address that is known at static link
+time.
 
 Dynamic relocations
 ^^^^^^^^^^^^^^^^^^^
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -18,6 +18,7 @@
 .. _SYM-VER: http://www.akkadia.org/drepper/symbol-versioning
 .. _TLSDESC: http://www.fsfla.org/~lxoliva/writeups/TLS/paper-lk2006.pdf
 .. _MTEEXTENSIONS: https://www.kernel.org/doc/html/latest/arm64/memory-tagging-extension.html#core-dump-support
+.. _STRUCTPROT: https://discourse.llvm.org/t/rfc-structure-protection-a-family-of-uaf-mitigation-techniques/85555
 .. _SYSVABI64: https://github.com/ARM-software/abi-aa/releases
 .. _VFABI64: https://github.com/ARM-software/abi-aa/releases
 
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,61 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
...
+``R_AARCH64_PATCHINST``.
+
+.. class:: structure-protection-data-relocations
+
+.. table:: Structure Protection Data Relocations
+
+  +------------+------------+----------------------------------------+--------------------------------------+----------------------+
+  | ELF64 Code | ELF32 Code | Name                                   | Operation                            | Comment              |
+  +============+============+========================================+======================================+======================+
+  |    317     | \-         | R\_AARCH64\_FUNCINIT64                 | FUNCINIT(S)                          | See below            |
```
pcc: Should this not be `FUNCINIT(S + A)`?

aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,61 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
...
+  +------------+------------+----------------------------------------+--------------------------------------+----------------------+
+
+The ``R_AARCH64_FUNCINIT64`` referenced symbol must be a function that
+does not have ``STB_GNU_INDIRECT`` or ``STB_LOCAL`` binding. The
```
pcc: Was this meant to be `STB_GLOBAL` instead of `STB_LOCAL`?

Do we even need to mention `STB_GLOBAL` given that it's implied by the following sentence?

IFUNC is a symbol type, not a binding. So I suppose it should say `STT_GNU_IFUNC` type.

aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,61 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
...
+``r_offset``. The ``R_AARCH64_PATCHINST`` relocation always starts a
+new relocation composition sequence.
+
+The requirements for a static linker that supports
+``R_AARCH64_PATCHINST`` are limited to resolving the relocation. All
+other static linker processing of relocations such as `Call and jump
+relocations`_, `Program Linkage Table (PLT) Sequences and Usage
+Models`_ and `Relocation optimization`_ may ignore
+``R_AARCH64_PATCHINST``.
+
```
pcc, line 1826: We talked about only permitting specific target instructions to begin with such as `NOP`, is that worth specifying here?

aaelf64/aaelf64.rst
smithp35: On reflection it can be S + A, I'll change it, although I expect the vast majority of cases to have A = 0 with S being the address of the function.

We have problems with GOT generating relocations and a non-zero addend, but as this is not GOT generating there isn't a good enough reason not to put it in.

aaelf64/aaelf64.rst
smithp35: I took the LOCAL binding part from the test case in https://github.com/llvm/llvm-project/pull/133531/files#diff-a57816b6863597bb5e48db73d609f1b23739cddda435dd3d69a6809aa55c33eeR7
```
.rodata
# ERR: relocation R_AARCH64_FUNCINIT64 cannot be used against local symbol
.8byte func@FUNCINIT
```
I think the error message has mislead me here. It is because with `-z text` the location can't have a dynamic relocation. Unfortunately the error message says something different.

I'll rewrite. My mistake with the STB_GNU_INDIRECT. That was my recent memory of how IFUNCs were specified, I looked it up, and I got mislead by search auto correcting it to STB_GNU_UNIQUE and my eyes missed it. A case of seeing what I wanted to see, not what was actually there.

aaelf64/aaelf64.rst
smithp35: I'll revisit the discussion to see if I can capture it.

aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,68 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
...
   +------------+------------+----------------------------------------+--------------------------------------+----------------------+
 
+Relocations for Structure Protection Extension
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Structure Protection Extension defines a number of static
+relocations. The Structure Protection Extension is described in
+[STRUCTPROT_]. The Structure Protection Extension is in Alpha state.
+
+The structure protection relocations use the following additionl operator:
```
pcc: nit: additional

pcc:
@smithp35 I hope you enjoyed your vacation. Maybe it's time to land this? This would unblock several of my LLVM patches.

smithp35:
> @smithp35 I hope you enjoyed your vacation. Maybe it's time to land this? This would unblock several of my LLVM patches.

Apologies, just got through my email backlog yesterday. I've fixed the typo. I'll land this by the end of the day unless I get any objections.

aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,68 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
   | 597        | \-         | R\_AARCH64\_AUTH\_TLSDESC\_ADD\_LO12   | G(ENCD(GTLSDESC(S)))                 | See `PAUTHABIELF64`_ |
   +------------+------------+----------------------------------------+--------------------------------------+----------------------+
 
+Relocations for Structure Protection Extension
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Structure Protection Extension defines a number of static
+relocations. The Structure Protection Extension is described in
+[STRUCTPROT_]. The Structure Protection Extension is in Alpha state.
```
stuij, line 1791: Wouldn't the more specific Deactivation Symbols RFC be more relevant to these relocations?

aaelf64/aaelf64.rst
```diff
@@ -1778,6 +1783,68 @@ The ``PAUTH`` and ``ENCD`` operators are defined in `PAUTHABIELF64`_.
...
+new relocation composition sequence.
+
+The requirements for a static linker that supports
+``R_AARCH64_PATCHINST`` are limited to resolving the relocation. All
+other static linker processing of relocations such as `Call and jump
+relocations`_, `Program Linkage Table (PLT) Sequences and Usage
+Models`_ and `Relocation optimization`_ may ignore
+``R_AARCH64_PATCHINST``.
+
+The intended use case for ``R_AARCH64_PATCHINST`` to be used to
```
stuij: Sorry for the last-minute fly-by nit, but I think this sentence is missing an `is`:

the intended <something> is to..

I think we can also remove `to be used to` so we'd go from:

`The intended use case for ``R_AARCH64_PATCHINST`` to be used to replace...` -> `The intended use case for ``R_AARCH64_PATCHINST`` is to replace...`.

smithp35: Thanks for the spot. I'll upload a new patch.

aaelf64/aaelf64.rst
smithp35: There is one relocation for structure protection and one for deactivation symbols. I chose to concentrate on Structure Protection as that is the higher level description, and it includes a link to the deactivation symbols in it.

I could split the two apart a bit more and include links to both, but as these are LLVM discourse posts rather than formal descriptions I thought it best to stick to one.

Let me know if you want me to add some more context here? I could write a brief summary of structure protection, but could do that in a follow up patch.

smithp35 (commented):
Thanks for comments. I've fixed the sentence. Decided to leave the reference at structure protection for now. I suggest I try and improve that in a follow up patch if necessary.

I've noticed the CI is currently bust. I think we may need to add roman to the rst2pdf requirements.txt . I'll attempt a separate PR with that change to see if that makes a difference.

aaelf64/aaelf64.rst
stuij: ok, fair enough.

stuij (approved):
LGTM

Issue #288, opened 2024-09-30
Medium code model PIC restrictions seem misleading


mmalcomson:
The table `Code Models` in the section https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#7code-models of the sysvabi64 seems to say that there are no restrictions on the GOT for `PIC` code.

The entry for the `large` model in this column mentions that the max distance between the GOT and text must be <4GiB.

From what I could see in the text there doesn't seem to be any mention of the `medium` code model having some special handling of a GOT being further than 4GiB away from the code access to it.

Is there a reason that I'm missing that this limitation needed for the `large` model is not needed for the `medium` model?
Or is the "no additional GOT restrictions" mention overlooking that 4GiB limit on distance between text and GOT?

smithp35:
This looks like an oversight to me. My understanding is that the restriction "max distance from text to GOT < 4 GiB" is that the linker generated PLT entries use ADRP so the `.got.plt` needs to be within 4 GiB of the `.got`. This will remain true for the medium code-model. I'd also expect the .text sections to access the `.got` via ADRP too, further strengthening the case for putting the same restriction on the medium code model.

PR #345, opened 2025-08-15
[sysvabi64] Code Models. Limit range to GOT to 2GiB

### LINKED ISSUE #343: Can the small code model assume the GOT is within 2 GiB of text?
https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#code-models explicitly states that for the large code model, the max distance from text to GOT has to be under 4 GiB. #288 clarifies that this should also apply to the medium code model. For the small code model, is it reasonable to assume a distance of 2 GiB or less?

I'm asking because the `R_AARCH64_GOTPCREL32` relocation introduced in #223 appears to implicitly make this assumption. I want to use the same assumption to encode Clang's indirect references to the personality function and typeinfo objects inside `eh_frame` sections as `sdata4` instead of `sdata8` (see my [LLVM Discourse post](https://discourse.llvm.org/t/using-gotpcrel-relocations-for-personality-and-typeinfo-eh-frame-references/87820) for more details if interested), and I'm wondering how safe that is. Note 2 of https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#code-models states:

> The definition of the text segment includes the shareable PLT, code and read-only data sections.

The GOT isn't technically read-only since relocations need to be applied to it, but if you're using RELRO (which I imagine is the case more often than not) then it's functionally read-only, so it seems useful to make the max text segment size of 2 GiB apply to the GOT as well, at least for the small code model. Are there problems I'm not foreseeing with this?

smithp35:
The R_AARCH64_GOTPCREL32 relocation, currently used in clang when relative vtables are enabled, has a signed offset so is limited to 2 GiB range. This limits the range of GOT from the vtables to 2GiB.

Document this restriction in the tables for the small and medium code model. Add a note to explain the reason, so that if code generation does not use the relocation the distance to the GOT can be increased.

Fixes #343

### DIFF
--- a/sysvabi64/sysvabi64.rst
+++ b/sysvabi64/sysvabi64.rst
@@ -701,7 +701,7 @@ along with the assumptions that the code model may make.
 .. table:: Code Models
 
   +--------+----------------+----------------+------------------------+
-  | Code   | Max text       | Max combined   | Additional GOT         |
+  | Code   | Max text       | Max combined   | Additional GOT size    |
   | Model  | segment size   | span of text   | restrictions           |
   |        |                | and data       |                        |
   |        |                | segments       |                        |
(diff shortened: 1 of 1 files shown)


sysvabi64/sysvabi64.rst
```diff
@@ -750,23 +757,26 @@ along with the assumptions that the code model may make.
   5. The text segment maximum size is limited to 2GiB by
   R_AARCH64_PLT32 relocations from ``.eh_frame`` sections.
 
-  6. While designing the code models it was estimated that only 2.6%
+  6. The distance to the GOT in the small and medium code model is
+  limited to 2GiB by the R_AARCH64_GOTPCREL32 relocations.
```
Wilco1: Basically in all cases (including large model) the max distance from text to GOT is 2GB if R_AARCH64_GOTPCREL32 is used. It is 4GB otherwise as the basic range of ADRP.

So I think it doesn't really need to be added to the table since there is no difference between the code models.

smithp35: Main reason I did that was the large code model (need to expand the text of the diff to see) says
```
| large  | 2GiB           | no restriction | max distance from text |
|        |                |                | to GOT < 4 GiB         |
```
If I take the table entries out, I'll need to take that out too. Will update the PR.

smeenai:
Thank you! Taking out the table entries also addresses #288, right?

Issue #327, opened 2025-04-29
DW_CFA_AARCH64_negate_ra_state_with_pc is fundamentally flawed


smithp35:
The `DW_CFA_AARCH64_negate_ra_state_with_pc` https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst#44call-frame-instructions

A dwarf 'program' reads linearly through a file from a start location and needs to recreate the frame status that the CPU would see if it were to execute to that point.  Most importantly, it does not follow the flow-graph of the program that the CPU executes.

The following program shows how control flow could lead to an incorrect location being saved.
```
  .cfi_startproc 
func:
   // Frame setup
   .cfi_save ...
   .cfi_negate_ra_with_pc  // Save auth state and remember location
   some code
   CBZ L0
   some more code
   // Restore frame
   .cfi_negate_ra_with_pc  // Restore state (using previously stored location)
   .cfi_restore ...
   RET
L0:
   .cfi_save ...
   .cfi_negate_ra_with_pc  // ERROR - this saves the wrong location
   Some additional code
   .cfi_negate_ra_with_pc     // Uses wrong location.
   RET
...  
```
It is not sufficient to define that the first `.cfi_negate_ra_with_pc` does the save as this may not be the first directive executed by the control flow, and we do not want to restrict the control flow of a program so that this directive can work.

A proposal will be made for a replacement for `.cfi_negate_ra_with_pc` and this directive will be deprecated.

PR #346, opened 2025-08-21
[aadwarf64]] Add DW_CFA_AARCH64_set_ra_state stack frame operation

walkerkd:
The DW_CFA_AARCH64_set_ra_state operation updates the RA_SIGN_STATE pseudo register with the current signing state. If the signing state includes signing is DW_AARCH64_RA_SIGNED_SP_PC, then it also provides the offset to the signing instruction so that the PC value used in the signing can be calculated.

The DW_CFA_AARCH64_negate_ra_state_with_pc operation has been marked as deprecated.  This is because it has been found that it is not suitable for describing all cases where the PC was used to sign the return address (see https://github.com/ARM-software/abi-aa/issues/327)

The contents of the RA_SIGN_STATE pseudo register is also changed from being described in terms of a set of bits to being a series of defined values.

Previously the state of the RA_SIGN_STATE pseudo register was changed implicitly by the DW_CFA_AARCH64_negate_ra_state and DW_CFA_AARCH64_set_ra_state operations. This meant that the actual encoding was actually internal to any implementation.

Now with the introduction of the DW_CFA_AARCH64_set_ra_state operation the encoding has been made externally visible.  So the opportunity has been taken now to change the encoding to a simpler form.

### DIFF
--- a/aadwarf64/aadwarf64.rst
+++ b/aadwarf64/aadwarf64.rst
@@ -522,7 +522,7 @@ integers.
    .. _Note 9:
 
    9. Normally, the program counter is restored from the return address, however
-      having both LR and PC diversifiers are useful for describing asynchronously
+      having both LR and PC diversifiers is useful for describing asynchronously
       created stack frames. A DWARF expression may use this register to restore
       the context in case of a signal context.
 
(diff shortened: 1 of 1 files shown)


aadwarf64/aadwarf64.rst
```diff
@@ -234,6 +246,14 @@ changes to the content of the document for that release.
   |        |                             | support for unwinding with             |
   |        |                             | FEAT_PAuth_LR enabled.                 |
   +--------+-----------------------------+----------------------------------------+
+  |        |                             | - Added DW_CFA_AARCH64_set_ra_state at |
+  |        |                             |   **Alpha** quality.                   |
+  |        |                             | - Deprecated                           |
+  |        |                             |   DW_CFA_AARCH64_negate_ra_state.      |
```
bgergely0: I believe this should be `DW_CFA_AARCH64_negate_ra_state_with_pc`.

walkerkd: Thanks for spotting that.  I'll fix it.

aadwarf64/aadwarf64.rst
```diff
@@ -478,33 +498,31 @@ integers.
...
+      | ``DW_AARCH64_RA_SIGNED_SP_PC`` | 2      | Return address signed with SP+PC                                |
+      +--------------------------------+--------+-----------------------------------------------------------------+
+      |                                | 3 - 15 | (reserved for future use)                                       |
+      +--------------------------------+--------+-----------------------------------------------------------------+
 
    .. _Note 9:
 
    9. Normally, the program counter is restored from the return address, however
-      having both LR and PC columns is useful for describing asynchronously
+      having both LR and PC diversifiers are useful for describing asynchronously
```
rearnsha: I think in this context, 'having both ... _is_ useful' is the correct form. 'having both' here creates a singular subject for the sentence.

aadwarf64/aadwarf64.rst
```diff
@@ -600,25 +618,37 @@ Call frame instructions
...
+   | ``DW_CFA_AARCH64_set_ra_state`` (**Alpha**)                 | 0           | ``0x2B``   | ULEB128 ra_state | SLEB128 offset |
+   +-------------------------------------------------------------+-------------+------------+------------------+----------------+
+
+The ``DW_CFA_AARCH64_negate_ra_state`` operation toggles between the
+``DW_AARCH64_RA_NOT_SIGNED`` and ``DW_AARCH64_RA_SIGNED_SP`` return address
+states in RA_SIGN_STATE pseudo-register. It does not take any operands.
+
+The ``DW_CFA_AARCH64_set_ra_state`` instruction takes two operands; an unsigned
+LEB128 value representing a return address state ra_state and a signed LEB128
+factored offset. The required action is to set the RA_SIGN_STATE pseudo-register
```
rearnsha: I would use '...takes two operands: an unsigned ...; and a signed ...' to

aadwarf64/aadwarf64.rst
```diff
@@ -600,25 +618,37 @@ Call frame instructions
...
+to the ra_state value and if the ra_state value is ``DW_AARCH64_RA_SIGNED_SP_PC``
+to capture the current code location + (offset * ``code_alignment_factor``)
+as the signing/authenticating PAC instruction, otherwise it is has the value 0.
+The code location information can be used for authenticating the return address.
+
+The ``DW_CFA_AARCH64_negate_ra_state_with_pc`` operation toggles between the
```
rearnsha: Shouldn't this re-emphasize that this is deprecated (and perhaps explain why)?

aadwarf64/aadwarf64.rst
walkerkd: Done

aadwarf64/aadwarf64.rst
walkerkd: Done.  The current style is what is used in the base DWARF specification.  But the suggested change does make it clearer what are the 2 operands.

aadwarf64/aadwarf64.rst
walkerkd: I have added a statement to re-emphasize that this instruction is deprecate and a short explanation in non-normative text (the base DWARF specification does this by using italics for non-normative text so I have used the same convention here).

Stylie777 (commented):
From a Compiler Development POV this change makes sense, and addresses the issue of the CFI collecting an incorrect PC value.

Have you got an example that demonstrates the change in Assembly that will be generated? Would be helpful for me so I can visualise the changes being made.

PR #351, opened 2025-09-15
[atomicsabi64] Document release fence using ISHST+ISHLD barriers

Wilco1:
This allows more reordering than ISH.

### DIFF
--- a/atomicsabi64/atomicsabi64.rst
+++ b/atomicsabi64/atomicsabi64.rst
@@ -411,6 +411,9 @@ Synchronization Fences
   | ``atomic_thread_fence(seq_cst)``                    |    DMB ISH                           |
   +-----------------------------------------------------+--------------------------------------+
 
+The release fence has two alternative implementations.  Using ``DMB ISHLD`` and ``DMB ISHST``
+allows for more reordering since the combination is not a store-load barrier.
+
 32-bit types
 ------------
 


atomicsabi64/atomicsabi64.rst
```diff
@@ -399,10 +399,17 @@ Synchronization Fences
   +-----------------------------------------------------+--------------------------------------+
   | ``atomic_thread_fence(release)``                    | .. code-block:: none                 |
   |                                                     |                                      |
-  | ``atomic_thread_fence(acq_rel)``                    |    DMB ISH                           |
+  |                                                     |    DMB ISHLD                         |
```
smithp35, line 402: To check my understanding, a combination of `DMB ISHLD` and `DMB ISHST` is a `load-load`, `load-store` and `store-store` barrier, but non `store-load`; but `DMB ish` is also a `store-load` barrier?

If so could be worth putting that or an explanation of why in the description/commit-message.

smithp35, line 402: Although probably obvious, could be worth mentioning that if there are no associated features, then multiple boxes means multiple choices.

Wilco1: I've added an explanation

smithp35 (approved):
LGTM. Thanks for the update.

Issue #328, opened 2025-05-06
Add condition to prevent partial application of paired AArch64 linker instruction sequence optimizations to same symbol.


smithp35:
In https://github.com/llvm/llvm-project/issues/138254 a case has emerged where there are two ADRP/LDR accesses to a symbol within a function. One of these sequences can be optimised, the other cannot. The function also has a branch destination between the optimisable ADRP/LDR. This can result in the unoptimised ADRP being combined with the optimised LDR (ADD in this case) leading to incorrect results.

The assembly (derived from C code compiled with GCC 9.3)
```
foo:
.LFB1:
        .cfi_startproc
        cmp     x0, 0
        bge     .L8
        adrp    x2, :got:b // Optimised to adrp x2, b
.L9:
        ldr     x2, [x2, :got_lo12:b] // Optimised to adr x2, [x2 :lo12: b]
        add     x0, x2, x1
        b       bar
        .p2align 2,,3
.L8:
        stp     x29, x30, [sp, -32]!
        .cfi_def_cfa_offset 32
        .cfi_offset 29, -32
        .cfi_offset 30, -24
        adrp    x2, :got:b // Can't be optimised due to add that is in way.
        add     x0, x0, 1 // although in theory could be out of range of b.
        ldr     x3, [x2, :got_lo12:b]
        add     x0, x3, x0
        mov     x29, sp
        stp     x2, x1, [sp, 16]
        bl      bar
        ldp     x2, x1, [sp, 16]
        ldp     x29, x30, [sp], 32
        .cfi_restore 30
        .cfi_restore 29
        .cfi_def_cfa_offset 0
        b .L9
```

At the moment, according to our existing condition list [1] the linker is permitted to make this transformation. We need to find some words to ensure that the linker only apply the optimisation for `b` if all of the ADRP/LDR GOT relative accesses to `b` can be optimised.

[1] https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#579relocation-optimization

Wilco1:
Fixed

PR #352, opened 2025-09-15
[aaelf64] Clarify relocation optimization [issue #328]

Wilco1:
Make it clearer that a GOT sequence can only be optimized if all GOT relocations to the symbol are part of a sequence.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1427,9 +1427,8 @@ Linkers may optionally optimize instructions affected by relocation. Relocation 
   - The relocations apply to consecutive instructions in the order specified.
   - The relocations use the same symbol.
   - The relocated instructions have the same source and destination register.
-  - The relocations do not appear separately or in a different order.
 
-  In this case each set of relocations is independent and may be optimized. The following sequences are defined:
+  The following sequences are defined:
 
   - Large GOT indirection
 
@@ -1452,8 +1451,9 @@ Linkers may optionally optimize instructions affected by relocation. Relocation 
     - ``symbol`` does not have a ``st_shndx`` of ``SHN_ABS`` or the output is not required to be position independent.
     - ``symbol`` is within range of the ``R_<CLS>_ADR_PREL_PG_HI21`` relocation.
     - The addend of both relocations is zero.
+    - All ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` relocations to ``symbol`` are part of a sequence.
 
-    The optimized sequence does not require a GOT entry. A linker may avoid creating a GOT entry if no other GOT relocations exist for the symbol.
+    If the GOT relocations are used outside of a valid sequence, the optimization is not legal (for that symbol). The optimized sequence does not require a GOT entry. A linker may avoid creating a GOT entry if no other GOT relocations exist for the symbol.
 
   - PC-relative addressing
 
@@ -1468,6 +1468,8 @@ Linkers may optionally optimize instructions affected by relocation. Relocation 
 
       NOP
       ADR   x0, symbol                  // R_<CLS>_ADR_PREL_LO21
+
+    Each sequence is independent and may be optimized if in range.
 
 Proxy-generating relocations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^


smithp35:
Can I check what you mean by `All GOT relocations`? To give an example

```
// Relocations in sequence, for the sake of the example; out of range
ADRP  x0, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
LDR   x0, [x0 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC
...
// Relocations in sequence, in range
ADRP  x1, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
LDR   x1, [x1 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC
...
// Instructions not in sequence, although relocations could be consecutive
ADRP  x2, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
NOP
LDR   x2, [x1 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC
```

I think you mean that in each of these three pairs of GOT relocations to `symbol` then all must be optimised or none. This would avoid the problem observed in https://github.com/llvm/llvm-project/issues/138254 which had the following example:
```
foo:
        cmp     x0, 0
        bge     .L8
        adrp    x2, :got:b // Optimised to adrp x2, b
.L9:
        ldr     x2, [x2, :got_lo12:b] // Optimised to adr x2, [x2 :lo12: b]
        add     x0, x2, x1
        b       bar
        .p2align 2,,3
.L8:
        stp     x29, x30, [sp, -32]!
        adrp    x2, :got:b // Can't be optimised due to add that is in way.
        add     x0, x0, 1 // although in theory could be out of range of b.
        ldr     x3, [x2, :got_lo12:b]
        add     x0, x3, x0
        mov     x29, sp
        stp     x2, x1, [sp, 16]
        bl      bar
        ldp     x2, x1, [sp, 16]
        ldp     x29, x30, [sp], 32
        b .L9 // Jump to after optimised ADRP expecting the ldr, but getting adr.
```

Wilco1:
No - as described, all must be part of a sequence in order to be optimizable since that is proof they are independent. However not all need to be optimized (eg. if out of range). So in your example only the 3rd ADRP is not a valid sequence and thus blocks the optimization for the other 2.

An alternative would be to optimize all or nothing without considering pairs. But then a single ADRP that ends up out of range would block all optimizations, making it unsuitable for the medium/large model. Also keeping the ADRP/LDR as pairs results in better code quality.

aaelf64/aaelf64.rst
```diff
@@ -1427,9 +1427,8 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
   - The relocations apply to consecutive instructions in the order specified.
   - The relocations use the same symbol.
   - The relocated instructions have the same source and destination register.
-  - The relocations do not appear separately or in a different order.
 
-  In this case each set of relocations is independent and may be optimized. The following sequences are defined:
+  The following sequences are defined:
```
smithp35, line 1431: Line 1425 above (can't put the comment on the line). I suggest we insert a "base" before conditions
"if all the following base conditions are true."

aaelf64/aaelf64.rst
```diff
@@ -1452,8 +1451,9 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
     - ``symbol`` does not have a ``st_shndx`` of ``SHN_ABS`` or the output is not required to be position independent.
     - ``symbol`` is within range of the ``R_<CLS>_ADR_PREL_PG_HI21`` relocation.
     - The addend of both relocations is zero.
+    - All ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` relocations to ``symbol`` are part of a sequence.
```
smithp35, line 1454: I think this can be combined with the line below and removed from the specific conditions. For example

If the ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` relocations to ``symbol`` are used outside a sequence that satisfies the base conditions then the Large GOT indirection optimization is not legal for ``symbol``.

Examples that prevent the Large GOT indirection optimization for ``symbol``:

::

      // Different destination register
      ADRP  x0, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
      LDR   x1, [x0 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC

      // Instructions not in sequence.
      ADRP  x0, :got: symbol            // R_<CLS>_ADR_GOT_PAGE
      NOP
      LDR   x0, [x0 :got_lo12: symbol]  // R_<CLS>_LD64_GOT_LO12_NC

A linker may avoid creating a GOT entry if no other GOT relocations exist for the symbol.

Wilco1: I've further improved the description similar in a new pull request, see https://github.com/ARM-software/abi-aa/pull/362/files

PR #356, opened 2025-12-07
[aaelf64-morello] Drop pointless GD/IE-to-LE relaxation

jrtc27:
These LE "relaxations" do not change the instruction sequence at all
compared with IE, they just change where the 16 bytes are stored, moving
them from the GOT to somewhere in a read-only data section. This doesn't
achieve anything and just complicates linker implementations to do, when
instead they can just do nothing and let the existing implementation
recognise that the GOT entries are constants.

Thus, remove these relaxations, and note that there is no need for a
dynamic relocation in the case where you would normally relax to LE.

### DIFF
--- a/aaelf64-morello/aaelf64-morello.rst
+++ b/aaelf64-morello/aaelf64-morello.rst
@@ -1057,7 +1057,7 @@ and offset of the TLS variable is stored in a GOT slot (first 8 bytes contains t
 offset and the second 8 bytes the size). This GOT slot is initialized by a
 ``R_MORELLO_TPREL128`` dynamic relocation. The access must use the
 ``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` and ``R_MORELLO_TLSIE_ADD_LO12``
-relocations in order to allow relaxation to Local Exec. There are no other
+relocations. There are no other
 requirements on how this is performed or the registers used. A possible
 instruction sequence could be:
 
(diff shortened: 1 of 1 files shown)


jrtc27:
C.f. https://git.morello-project.org/morello/llvm-project/-/merge_requests/337

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -1057,7 +1057,7 @@ and offset of the TLS variable is stored in a GOT slot (first 8 bytes contains t
 offset and the second 8 bytes the size). This GOT slot is initialized by a
 ``R_MORELLO_TPREL128`` dynamic relocation. The access must use the
 ``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` and ``R_MORELLO_TLSIE_ADD_LO12``
-relocations in order to allow relaxation to Local Exec. There are no other
+relocations. There are no other
```
smithp35, line 1060: Could get rid of the full sentence, or possibly change `must use` to `uses`
> The access must use the
``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` and ``R_MORELLO_TLSIE_ADD_LO12``
relocations.

Not got a strong opinion as I don't have any other candidates that could be used instead.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -1073,37 +1073,13 @@ instruction sequence could be:
...
-  not valid assembler directives.
-
-The relaxation is performed by:
-
-- changing the ``R_MORELLO_TLSIE_ADR_GOTTPREL_PAGE20`` relocation on the symbol
-  ``sym`` to a ``R_MORELLO_ADR_PREL_PG_HI20`` with the symbol ``_sym_data``
-
-- changing the ``R_MORELLO_TLSIE_ADD_LO12`` relocation on symbol the ``sym`` to
-  a ``R_AARCH64_ADD_ABS_LO12_NC`` relocation with the symbol ``_sym_data``.
+No such relaxation is specified.
```
smithp35, line 1076: Resolving the ``R_MORELLO_TPREL128`` at static link time could be considered a relaxation. 

Could delete lines 1076 to 1078. 

I guess if the IE sequence had been defined with sufficient NOPS (with an identifying null relocation) it may have been possible to replace the IE sequence with the LE MOVK/MOVZ. That's out of scope of this change though.

aaelf64-morello/aaelf64-morello.rst
jrtc27: Prescriptive is probably better than descriptive for a specification though?

aaelf64-morello/aaelf64-morello.rst
jrtc27: I wouldn't count it as a linker relaxation, otherwise "not emitting R_AARCH64_ABS64 for PDEs" would also be a relaxation. And yes, the lack of NOPs is the main reason no such relaxation is possible, there's just not enough space to cram in materialising the size for SCBNDS (and you'd likely need more structure on the instruction sequence than currently required).

PR #362, opened 2026-01-20
[aaelf64] Improve relocation optimizations [issue #328]

Wilco1:
Improve relocation optimizations to make the conditions clearer. Explain more explicitly explain that the Large GOT indirection optimization is all or nothing.

### DIFF
--- a/aaelf64/aaelf64.rst
+++ b/aaelf64/aaelf64.rst
@@ -1452,7 +1452,8 @@ Linkers may optionally optimize instructions affected by relocation. Relocation 
     The Large GOT indirection optimization is valid for ``symbol`` if all
     ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` relocations are
     part of a sequence which satisfies all conditions above. If so, any
-    sequences where ``symbol`` is within +-4GiB range may now be optimized.
+    sequences where ``symbol`` is within the range of the ``R_<CLS>_ADR_PREL_PG_HI21``
+    relocation applied to the location of the ``ADRP`` may now be optimized.
     A linker may avoid creating a GOT entry if no other GOT relocations exist
     for the symbol.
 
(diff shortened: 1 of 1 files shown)


aaelf64/aaelf64.rst
```diff
@@ -1449,19 +1441,34 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
...
     - ``symbol`` is not of type ``STT_GNU_IFUNC``.
-    - ``symbol`` does not have a ``st_shndx`` of ``SHN_ABS`` or the output is not required to be position independent.
-    - ``symbol`` is within range of the ``R_<CLS>_ADR_PREL_PG_HI21`` relocation.
-    - The addend of both relocations is zero.
+    - ``symbol`` does not have a ``st_shndx`` of ``SHN_ABS``.
+
+    The Large GOT indirection optimization is valid for ``symbol`` if all
+    ``R_<CLS>_ADR_GOT_PAGE`` and ``R_<CLS>_LD64_GOT_LO12_NC`` relocations are
+    part of a sequence which satisfies all conditions above. If so, any
+    sequences where ``symbol`` is within +-4GiB range may now be optimized.
```
smithp35: I think it could be worth using the range of the replacement relocation to define the range. That would be more precise than +-4GiB. Perhaps
```
If so, any sequences where ``symbol`` is within the range of R_<CLS>_ADR_PREL_PG_HI21 applied to the location of the ADRP, may now be optimized.

File: aaelf64/aaelf64.rst [hunk 2]
```diff
@@ -1473,6 +1480,10 @@ Linkers may optionally optimize instructions affected by relocation. Relocation
       NOP
       ADR   x0, symbol                  // R_<CLS>_ADR_PREL_LO21
 
+    The optimization is valid if ``symbol`` is within +-1MiB range, the
```
smithp35: Similar to the comment above about using +-1MiB.

The optimization is valid if ``symbol`` is within range of of the `R_<CLS>_ADR_PREL_LO21` relocation applied to the location of the ADD instruction, the instructions are consecutive ...

aaelf64/aaelf64.rst
Wilco1: OK I've added this

smithp35 (approved):
Thanks for the update. LGTM.

PR #365, opened 2026-01-25
[aaelf64-morello][aapcs64-morello] Document pure-capability benchmark ABI

jrtc27:
This has existed for several years but was never specified, so attempt
to do so in key places. Omissions are likely but this should give enough
detail that anyone caring about this niche within a niche can figure out
any missed implications themselves.

### DIFF
--- a/aaelf64-morello/aaelf64-morello.rst
+++ b/aaelf64-morello/aaelf64-morello.rst
@@ -306,6 +306,24 @@ A Morello toolchain can emit ELF Note sections in accordance to [CHERI_ELF_].
 The following Morello-specific ELF Note types are used, allocated from the
 space reserved by [CHERI_ELF_] for processor-specific use:
 
+.. _Morello-specific note types:
+
+.. class:: aaelf64-morello-note-types
+
+.. table:: Morello-specific note types
+
+  +------------+----------------------------------------+-----------------------------------------------------------+
+  | Value      | Name                                   | Description                                               |
+  +------------+----------------------------------------+-----------------------------------------------------------+
+  | 0x80000000 | NT_CHERI_MORELLO_PURECAP_BENCHMARK_ABI | Whether the object uses the pure-capability benchmark ABI |
+  +------------+----------------------------------------+-----------------------------------------------------------+
+
+.. note::
+
+  NT_CHERI_MORELLO_PURECAP_BENCHMARK_ABI has a Desc Size of 4, and Desc should
+  be interpreted as a 4-byte boolean value, with values other than 0 and 1
+  reserved.
+
 .. _Morello-specific NT_CHERI_TLS_ABI types:
 
 .. class:: aaelf64-morello-NT_CHERI_TLS_ABI-types
@@ -346,8 +364,10 @@ expected definition.
 The type of any other symbol defined in an executable section can be
 ``STT_NOTYPE``. A linker is only required to provide long-branch and PLT support
 for symbols of type ``STT_FUNC``. A linker is also only required to provide
-interworking support for A64 and C64 symbols of type ``STT_FUNC`` (interworking
-for untyped symbols must be encoded directly in the object file)
+interworking support for A64 and C64 symbols of type ``STT_FUNC``, and only if
+not using the pure-capability benchmark ABI (interworking for untyped symbols
+or the pure-capability benchmark ABI must be encoded directly in the object
+file).
 
 Symbol names
 ^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


aaelf64-morello/aaelf64-morello.rst
```diff
@@ -306,6 +306,24 @@ A Morello toolchain can emit ELF Note sections in accordance to [CHERI_ELF_].
...
+  | Value      | Name                                   | Description                                               |
+  +------------+----------------------------------------+-----------------------------------------------------------+
+  | 0x80000000 | NT_CHERI_MORELLO_PURECAP_BENCHMARK_ABI | Whether the object uses the pure-capability benchmark ABI |
+  +------------+----------------------------------------+-----------------------------------------------------------+
+
+.. note::
+
+  NT_CHERI_MORELLO_PURECAP_BENCHMARK_ABI has a Desc Size of 4, and Desc should
+  be interpreted as a 4-byte boolean value, with values other than 0 and 1
+  reserved.
```
smithp35, line 325: Some things that may be worth mentioning:
* Does the absence of the note imply a value of 0?
* Do all input objects have to have the same value?
* Is the linker expected to propagate a consolidated note into the executable/shared-object like .note.gnu.property or can it be discarded?

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -451,14 +471,19 @@ apply to symbols of type ``STT_FUNC`` and ``STT_GNU_IFUNC``:
...
+  bit 0 set) if not using the pure-capability benchmark ABI, otherwise it is
+  the same as for a symbol addressing an A64 instruction.
 
 .. note::
   This allows a linker to distinguish A64 and C64 code symbols without having
   to refer to the map. An A64 symbol will always have an even value, while a
   C64 symbol will always have an odd value. However, a linker should strip the
   discriminating bit from the value before using it for relocation.
 
+  Due to the pure-capability benchmark ABI using integer BR/BLR for indirect
```
smithp35, line 483: Following https://github.com/ARM-software/abi-aa/blob/main/aapcs64-morello/aapcs64-morello.rst#611general-purpose-registers could this be something like:
```
Due to the pure-capability benchmark ABI using a 64-bit context for indirect calls using BL/BLR, bit 0 ...
```

I don't have a particularly strong opinion here. If integer vs capability is a well used term in the CHERI/Morello community then leave it as is.

aaelf64-morello/aaelf64-morello.rst
```diff
@@ -451,14 +471,19 @@ apply to symbols of type ``STT_FUNC`` and ``STT_GNU_IFUNC``:
 
 - If the symbol addresses a C64 instruction, its value is the address of the
   instruction with bit 0 set (in a relocatable object, the section offset with
-  bit 0 set).
+  bit 0 set) if not using the pure-capability benchmark ABI, otherwise it is
```
smithp35, line 474: I think it would be worth moving the ABI distinction further up in the paragraph, or perhaps using a full stop before otherwise.

A suggestion:
```
If the symbol addresses a C64 instruction, its value depends on the ABI used. If not using the pure-capability benchmark ABI the value is the address of the instruction with bit 0 set (in a relocatable object, the section offset with
Comment view  bit 0 set). If using the pure-capability benchmark ABI the value is the same as for a symbol addressing an A64 instruction.
```

aaelf64-morello/aaelf64-morello.rst
jrtc27: For the first two, yes, that should be documented here (and the answer to both is yes). For the last one, that's not really something that's up to this spec to document, that's a CHERI ELF gABI point (to which the answer is, yes, they get propagated as they're all about annotating binaries with the details ABI variant is in use, normally incompatible if different with the compat TLS ABI as a rare exception).

aaelf64-morello/aaelf64-morello.rst
jrtc27: I think that makes it less clear, and to me "in a 64-bit context" in the linked description means "in the context of an operand that is to be interpreted as a 64-bit integer", i.e. the context is the instruction and operand number, not that the instruction is within such a context. And yes, integer (or (integer) address) vs capability is a normal distinction to be making within CHERI. I can change it to be a bit less terse though with something like "Due to the pure-capability benchmark ABI using BR/BLR with a 64-bit integer X register for ..." if you think that would be clearer?

aaelf64-morello/aaelf64-morello.rst
jrtc27: Sure, I can try to rework this bullet point a bit more, I was trying to be minimally disruptive of the existing wording but you're right that affected the clarity here (and found this one the hardest to write at the time)

aaelf64-morello/aaelf64-morello.rst
smithp35: I think your suggested replacement looks good to me.
> "Due to the pure-capability benchmark ABI using BR/BLR with a 64-bit integer X register for ..."

aaelf64-morello/aaelf64-morello.rst
smithp35: Thanks. I agree that isn't an easy edit to make.

PR #370, opened 2026-03-02
aapcs64: fix broken link

### LINKED ISSUE #347: Broken link in section 6.6 of aapcs64
There are 3 links in section 6.6 of aapcs64/aapcs64.rst rendered as

> [every time](https://github.com/ARM-software/abi-aa/blob/main/aapcs64/eachnormalreturn)

that, when clicked, lead to a 404 not found error. Possibly it's just a rendering issue.

yury-khrustalev:
Fixes #347

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -1286,7 +1286,7 @@ X will be the return address supplied by S1's caller.  The act of
 resuming execution at X is then a normal return from S2 to S1 and a
 normal return from S1 to S1's caller.
 
-.. _`each normal return`:
+.. _`every time`:
 
 A subroutine call might return normally more than once.  For example,
 the C subroutine ``setjmp`` can return normally twice: once to complete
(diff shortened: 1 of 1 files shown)


aapcs64/aapcs64.rst
```diff
@@ -1646,8 +1646,8 @@ conditions is true:
     entry to S.
 
 More generally, a call to a subroutine S is said to “preserve ZA” if
-the call preserves ZA `every time <each normal return>`_ that the call
-returns normally.  A call trivially satisfies this requirement if the
+the call preserves ZA every time that the call
+`returns normally`_.  A call trivially satisfies this requirement if the
```
smithp35: There is an invisible anchor on line 1289 that this is supposed to target, which is close to `returns normally`
```
.. _`each normal return`:
```

I've been trying to get that to work, but have been failing [1]

However as that anchor is invisible, we can just rename it:
```
.. _`every time`:
```
Then in each of the references we can just use
```
`every time`_
```

I've checked that works with rstcheck and rst2pdf and doesn't need any wording or location changes.


[1]
I expect the rendering problems are caused by:
```
`every time <each normal return>`_
```
Which seems to interpret `each normal return` as a hyperlink.

According to Sphinx https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

we should be using 
```
:ref:`every time <each normal return>`
```
But I can't get our rst2pdf or rstcheck to understand :ref:

yury-khrustalev: Thanks, I've updated patch as recommended.

Issue #266, opened 2024-06-17
[aapcs64] Unclear callee/caller wording in aapcs64.rst


a74nh:
In aapcs64.rst

> z0-z7 are used to pass scalable vector arguments to a subroutine, and to return scalable vector results from a function. If a subroutine takes at least one argument in scalable vector registers or scalable predicate registers, or if it is a function that returns results in such registers, **it must ensure that** the entire contents of z8-z23 are preserved across the call. In other cases it need only preserve the low 64 bits of z8-z15, as described in [SIMD and Floating-Point registers](https://github.com/ARM-software/abi-aa/blob/2982a9f3b512a5bfdc9e3fea5d3b298f9165c36b/aapcs64/aapcs64.rst#simd-and-floating-point-registers).

> p0-p3 are used to pass scalable predicate arguments to a subroutine and to return scalable predicate results from a function. If a subroutine takes at least one argument in scalable vector registers or scalable predicate registers, or if it is a function that returns results in such registers, **it must ensure that** p4-p15 are preserved across the call. In other cases it need not preserve any scalable predicate register contents.

In both cases in `it must ensure that` it is not clear whether `it` refers to the caller or the callee.

Eg: if `it` is the callee then the wording should be `the subroutine must ensure that`.

This wording caused issues when designing SVE support for [.NET](https://github.com/dotnet/runtime).

kunalspathak:
.NET issue that describes SVE support: https://github.com/dotnet/runtime/issues/93095

smithp35:
I agree that the sentence can be clarified.

Assuming the confusion hasn't been resolved already, there's a couple of other parts that may help parsing the text:
https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#22terms-and-abbreviations
```
Routine, subroutine
A fragment of program to which control can be transferred that, on completing its task, returns control to its caller at an instruction following the call. Routine is used for clarity where there are nested calls: a routine is the caller and a subroutine is the callee.
```

https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#3scope
```
Obligations on the called routine to preserve the program state of the caller across the call.
```

Combining this with the original text, the `it` is referring to the `callee`.

a74nh:
> Combining this with the original text, the `it` is referring to the `callee`.

Thanks! That was the conclusion we came to after reading around the issue elsewhere. But it would be nice for it to be clearer.

[...]

PR #371, opened 2026-03-02
[aapcs64] Standardize on "caller-saved" and "callee-saved"

AlfieRichardsArm:
This addresses some issues raised in #266 by standardizing on
"callee-saved" and "caller-saved" for describing register preservation
rules in PCS.

For now this only applies to AAPCS64.

### DIFF
--- a/aapcs64-morello/aapcs64-morello.rst
+++ b/aapcs64-morello/aapcs64-morello.rst
@@ -187,17 +187,19 @@ changes to the content of the document for that release.
 
 .. table::
 
-    +----------+------------------------------+----------------------------------------+
-    | Issue    | Date                         | Change                                 |
-    +==========+==============================+========================================+
-    | 00alpha  | 1st October 2020             | Alpha release.                         |
-    +----------+------------------------------+----------------------------------------+
-    | 2020Q4   | 21\ :sup:`st` December 2020  | Document released on Github.           |
-    +----------+------------------------------+----------------------------------------+
-    | 2022Q1   | 1\ :sup:`st` April 2022      | Fix up rule C.8 on capability passing. |
-    +----------+------------------------------+----------------------------------------+
-    | 2022Q3   | 20\ :sup:`th` October 2022   | Rework varargs for Morello.            |
-    +----------+------------------------------+----------------------------------------+
+    +----------+------------------------------+-----------------------------------------------------+
+    | Issue    | Date                         | Change                                              |
+    +==========+==============================+=====================================================+
+    | 00alpha  | 1st October 2020             | Alpha release.                                      |
+    +----------+------------------------------+-----------------------------------------------------+
+    | 2020Q4   | 21\ :sup:`st` December 2020  | Document released on Github.                        |
+    +----------+------------------------------+-----------------------------------------------------+
+    | 2022Q1   | 1\ :sup:`st` April 2022      | Fix up rule C.8 on capability passing.............. |
+    +----------+------------------------------+-----------------------------------------------------+
+    | 2022Q3   | 20\ :sup:`th` October 2022   | Rework varargs for Morello.                         |
+    +----------+------------------------------+-----------------------------------------------------+
+    | 2025Q4   | 3\ :sup:`rd` March 2026      | - Standardize on "caller-saved" and "callee-saved". |
+    +----------+------------------------------+-----------------------------------------------------+
 
 
 References
@@ -361,31 +363,31 @@ Additionally, a stack-pointer register, SP in a 64-bit context or CSP in a capab
 
 .. table:: General purpose registers and AAPCS64-cap usage
 
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | Register   | Special  | Role in AAPCS64-cap                                                                                |
-    +============+==========+====================================================================================================+
-    | r31        | CSP      | The Capability Stack Pointer.                                                                      |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r30        | CLR      | The Capability Link Register.                                                                      |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r29        | CFP      | The Capability Frame Pointer.                                                                      |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r19-r28    |          | Registers r19-r28 (c19-c28) are callee-saved.                                                      |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r18        |          | The Platform Register, if needed; otherwise a temporary register. See notes.                       |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r17        | CIP1     | The second intra-procedure-call temporary register (can be used by call veneers and PLT code).     |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r16        | CIP0     | The first intra-procedure-call scratch register (can be used by call veneers and PLT code).        |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r10-r15    |          | Temporary registers.                                                                               |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r9         |          | Parameter register for variadic calls, temporary register otherwise.                               |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r8         |          | The capability indirect result location register.                                                  |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
-    | r0-r7      |          | Parameter/result registers.                                                                        |
-    +------------+----------+----------------------------------------------------------------------------------------------------+
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | Register   | Special  | Role in AAPCS64-cap                                                                            |
+    +============+==========+================================================================================================+
+    | r31        | CSP      | The Capability Stack Pointer.                                                                  |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r30        | CLR      | The Capability Link Register.                                                                  |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r29        | CFP      | The Capability Frame Pointer.                                                                  |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r19-r28    |          | Registers r19-r28 (c19-c28) are callee-saved.                                                  |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r18        |          | The Platform Register, if needed; otherwise a caller-saved register. See notes.                |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r17        | CIP1     | The second intra-procedure-call temporary register (can be used by call veneers and PLT code). |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r16        | CIP0     | The first intra-procedure-call temporary register (can be used by call veneers and PLT code).  |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r10-r15    |          | Caller-saved registers.                                                                        |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r9         |          | Parameter register for variadic calls, caller-saved register otherwise.                        |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r8         |          | The capability indirect result location register (caller-saved).                               |
+    +------------+----------+------------------------------------------------------------------------------------------------+
+    | r0-r7      |          | Parameter/result registers (caller-saved).                                                     |
+    +------------+----------+------------------------------------------------------------------------------------------------+
 
 .. _General purpose registers and AAPCS64 usage:
 
@@ -393,29 +395,29 @@ Additionally, a stack-pointer register, SP in a 64-bit context or CSP in a capab
 
 .. table:: General purpose registers and AAPCS64 usage
 
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | Register  | Special  | Role in AAPCS64                                                                                    |
-    +===========+==========+====================================================================================================+
-    | r31       | SP       | The Stack Pointer.                                                                                 |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r30       | LR       | The Link Register.                                                                                 |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r29       | FP       | The Frame Pointer.                                                                                 |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r19-r28   |          | The lower 64 bits of the registers (x19-x28) is callee-saved.                                      |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r18       |          | The Platform Register, if needed; otherwise a temporary register. See notes.                       |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r17       | IP1      | The second intra-procedure-call temporary register (can be used by call veneers and PLT code).     |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r16       | IP0      | The first intra-procedure-call scratch register (can be used by call veneers and PLT code).        |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r9-r15    |          | Temporary registers.                                                                               |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r8        |          | The indirect result location register.                                                             |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
-    | r0-r7     |          | Parameter/result registers.                                                                        |
-    +-----------+----------+----------------------------------------------------------------------------------------------------+
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | Register  | Special  | Role in AAPCS64                                                                                |
+    +===========+==========+================================================================================================+
+    | r31       | SP       | The Stack Pointer.                                                                             |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r30       | LR       | The Link Register.                                                                             |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r29       | FP       | The Frame Pointer.                                                                             |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r19-r28   |          | The lower 64 bits of the registers (x19-x28) is callee-saved.                                  |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r18       |          | The Platform Register, if needed; otherwise a caller-saved register. See notes.                |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r17       | IP1      | The second intra-procedure-call temporary register (can be used by call veneers and PLT code). |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r16       | IP0      | The first intra-procedure-call temporary register (can be used by call veneers and PLT code).  |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r9-r15    |          | Caller-saved registers.                                                                        |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r8        |          | The indirect result location register (caller-saved).                                          |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
+    | r0-r7     |          | Parameter/result registers (caller-saved).                                                     |
+    +-----------+----------+------------------------------------------------------------------------------------------------+
 
 
 The first eight registers, r0-r7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).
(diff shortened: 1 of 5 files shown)


aapcs64/aapcs64.rst
```diff
@@ -427,6 +427,11 @@ Scratch register, _`temporary register`, caller-saved register
 
 Callee-saved register
    A register whose value must be preserved over a function call. If the function being called (the callee) needs to use the register, then it is responsible for saving and restoring the old value.
+   Also known as "callee preserved".
```
smithp35: There's quite a large change to aapcs64. Can we add an entry to the `aapcs64-change-history`

aapcs64/aapcs64.rst
```diff
@@ -1317,17 +1320,16 @@ SME implementation, ZA can handle twice as much data as the vector registers,
 and so on.
 
 Suppose that a subroutine S1 with live data in ZA calls a subroutine S2 that
-has no knowledge of S1.  If the AAPCS64 defined ZA to be “call-preserved”
-(“callee-saved”), S1 would need to save and restore ZA around S1's own
-use of ZA, in case S1's caller also had live data in ZA.  If the AAPCS64
-defined ZA to be “call-clobbered” (“caller-saved”), S1 would need to
-save and restore ZA around the call to S2, in case S2 also used ZA.
-However, nested uses of ZA are expected to be rare, so these saves and
+has no knowledge of S1.  If the AAPCS64 defined ZA to be “callee-saved”, S1
```
smithp35: Although this was in the original text. I think it should be S2 (callee) that would be doing the saving and restoring.
```
If the AAPCS64 defined ZA to be "callee-saved", S2 would need to save and restore ZA around S2's own use of ZA, in case S1's caller also had live data in ZA.
```

aapcs64/aapcs64.rst
smithp35: Not in your patch, just noticed a typo:
"to refer to".

Assuming I'm right about the S1/S2 mixup it may be worth a separate patch to fix typos first.

aapcs64/aapcs64.rst
```diff
@@ -2496,9 +2498,9 @@ by PSTATE.ZA.
 * The subroutine has a `streaming-compatible interface`_ with
   the following properties:
 
-  * X1-X15, X19-X29 and SP are call-preserved.
-  * Z0-Z31 are call-preserved.
-  * P0-P15 are call-preserved.
+  * X1-X15, X19-X29 and SP are callee-saved.
+  * Z0-Z31 are caleelsaved-saved.
```
smithp35: Typo: callee-saved

aapcs64/aapcs64.rst
```diff
@@ -2496,9 +2498,9 @@ by PSTATE.ZA.
...
+  * P0-P15 are eecsaved-saved.
```
smithp35: Typo: callee-saved

vfabia64/vfabia64.rst
```diff
@@ -433,12 +433,12 @@ can be found on `developer.arm.com
 <https://developer.arm.com/docs/100986/latest/>`_.
 
 .. note:: The SVE-specific rules of the AAPCS are in beta version. The
-          list of SVE call-clobbered and call-preserved registers in
+          list of SVE callee-saved and caller-saved registers in
```
smithp35, line 436: Can we update the Change history as there's an entry that refers to call-preserved and call-clobbered.

0001-aapcs64-Stadardize-on-caller-saved-and-callee-saved.patch
```diff
@@ -0,0 +1,278 @@
+From 122318962127bcb5b4f061402794e2022d77c1a0 Mon Sep 17 00:00:00 2001
```
smithp35: Did you manage to include this file by mistake?

aapcs64/aapcs64.rst
```diff
@@ -1317,17 +1320,16 @@ SME implementation, ZA can handle twice as much data as the vector registers,
...
+ZA. However, nested uses of ZA are expected to be rare, so these saves and
 restores would usually be wasted work.
 
 The AAPCS64 therefore defines a “lazy saving” scheme that often reduces
 the total number of saves and restores compared to the two approaches
-above.  Informally, the scheme allows “ZA is call-preserved” to become a
+above.  Informally, the scheme allows “ZA is callee-saved” to become a
 dynamic rather than a static property: if S2 `complies with the lazy saving
 scheme`_, S1 can test after the call to S2 whether the call did in fact
 preserve ZA.  If the call did not preserve ZA, S1 is able to restore the
```
smithp35: As an aside. I tried to see if I could reword this in terms of callee-save but I don't think it can be. One of the advantages of preserve rather than save, is that callee-save is really callee-responsible-for-save, not that it must save, and preserve captures that.

smithp35 (commented):
Typo in the title (Stadardize -> Standardize)

Will be worth getting some wider input on whether standardising on caller-saved and callee-saved is the best choice.

FWIW despite its problems, like only having one letter difference in names, I'm happy with caller/callee saved as that seems to be the most commonly used term on the most active ABIs.

What I could check:
x86_64 caller/callee saved.
RISCV caller/callee saved.
GO caller/callee saved [1]
PPC volatile/non-volatile.
Sparc volatile.
Microsoft (including ARM64EC [2]) volatile/non-volatile

I couldn't find any saying call-preserved or call-clobbered, but I did find explanatory text that said preserved across calls.

[1] https://go.googlesource.com/go/+/refs/heads/master/src/cmd/compile/abi-internal.md (actually all caller-saved)
[2] https://learn.microsoft.com/en-us/cpp/build/arm64ec-windows-abi-conventions https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-180

aapcs64/aapcs64.rst
```diff
@@ -427,6 +427,11 @@ Scratch register, _`temporary register`, caller-saved register
 
 Callee-saved register
    A register whose value must be preserved over a function call. If the function being called (the callee) needs to use the register, then it is responsible for saving and restoring the old value.
+   Also known as "callee preserved".
+
+Caller-saved register
+   A register whose value is not preserved over a function call. If the caller needs to use the register over the function call, then it is responsible for saving and restoring the old value.
+   Also called "call clobbered" or "call preserved".
```
momchil-velikov: Can't be both.

aapcs64/aapcs64.rst
```diff
@@ -902,7 +907,7 @@ The Arm 64-bit architecture also has a further thirty-two registers, v0-v31, whi
 
 The first eight registers, v0-v7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).
 
-Registers v8-v15 must be preserved by a callee across subroutine calls; the remaining registers (v0-v7, v16-v31) do not need to be preserved (or should be preserved by the caller). Additionally, only the bottom 64 bits of each value stored in v8-v15 need to be preserved [#aapcs64-f7]_; it is the responsibility of the caller to preserve larger values.
+Registers v8-v15 are callee-saved across subroutine calls; the remaining registers (v0-v7, v16-v31) do not need to be preserved (or should be preserved by the caller). Additionally, only the bottom 64 bits of each value stored in v8-v15 need to be preserved [#aapcs64-f7]_; it is the responsibility of the caller to preserve larger values.
```
momchil-velikov: I don't really see the need to change this phrase, it's completely clear and flows better with the second part of the sentence ("must be preserved"/"do not need to be preserved").

Generally, I would suggest standardising to "caller-saved"/"calle-saved"  instead of using other adjectives, or if there's ambiguity, but not replace any phrase that speaks about (not) preserving.

vfabia64/vfabia64.rst
```diff
@@ -433,12 +433,12 @@ can be found on `developer.arm.com
...
           table `AAVPCS Table`_ will be updated when the
           final version of the AAPCS is published.
 
 The procedural calling standard of the AAPCS requires that none of the
-32 Advanced SIMD vector registers V0-V31 are treated as call-preserved
+32 Advanced SIMD vector registers V0-V31 are treated as callee-saved
```
georges-arm: "requires that none [of the] registers are treated as callee-saved (...)" is an unnecessary negation. Can we just say "requires that V0-V31 are caller-saved (...)" instead?

aapcs64/aapcs64.rst
georges-arm: The FFR is caller-saved?

aapcs64/aapcs64.rst
```diff
@@ -956,10 +960,9 @@ Each register can store the contents of a Scalable Predicate Type
 
 p0-p3 are used to pass scalable predicate arguments to a subroutine
 and to return scalable predicate results from a function. If a
-subroutine takes at least one argument in scalable vector registers or
-scalable predicate registers, or returns results in such registers,
-the subroutine must ensure that p4-p15 are preserved across the
-call. In other cases it need not preserve any scalable predicate
+subroutine takes at least one argument in scalable vector registers or scalable
+predicate registers, or returns results in such registers, then p4-p15 are
+callee-saved. In other cases it need not preserve any scalable predicate
```
georges-arm: "In other cases all scalable predicate registers are caller-saved" ?

aapcs64/aapcs64.rst
```diff
@@ -939,10 +944,9 @@ Floating-Point register v0.
 z0-z7 are used to pass scalable vector arguments to a subroutine, and
 to return scalable vector results from a function. If a subroutine
 takes at least one argument in scalable vector registers or scalable
-predicate registers, or returns results in such registers, the
-subroutine must ensure that the entire contents of z8-z23 are
-preserved across the call. In other cases it need only preserve the
-low 64 bits of z8-z15, as described in `SIMD and Floating-Point
+predicate registers, or returns results in such registers, then
+the entire contents of z8-z23 are callee-saved. In other cases only the low
+64 bits of z8-z15 are callee-saved, as described in `SIMD and Floating-Point
```
georges-arm: "In other cases only the low 64 bits of z8-z15 are callee-saved, and the remaining vector registers (z0-z7, z16-z31, and rest of z8-z15) are caller-saved, as described ..." ?

aapcs64/aapcs64.rst
AlfieRichardsArm: Ah sorry it was supposed to be `caller-preserved`. Fixed. Thank you!

aapcs64/aapcs64.rst
AlfieRichardsArm: I agree in principle but I think this case is re-describing what "callee-saved" means so could be confusing to not use the term we standardize on?

vfabia64/vfabia64.rst
AlfieRichardsArm: Yes agreed, thank you!

aapcs64/aapcs64.rst
AlfieRichardsArm: I deliberately didn't change this case because I think "caller-saved" implies that you would ever actually want to save and restore the FFR. (going back to Peters point about preferring "caller-preserved").
 
My impression is the FFR should only hold a live value for very short periods of time. But perhaps its wrong to say how the registers are intended to be used here?

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, thanks!

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, ty

aapcs64/aapcs64.rst
AlfieRichardsArm: I agree, I think this is also relevant for Georges comment on FFR where I am hesitant to say "caller-saved" because I think it implies the wrong thing.

0001-aapcs64-Stadardize-on-caller-saved-and-callee-saved.patch
AlfieRichardsArm: Ah sorry about that. Removed for the next version.

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, thanks!

aapcs64/aapcs64.rst
AlfieRichardsArm: I will send another patch for this and the other typo

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, thank you 🤦

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, thank you 🤦

vfabia64/vfabia64.rst
AlfieRichardsArm: Done!

aapcs64/aapcs64.rst
AlfieRichardsArm: Fixed in latest after wider discussion internally.

aapcs64/aapcs64.rst
AlfieRichardsArm: Fixed in #371

aapcs64/aapcs64.rst
AlfieRichardsArm: Agreed about the typo. Fixed in #371

aapcs64/aapcs64.rst
AlfieRichardsArm: I have updated this and the surrounding section to only use "callee-saved" and "caller-saved" to refer to preservation of registers.

I do agree that these sections flow worse after these changes, but I think the change to be consistent is worth it overall. But I would be grateful for your feedback.

aapcs64/aapcs64.rst
```diff
@@ -956,11 +961,9 @@ Each register can store the contents of a Scalable Predicate Type
...
 p0-p3 are used to pass scalable predicate arguments to a subroutine
 and to return scalable predicate results from a function. If a
-subroutine takes at least one argument in scalable vector registers or
-scalable predicate registers, or returns results in such registers,
-the subroutine must ensure that p4-p15 are preserved across the
-call. In other cases it need not preserve any scalable predicate
-register contents.
+subroutine takes at least one argument in scalable vector registers or scalable
+predicate registers, or returns results in such registers, then p4-p15 are
+callee-saved. In other all scalable predicate registers are caller-saved.
```
georges-arm: Typo: "In other" ?

aapcs64/aapcs64.rst
```diff
@@ -422,11 +425,13 @@ Global register
 Program state
    The state of the program’s memory, including values in machine registers.
 
-Scratch register, _`temporary register`, caller-saved register
-   A register used to hold an intermediate value during a calculation (usually, such values are not named in the program source and have a limited lifetime). If a function needs to preserve the value held in such a register over a call to another function, then the calling function must save and restore the value.
-
 Callee-saved register
    A register whose value must be preserved over a function call. If the function being called (the callee) needs to use the register, then it is responsible for saving and restoring the old value.
+   Also known as "callee preserved register" or "non-volatile register".
```
smithp35: I think it is "call preserved register". While "callee preserved" makes sense, I don't think it is in widespread use.

aapcs64-morello/aapcs64-morello.rst
smithp35: Apologies for being a pain. Could you add the same Change Log update as aapcs64 here too?

aapcs64/aapcs64.rst
```diff
@@ -902,7 +907,7 @@ The Arm 64-bit architecture also has a further thirty-two registers, v0-v31, whi
 
 The first eight registers, v0-v7, are used to pass argument values into a subroutine and to return result values from a function. They may also be used to hold intermediate values within a routine (but, in general, only between subroutine calls).
 
-Registers v8-v15 must be preserved by a callee across subroutine calls; the remaining registers (v0-v7, v16-v31) do not need to be preserved (or should be preserved by the caller). Additionally, only the bottom 64 bits of each value stored in v8-v15 need to be preserved [#aapcs64-f7]_; it is the responsibility of the caller to preserve larger values.
+Registers v8-v15 are callee-saved across subroutine calls; the remaining registers (v0-v7, v16-v31) are callee-saved. Additionally, only the bottom 64 bits of each value stored in v8-v15 need to be callee-saved [#aapcs64-f7]_; it is the responsibility of the caller to preserve larger values.
```
smithp35: I think you can remove the "across subroutine calls" that matches the language for core-registers.

aapcs64/aapcs64.rst
AlfieRichardsArm: Fixed

vfabia64/vfabia64.rst
```diff
@@ -1575,7 +1574,7 @@ for inactive lanes.
 .. note:: Using a mask parameter in AArch64 Advanced SIMD is not
           generally recommended for functions that operate on scalars
           of different widths, as widening of the input mask for wider
-          types might require using call-preserved temporary registers
+          types might require using callee-saved temporary registers
```
smithp35: I'm wondering if it is worth removing "temporary". We define temporary registers as caller-saved so this is confusing (as was the original). 

I think this means that you may need some extra registers, and under the vfabi64 these registers are callee-saved.

smithp35 (commented):
Only a few small comments from me. This is looking close.

aapcs64/aapcs64.rst
AlfieRichardsArm: Ah agreed, Ive read them all so much they're starting to blur together. thank you!

aapcs64-morello/aapcs64-morello.rst
AlfieRichardsArm: NP, done!

aapcs64/aapcs64.rst
AlfieRichardsArm: Done, that makes sense

vfabia64/vfabia64.rst
AlfieRichardsArm: Yeah I was quite confused by this so tried to keep as much of the original intent there as possible.

I agree with your interpretation and will remove.

vfabia64/vfabia64.rst
```diff
@@ -459,7 +458,7 @@ AAVPCS Table
 .. table:: Modified PCS for vector functions (AAVPCS)
 
   +-------------+--------------------+--------------+--------------+
-  |Extension    |Parameter and Result|Call-clobbered|Call-preserved|
+  |Extension    |Parameter and Result|Caller-saved  |Callee-saved  |
```
MaskRay, line 461: IMHO 'call-clobbered' and 'call-preserved' are much better terms. 'Caller-saved' and 'callee-saved' are only a letter apart, making them unnecessarily easy to mix up when reading quickly...

momchil-velikov:
Out compiler engineer bias might be showing here, but I'm under the subjective impression that "callee-saved" and "caller-saved" are much more established terms in the community and the literature.

FWIW, I did a quick search in ACM DL:

* "call-clobbered": 0 results
* "call-preserved": 7 results
* "callee-saved": 285 results
* "caller-saved": 186 results

aapcs64/aapcs64.rst
```diff
@@ -2234,9 +2234,9 @@ following properties:
 * The function has a `private-ZA`_ `streaming-compatible interface`_ with
   following properties:
 
-  * X2-X15, X19-X29 and SP are call-preserved.
-  * Z0-Z31 are call-preserved.
-  * P0-P15 are call-preserved.
+  * X2-X15, X19-X29 and SP are callee-saved.
```
Wilco1: Since there was some confusion about this recently, we should explicitly state these private helpers must not clobber X18 (since it is reserved and has various uses). GCC currently treats X18 as clobbered by these helpers, which interferes with some uses of X18.

This also applies to the other helpers below.

vfabia64/vfabia64.rst
richfelker: > 'Caller-saved' and 'callee-saved' are only a letter apart, making them unnecessarily easy to mix up when reading quickly...

This is indeed another very good reason to prefer the call-clobbered/call-preserved terminology.

aapcs64/aapcs64.rst
AlfieRichardsArm: Thanks for the review. I may do this in another patch if that's okay with you? Keep this one "no functional change intended"?

aapcs64/aapcs64.rst
```diff
@@ -422,11 +425,13 @@ Global register
 Program state
    The state of the program’s memory, including values in machine registers.
 
-Scratch register, _`temporary register`, caller-saved register
```
smithp35, line 425: In light of the caller-saved -> call-clobbered discussion, it could be worth making all references to callee-saved and caller-saved into links back to the definitions. 

This would mean a lot of references, but it would at least mean that it could be looked up quickly.

smithp35 (commented):
Thanks for the updates. I'm happy all other existing comments from me have been addressed.

On the caller-saved/callee-saved vs call-preserved/call-clobbered question.

I think what we have now is an awkward mix of the two terms, which is the worse situation. Standardising on one way, even if it is not our preference is a step in the right direction.

I do agree that there are advantages to call-preserved and call-clobbered that have been mentioned already. There are also advantages to caller and callee-saved, mostly that they are the most commonly used terms in ABIs, including historically the Arm and AArch64 ABIs and the compiler literature.

I think we should go ahead with this PR. Leaving the door open for a change to call-preserved, call-clobbered if that term becomes more widely used, or there is a consensus internally and externally that it is worth changing. We'll at least be starting from a consistent base at that point.

aapcs64/aapcs64.rst
AlfieRichardsArm: Ah yes thats a good idea. Done!

smithp35 (approved):
Thanks for the addition of the links. I've no more comments and have given it my +1 with Approve.

We should leave this open for at least a week for any further comments from other reviewers.

smithp35:
@AlfieRichardsArm I've merged #373 I think this needs rebasing. As there have been no more comments since last week I can merge this once it has been rebased.

AlfieRichardsArm:
@smithp35 Rebased now, thanks for your help with this!

vfabia64/vfabia64.rst
lemon32767: > > 'Caller-saved' and 'callee-saved' are only a letter apart, making them unnecessarily easy to mix up when reading quickly...
> 
> This is indeed another very good reason to prefer the call-clobbered/call-preserved terminology.

Case in point: #385

iknowimnotgood (approved):

PR #372, opened 2026-03-02
[pauthabi64][aaelf64] Standardise on Delta capitalisation. [NFC]

smithp35:
The Delta relocation operator defined in aaelf64.rst has been referred to by DELTA. This change standardises all documents to use Delta which was the original definition, and the same capitalisation as used for other relocation operators that are also English words.


tom-cosgrove-arm (approved):
LGTM FWIW :)

PR #374, opened 2026-03-05
[pauthabi64] Correct relocation codes for memtag section

smithp35:
PR https://github.com/ARM-software/abi-aa/pull/366 which introduced a section to clarify interaction between Memtag and PAuthABI has the wrong relocation codes for the R_AARCH64_AUTH_RELATIVE and R_AARCH64_GLOB_DAT dynamic relocations. Make sure that they are consistent with the existing document and the implementation in LLVM.


yury-khrustalev (approved):
LGTM Reviewed-by: @yury-khrustalev

PR #378, opened 2026-03-10
[aapcs64] Specify the `__aarch64_sme_accessible` SME support routine

MacDue:
This is documenting an existing practice by LLVM's compiler-rt to depend (weakly) on `__aarch64_sme_accessible`. This is used to determine if SME is available on platforms where there is not a defined way of checking the CPU features.

The intent is to document this so libc implementations know what they need to provide.

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -2218,6 +2218,8 @@ support routines:
 ``__arm_sme_restore``
    Provides a safe way to restore state enabled by PSTATE.ZA from a buffer.
 
+``__aarch64_sme_accessible``
+   Provides a safe way to query if the current thread has `access to SME`_.
 
 ``__arm_sme_state``
 ^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


MacDue:
cc @sdesmalen-arm (note: I can't add reviewers in this repo).

sdesmalen-arm:
I'm not sure that this is the right solution, because now we have two functions that define whether SME is available (`__arm_sme_state` and `__aarch64_sme_accessible`). `__aarch64_sme_accessible` is not explicitly specified, but it is also used by libgcc. A libc for a baremetal target could implement `__aarch64_sme_accessible` if they want LLVM's compiler-rt routines to handle SME code. Because the ABI doesn't really specify where these routines should live this is more of an implementation decision so that the routines can live in the compiler runtime. Perhaps the better solution would be to let LLVM's compiler-rt define `__arm_sme_state` as a weak symbol so that a libc for a bare-metal target can override it if they want to use a custom implementation that doesn't rely on `__aarch64_sme_accessible`.

MacDue:
One thing to keep in mind is this does not just effect `__arm_sme_state`. Several ABI routines include a step that tests if SME is available. So I can see that it could make sense to separate the "is SME available" step (that compiler-rt can't implement for baremetal) to a common function. 

Alternatively, I think it'd make more sense not to implement `__arm_sme_state` on baremetal at all (no weak symbol) and ensure all availability checks are in terms of ``__arm_sme_state``.

Issue #239, opened 2024-01-22
Clarify use of DWARF versions


Stammark:
Hi all,

In:

https://github.com/ARM-software/abi-aa/blob/main/aadwarf32/aadwarf32.rst#overview
https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst#overview

We have references to DWARF 3.0:
`The ABI for the Arm architecture specifies the use of DWARF 3.0-format debugging data.`
`The ABI for the Arm 64-bit architecture specifies the use of DWARF 3.0 format debugging data. `

These could be misread as a "Arm requires you to use DWARF 3.0" which isn't valid in practice, these days.

I'm raising this issue to suggest that be changed to more flexible wording. That could be something like:
* We recommend/require? a minimum of DWARF 3.0. Then the rest of the doc is version-agnostic (>= 3.0 is implied)
* We recommend/require? a minimum of DWARF 2.0. Then the rest of the doc is version-agnostic (>= 2.0 is implied)
* We make the whole doc version-agnostic (unlikely to be possible, as that would include DWARF1)

And in all those cases we could come up with a statement like "refer to the documentation of your debugger for the best compatible DWARF version" (since that is outside the scope of the aadwarf doc)

sivan-shani:
In my understanding, the document does not have to recommend or require any version, but rather to refer to a base line version.
The document defines DWARF extensions, which are built on a baseline Dwarf definitions. This baseline is Dwarf 3. That is all the document has to convey.
I will suggest  an edit base on this take.

sivan-shani:
Commit #388 address this issue, merged now, closing this.

PR #388, opened 2026-04-01
[aadwarf<32/64] Clarify DWARF standard baseline [issue #239]

sivan-shani:
Clarify the relation between aadwarf and the DWARF standard baseline it reffers.

### DIFF
--- a/aadwarf32/aadwarf32.rst
+++ b/aadwarf32/aadwarf32.rst
@@ -312,12 +312,11 @@ River.
 Overview
 ========
 
-The ABI for the Arm architecture specifies the use of DWARF 3.0-format
-debugging data.  For details of the base standard see GDWARF_.
+The ABI for the Arm architecture uses the DWARF 3.0 standard as its baseline.
+For details of the base standard, see GDWARF_.
 
-The ABI for the Arm architecture gives additional rules for how DWARF 3.0
-should be used, and how it is extended in ways specific to the Arm
-architecture. The following topics are covered in detail.
+It defines additional rules and Arm-specific extensions to DWARF 3.0,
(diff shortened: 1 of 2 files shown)


sivan-shani:
Change suggestion in response to issue: https://github.com/ARM-software/abi-aa/issues/239
@Stammark @smithp35 @walkerkd

aadwarf32/aadwarf32.rst
```diff
@@ -312,12 +312,11 @@ River.
...
 
-The ABI for the Arm architecture specifies the use of DWARF 3.0-format
-debugging data.  For details of the base standard see GDWARF_.
+The ABI for the Arm architecture uses the DWARF 3.0 standard as its baseline.
+For details of the base standard, see GDWARF_.
 
-The ABI for the Arm architecture gives additional rules for how DWARF 3.0
-should be used, and how it is extended in ways specific to the Arm
-architecture. The following topics are covered in detail.
+It further defines additional rules and Arm-specific extensions to DWARF 3.0,
```
smithp35: I think this should be "It defines additional ..." or "The ABI for the Arm Architecture defines additional ..."

It further defines is normally used to expand on a specific topic in more detail. Here we are stating that we are defining additional rules.

smithp35 (commented):
One small suggestion on rewording "It further defines", but otherwise looks good to me.

Thanks for the fix!

aadwarf32/aadwarf32.rst
sivan-shani: That makes sense, changing accordingly.

smithp35 (approved):
Thanks for the update. LGTM.

PR #395, opened 2026-05-14
[pauthabielf64] Define R_AARCH64_AUTH_TLSDESC_CALL.

### LINKED ISSUE #393: [PAUTHABIELF64] Introduce `R_AARCH64_AUTH_TLSDESC_CALL` and `.tlsdescauthcall`
## Background

The PAUTHABIELF64 document currently specifies the following relocations for the AUTH TLSDESC sequence:

- `R_AARCH64_AUTH_TLSDESC_ADR_PAGE21`
- `R_AARCH64_AUTH_TLSDESC_LD64_LO12`
- `R_AARCH64_AUTH_TLSDESC_ADD_LO12`

The full AUTH TLSDESC call sequence is:

```asm
  adrp  x0, :tlsdesc_auth:var              // R_AARCH64_AUTH_TLSDESC_ADR_PAGE21
  ldr   x16, [x0, #:tlsdesc_auth_lo12:var] // R_AARCH64_AUTH_TLSDESC_LD64_LO12
  add   x0, x0, #:tlsdesc_auth_lo12:var    // R_AARCH64_AUTH_TLSDESC_ADD_LO12
  blraa x16, x0
  // TPIDR_EL0 offset now in x0
```

Note that `blraa` is required here because the AUTH TLSDESC handler pointer is stored in a signed GOT entry using the IA key.

## Problem

For non-auth TLSDESC, the `R_AARCH64_TLSDESC_CALL` relocation (emitted via the `.tlsdesccall` assembler directive) annotates the `blr` instruction, allowing the linker to locate it for relaxation:

```asm
  adrp  x0, :tlsdesc:var             // R_AARCH64_TLSDESC_ADR_PAGE21
  ldr   x1, [x0, #:tlsdesc_lo12:var] // R_AARCH64_TLSDESC_LD64_LO12
  add   x0, x0, #:tlsdesc_lo12:var   // R_AARCH64_TLSDESC_ADD_LO12
  .tlsdesccall var                   // R_AARCH64_TLSDESC_CALL
  blr   x1
```

No equivalent annotation relocation exists for the AUTH TLSDESC sequence. Without it, the linker cannot locate the `blraa` instruction to perform relaxation.

## Motivation

For undefined weak non-preemptible symbols, the linker knows at link time that the symbol resolves to NULL. In this case, the entire AUTH TLSDESC sequence can be relaxed:

```asm
  // Before (AUTH TLSDESC):
  adrp  x0, :tlsdesc_auth:var              // R_AARCH64_AUTH_TLSDESC_ADR_PAGE21
  ldr   x16, [x0, #:tlsdesc_auth_lo12:var] // R_AARCH64_AUTH_TLSDESC_LD64_LO12
  add   x0, x0, #:tlsdesc_auth_lo12:var    // R_AARCH64_AUTH_TLSDESC_ADD_LO12
  .tlsdescauthcall var                     // R_AARCH64_AUTH_TLSDESC_CALL
  blraa x16, x0

  // After (relaxed):
  mov   x0, #0x0
  nop
  nop
  nop
```

This is the only relaxation case applicable to AUTH TLSDESC. Broader relaxations (such as GD->IE or GD->LE with non-statically-known-NULL symbols) are not possible because pointer authentication requires signing at program start-up.

## Proposal

Introduce:

1. **`R_AARCH64_AUTH_TLSDESC_CALL`** — a relocation that annotates the `blraa` instruction in an AUTH TLSDESC call sequence, analogous to `R_AARCH64_TLSDESC_CALL` for the non-auth case.
2. **`.tlsdescauthcall`** — an assembler directive that emits this relocation, analogous to `.tlsdesccall`.

---

See also https://github.com/llvm/llvm-project/issues/173296 tracking progress on related issue.

smithp35:
The TLSDESC sequence for accessing an authenticated pointer is similar to the traditional TLSDESC sequence. As pointer signing must be done at run-time there are much more limited opportunities to relax a TLSDESC AUTH, we introduce a R_AARCH64_AUTH_TLSDESC_CALL relocation that permits a static linker to transform the blraa to a nop, when relaxation is possible.

A new relocation code has been introduced rather than reusing R_AARCH64_TLSDESC_CALL so that a static linker can assume that the destination symbol is signed, without having to derive it from other R_AARCH64_TLSDESC_* relocations.

TLSDESC sequence

adrp x0, :tlsdesc:v             //R_AARCH64_TLSDESC_ADR_PAGE21
ldr  x1, [x0, #:tlsdesc_lo12:v] //R_AARCH64_TLSDESC_LD64_LO12
add  x0, x0, #:tlsdesc_lo12:v   //R_AARCH64_TLSDESC_ADD_LO12
.tlsdesccall var                //R_AARCH64_TLSDESC_CALL
blr  x1

TLSDESC AUTH sequence

adrp x0, :tlsdesc_auth:v             //R_AARCH64_AUTH_TLSDESC_ADR_PAGE21
ldr  x16, [x0, #:tlsdesc_auth_lo12:v]//R_AARCH64_AUTH_TLSDESC_LD64_LO12
add  x0, x0, #:tlsdesc_auth_lo12:v   //R_AARCH64_AUTH_TLSDESC_ADD_LO12
.tlsdescauthcall v                   //R_AARCH64_AUTH_TLSDESC_CALL
blraa x16, x0

fixes #393


kovdan01:
Thanks @smithp35!

Just for transparency - PRs implementing support in llvm-project:
- backend: https://github.com/llvm/llvm-project/pull/198327
- lld linker: https://github.com/llvm/llvm-project/pull/194636

pauthabielf64/pauthabielf64.rst
```diff
@@ -1201,6 +1203,33 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
...
+  of TLS descriptors are marked to permit accurate relaxation.
+
+.. code-block:: asm
+
+  // Before (AUTH TLSDESC):
+  adrp x0, :tlsdesc_auth: undefined_weak            // R_AARCH64_AUTH_TLSDESC_ADR_PAGE21
+  ldr  x16, [x0, :tlsdesc_auth_lo12: undefined_weak // R_AARCH64_AUTH_TLSDESC_LD64_LO12
+  add  x0, x0 :tlsdesc_auth_lo12: undefined_weak    // R_AARCH64_AUTH_TLSDESC_ADD_LO12
+  .tlddescauthcall undefined_weak                   // R_AARCH64_AUTH_TLSDESC_CALL
+  autia x0, x8
```
jrtc27: https://github.com/ARM-software/abi-aa/issues/393#issue-4421518130 gives this as `blraa x16, x0` instead?

pauthabielf64/pauthabielf64.rst
```diff
@@ -1201,6 +1203,33 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
...
+
+  // After relaxation, assuming undefined_weak is known to be 0 at static-link time.
+  mov  x0, #0x0
```
jrtc27: If the resolver is returning a (potentially between distinct allocations) offset from TP, wouldn't this cause a change in behaviour from giving "NULL" to giving "TP"? At least within FreeBSD our normal AArch64 TLSDESC resolver for undefined weak symbols is to return -TP(+A) so adding it to TP gives NULL(+A). I know undefined weak TLS objects are historically very cursed and break in all kinds of ways, but this would be a regression over non-PAuth TLSDESC, I think.

jrtc27: (This also loses the addend entirely)

pauthabielf64/pauthabielf64.rst
smithp35: Thanks for spotting. Will update.

pauthabielf64/pauthabielf64.rst
smithp35: This is a good point, @kovdan01 is your TLSDESC resolver function for an undefined weak capable of returning -TP? I believe AArch64 glibc does this too for undefined weak TLS symbols.

I think this would be preferable to relaxing the sequence to 0 as currently that 0 would get added to TP, which would not result in a value of 0 for an undefined weak.

There's always the possibility of altering the TLSDESC sequence of checking for 0 before adding to TP as 0 isn't a valid offset (first offset is TCB size + alignment padding) which is a minimum of 16. However this seems worse than the resolver.

As an aside I've not been able to create a TLSDESC sequence with a non-zero addend from some simple compiled code. Always seems like the TLSDESC calculates the address of the symbol, then does an addition, or load with immediate offset instead.

smithp35:
I've fixed the typos and added a note that when 0 is added to the TP it will point at the thread control block.

Do we still need this? EDIT, yes as on other platforms there may be other relaxations possible.

pauthabielf64/pauthabielf64.rst
jrtc27: Oh, that may be true in practice, just like GOT entries typically don't actually have addends (except when turned to relative ones, of course) as it's annoying to have potentially multiple entries per symbol to track in the linker.

smithp35:
Reading through https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-ARM.txt again. There is a relaxable sequence that could be used when static linking (so we know that the weak reference won't be defined). In effect this is inlining the resolver function that returns -TP
```
mrs x0, TPIDR_EL0
neg x0, x0 // alias of sub x0, xzr, x0
nop
nop
```

pauthabielf64/pauthabielf64.rst
```diff
@@ -1201,6 +1203,37 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
   |                 |                                        |                                        | field to bits [11:0] of  |
   |                 |                                        |                                        | X. No overflow check.    |
   +-----------------+----------------------------------------+----------------------------------------+--------------------------+
+  | 0x266 (598)     | R\_AARCH64\_AUTH\_TLSDESC\_CALL        | None                                   | For relaxation only. See |
+  |                 |                                        |                                        | notes below.             |
+  +-----------------+----------------------------------------+----------------------------------------+--------------------------+
+
+.. note::
+
+  Relocation code ``R_AARCH64_AUTH_TLSDESC_CALL`` is needed to permit
```
smithp35, line 1219: I think I'll move the details about the relaxation to a separate document in the design-documents folder.

I think it is platform specific choice of whether both fields of the TLS descriptor are signed. If only the resolver function address is signed then more relaxations are possible.

Likely to be next week before I can do that.

pauthabielf64/pauthabielf64.rst
kovdan01: > @kovdan01 is your TLSDESC resolver function for an undefined weak capable of returning -TP? I believe AArch64 glibc does this too for undefined weak TLS symbols.

@smithp35 Yes, glibc does this, and we match that behavior in our PoC reference musl implementation - see https://github.com/access-softek/musl/blob/v1.2.5-pauth-rev2025-11-21/src/ldso/aarch64/tlsdesc.S#L63-L70

pauthabielf64/pauthabielf64.rst
jrtc27: Well, it's not about whether the second word of the descriptor is signed. The generated code never uses that, either it passes a pointer to it as an opaque blob to the resolver or it relaxes the entire sequence so there is no descriptor to sign? The question is whether the TLS data is being signed like globals can be, and therefore whether `&tls_var - TP` is the same for all threads or differs in the high bits due to signing. If the data isn't signed, you can just relax to that constant (which will be the same as non-PAuth, and is either an LE immediate or an IE run-time constant), it's only if the data is signed that IE/LE are fundamentally broken?

smithp35: Thanks for the comment. I'll probably just take the rationale/relaxation bits out of the main document for now until I've got the time to work through this slowly.

Reading through the initial issue again https://github.com/ARM-software/abi-aa/issues/393. I think I've put too much weight on the comment:

> Broader relaxations (such as GD->IE or GD->LE with non-statically-known-NULL symbols) are not possible because pointer authentication requires signing at program start-up.

I've missed the non-statically-known-NULL symbols part and in my haste reading. I had been trying to reconcile why other relaxations weren't possible with the code sequence the compiler uses for TLSDESC and a signed GOT.

As an aside:

Empirically using: `clang --target=aarch64-linux -march=armv8.3-a -S -O2 tlsdesc.c -o - -fptrauth-elf-got -mabi=pauthtest` with a `trivial __thread int x; int val() { return x; }`

I get:
```
        pacibsp
        stp     x29, x30, [sp, #-16]!           // 16-byte Folded Spill
        mov     x29, sp
        adrp    x0, :tlsdesc_auth:x
        ldr     x16, [x0, :tlsdesc_auth_lo12:x]
        add     x0, x0, :tlsdesc_auth_lo12:x
        blraa   x16, x0
        mrs     x8, TPIDR_EL0
        ldr     w0, [x8, x0]
        ldp     x29, x30, [sp], #16             // 16-byte Folded Reload
        retab
```
If I'm reading that correctly, the return value from the TLS resolver function isn't signed. Nor is the value of x. It looks like the only thing that is signed is the descriptor in the GOT.

That looks relaxable in principle, although maybe not in practice to initial exec. I'd expect if it were initial exec the (&tls_var - TP) would be signed in the GOT, there are enough spare instructions and registers to extract the unsigned (&tls_var - TP), but there aren't enough spare instructions to test whether the authenticate failed, which I believe is a requirement for `-fptrauth-traps` (for systems without FEAT_FPAC).

smithp35:
I've split out the commentary/rationale about TLS into a separate design document based on my understanding of how TLS is being used today, and what could be done in a different signing-schema (defined separately from the PAuthABI).

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+there are any discrepancies.
+
+TLS Dialect
+-----------
+
+The base ABI in `AAELF64`_ defines relocations for two TLS
+dialects. The "traditional" dialect and the "descriptor" dialect. In
+the traditional dialect global and local dynamic TLS use the
+``R_<CLS>_TLSGD`` and ``R_<CLS>_TLSLD`` prefixed relocations. These
+create a pair of GOT entries relocated by ``R_<CLS>_TLS_DTPMOD``. In
```
jrtc27: and DTPREL

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+the "descriptor" dialect, global and local dynamic TLS use the
+``R_<CLS>_TLSDESC`` prefixed relocations. These create a pair of GOT
+entries relocated by ``R_<CLS>_TLSDESC``. Local Exec and Initial Exec
+TLS are handled the same way in both dialects.
+
+The `PAUTHABIELF64`_ only supports the descriptor based dialect,
+primarily because clang only supports the "descriptor" based dialect.
```
jrtc27: I think a stronger reason these days is because the traditional one is legacy so a new ABI should just follow the new approach? Clang supports traditional for a bunch of architectures so adding it for AArch64 would be straightforward.

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+
+If a GOT entry is signed then AUTH variant static TLS relocations are
+required for the static linker to be able use an AUTH variant dynamic
+relocation that informs the dynamic linker to sign the GOT entries.
+
+At present there are AUTH variant static and dynamic relocations
+defined for TLSDESC, but not for Initial Exec.
+
+Local dynamic TLS does not use the GOT so it can be handled by the
+``R_<CLS>_TLSLE`` prefixed relocations defined in `AAELF64`_.
```
jrtc27: Maybe reiterate "from the base ABI" like a few paragraphs above to be clearer it's unchanged?

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+
+The choice of which GOT entries to sign is a property of the
+signing-schema for the platform. For example a signing-schema may only
+sign GOT entries containing code-pointers, which would permit Initial
+Exec TLS using the ``R_<CLS>_TLSIE`` prefixed relocations defined in
+`AAELF64`_. Alternatively a signing-schema may sign all GOT entries.
```
jrtc27: "..., which would require AUTH variant static and dynamic relocations to be defined for Initial Exec"?

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+
+TLS Relaxation
+--------------
+
+The static linker may relax a more general TLS model to a more
+constrained model when TLS variables meet the requirements for using
+the constrained model, and the relaxed sequence is permitted by the
```
jrtc27: I feel this comma hurts legibility, as the "and" relates to the "when" rather than the whole clause? At least, I had to backtrack in my head to parse it properly.

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,130 @@
...
+     nop
+
+A theoretical relaxation for a signed GOT is possible when the
+architectural feature ``FEAT_FPAC`` is available or a signing-schema
+does not require an immediate trap on failure of an AUTH
+(``-fno-ptrauth-traps`` in clang).
+
+  .. code
+
+     adrp x0, :gottprel:v           // R_AARCH64_AUTH_TLSIE_ADR_GOTTPREL_PAGE21 v
```
jrtc27: For this hypothetical support would you not need auth somewhere in the modifiers(?) to get the right relocation, like `:got_auth(_lo12):` (presumably `:gottprel_auth(_lo12):`)?

pauthabielf64/pauthabielf64.rst
```diff
@@ -1201,6 +1207,18 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
...
+  linker optimization of TLS descriptor code sequences involving
+  authenticated pointers, when undefined weak non-preemptible symbols
+  are known to resolve to 0; this can only be done if all relevant uses
+  of TLS descriptors are marked to permit accurate relaxation.
```
jrtc27: This should probably have a very brief outline of the cases discussed in the design doc and refer to it, rather than only list undefined weak non-preemptible symbols as the case that can be relaxed?

jrtc27 (commented):
Thanks, a lot clearer to me now as to what this is actually trying to achieve and how. Some comments inline.

design-documents/pauthabi-tls.rst
smithp35: ACK

design-documents/pauthabi-tls.rst
smithp35: ACK

design-documents/pauthabi-tls.rst
smithp35: ACK

design-documents/pauthabi-tls.rst
smithp35: ACK, have added that in.

design-documents/pauthabi-tls.rst
smithp35: Removed the comma.

design-documents/pauthabi-tls.rst
smithp35: Yes, I've added in auth to match the other ones.

pauthabielf64/pauthabielf64.rst
smithp35: I've removed the part about undefined weak, thought I'd done that already but my eyes were deceiving me.

I've replaced it with a reference to the design document.

smithp35 (commented):
Thanks very much for the review, and apologies for the delay in responding. I'll upload a new version.

pauthabielf64/pauthabielf64.rst
```diff
@@ -1201,6 +1210,17 @@ The GOT entries must be relocated by AUTH variant dynamic relocations.
...
+  | 0x266 (598)     | R\_AARCH64\_AUTH\_TLSDESC\_CALL        | None                                   | For relaxation only. See |
+  |                 |                                        |                                        | notes below.             |
+  +-----------------+----------------------------------------+----------------------------------------+--------------------------+
+
+.. note::
+
+  Relocation code ``R_AARCH64_AUTH_TLSDESC_CALL`` is needed to permit
+  linker optimization of TLS descriptor code sequences involving
+  signed GOT entries. Further information, including possible
+  relaxations is available in the `PAUTHABITLS`_ design document.
```
jrtc27: nit:
```suggestion
  relaxations, is available in the `PAUTHABITLS`_ design document.
```

smithp35: ACK, have adopted.

design-documents/pauthabi-tls.rst
kovdan01, line 1: Nit: GitHub's markdown rendering is not showing code blocks from this file for some reason (at least for me in chromium). Maybe it's worth considering using triple-backtick usual formatting or fixing it somehow else - I guess that many people would read it in GitHub's web interface and would just miss the code blocks as not displayed.

design-documents/pauthabi-tls.rst
```diff
@@ -0,0 +1,135 @@
...
+
+     adrp x0, :gottprel:v            // R_AARCH64_TLSIE_ADR_GOTTPREL_PAGE21 v
+     ldr  x0, [x0, #:gottprel_lo12:v] // R_AARCH64_TLSIE_LD64_GOTTPREL_LO12_NC v
+     nop
+     nop
+
+A theoretical relaxation for a signed GOT is possible when the
+architectural feature ``FEAT_FPAC`` is available or a signing-schema
+does not require an immediate trap on failure of an AUTH
+(``-fno-ptrauth-traps`` in clang).
```
kovdan01: Nit: I guess that `-fptrauth-auth-traps` was meant (`auth` missed in the middle) - see the flag definition https://github.com/llvm/llvm-project/blob/6e4e181c83d076bb8e4c208ae28817f3900b23fa/clang/include/clang/Options/Options.td#L4964

design-documents/pauthabi-tls.rst
smithp35: ACK, I'll switch to using `code-block:: asm`

design-documents/pauthabi-tls.rst
smithp35: ACK, will update to fix.

smithp35:
I'll ask internally to see if I can get someone to review this. If this PR is good enough for you, please do accept the changes. It will help an internal reviewer that may not be familiar with the PAuthABI.

kovdan01:
Having this reviewed and accepted in the docs sooner than later would be highly beneficial for LLVM mainline - it would allow to merge a corresponding dependent series of PRs which fix behavior for certain cases with undefined weak symbols.

And it would be nice if we can have it merged in LLVM before the upcoming release - to the best of my knowledge, the release branch is expected to be tagged on July 14th

DanielKristofKiss (approved):
LGMT

jrtc27 (approved):

smithp35:
Thanks for the reviews!

PR #397, opened 2026-06-01
[memtagabielf64] Clarify scope of memtagabielf64 document

### LINKED ISSUE #381: [MemtagABI] Clarify scope and status of the document
The MemtagABI document currently states in the "Platform Standards" section:

> each operating system that adopts components of this ABI specification will specify additional requirements and constraints that must be met by application code in binary form and the code-generation tools that generate such code

We should clarify that parts of the Memtag ABI that are applicable to Linux should be clarified in the SYSVABI specification:

Add this in the end:

> This document does not explicitly apply to Linux, see the SYSVABI and AAPCS64 specifications for Linux implementation details.

yury-khrustalev:
Clarify the scope of the document to avoid confusion about applicability of the recommendations in this document to Linux operating systems.

Fixes #381

### DIFF
--- a/memtagabielf64/memtagabielf64.rst
+++ b/memtagabielf64/memtagabielf64.rst
@@ -9,6 +9,8 @@
 
 
 .. _AAELF64: https://github.com/ARM-software/abi-aa/releases
+.. _AAPCS64: https://github.com/ARM-software/abi-aa/releases
+.. _SYSVABI: https://github.com/ARM-software/abi-aa/releases
 .. _RELR: https://groups.google.com/g/generic-abi/c/bX460iggiKg
 .. _CPPABI64: https://github.com/ARM-software/abi-aa/releases
 .. _MTE: https://developer.arm.com/documentation/102433/0100/Detecting-memory-safety-violations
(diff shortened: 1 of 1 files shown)


yury-khrustalev:
Hi @fmayer, @pcc, @zatrazz, looking for your feedback here. Thanks!

zatrazz:
I do not have a strong opinion about it, but maybe 'Linux' is too broad, since it encompasses not only the GNU ABI but also other unrelated ABIs/runtimes (Go, Rust, Java, etc.). And it is also misleading because it is already partially supported on Android (which is essentially Linux), and some pieces are in binutils and LLVM.

I would leave the text as-is; it already states that it is a recommendation, and the ecosystem is free to implement it or not.

memtagabielf64/memtagabielf64.rst
```diff
@@ -289,8 +293,11 @@ Platform Standards
 As is the case with the AAELF64_, we expect that each operating system that
 adopts components of this ABI specification will specify additional requirements
 and constraints that must be met by application code in binary form and the
-code-generation tools that generate such code. This document will present
-recommendations for a SysVr4 like operating system such as Linux.
+code-generation tools that generate such code.
+
+This document will present recommendations for a SysVr4 like operating systems
```
smithp35: Perhaps

This document will present recommendations for a SysVr4 like operating system such as Linux. Any specific GNU/Linux implementation details will be documented in the `SYSVABI`_ and `AAPCS64`_ documents.

smithp35 (commented):
I don't have a strong opinion about adding the clarification either. Perhaps we can wait until there is some specific differences that can be highlighted.

I've made a wording change that uses GNU/Linux which is perhaps a bit more distinct from Android.

smithp35 (approved):
I've marked this as approved. While there are no strong opinons for, there are no strong opinions against and I'm inclined to go with intent of the author here for a small NFC change.

Can wait a few days to see if there are any objections.

PR #398, opened 2026-06-10
[PAUTHABIELF64] Introduce an alternative relocation encoding

smithp35:
Embedded systems may require dynamic relocations to be idempotent as they need to initialize, then copy themselves to a new address for ASLR purposes, then resign.

As idempotent relocations are an ABI break for existing platforms and have additional limitations over the current encoding this is written as an Appendix for comment.

Platforms are expected to choose one form of encoding for their relocations.

It is possible to transform the existing encoding into the new one with a post-link tool, however the alternative format could be implemented directly in a static linker.

If there is a static-linker implementation, my current thought is that it would be best to encode the desire for idempotent relocations via the signing-schema in the object file. That way it is harder to get incompatibilites arising from someone forgetting a command-line options.

### DIFF
--- a/pauthabielf64/pauthabielf64.rst
+++ b/pauthabielf64/pauthabielf64.rst
@@ -250,6 +250,9 @@ changes to the content of the document for that release.
   |            |                             |   Range.                                                         |
   +------------+-----------------------------+------------------------------------------------------------------+
   | 2026Q2     | 6\ :sup:`th` January 2026   | - Clarify combination of MemtagABI and PAuthABI relocations      |
+  |            |                             | - Add alternative format of dynamic relocations for platforms    |
+  |            |                             |   that require idempotency and cannot store the signing-schema   |
+  |            |                             |   in the place.                                                  |
   +------------+-----------------------------+------------------------------------------------------------------+
 
 References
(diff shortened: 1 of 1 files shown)


smithp35:
This has been created as a result of Discord Thread https://discord.com/channels/636084430946959380/1133112394701348895/1513540731615314050 which contains more background for the use case.

EliaGeretto:
I have read through this PR and having this workaround described in the PAuth ABI extension document would work for us because we would have an authoritative reference to justify our implementation. Another possibility is to copy the signature schema from the "place" into the addend field on the first relocation round, assuming the added has at most 32 bits.

I would just like to raise the issue of finding a "proper" solution to store additional relocation metadata without sacrificing neither addend space nor the idempotency guarantee given here: https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#idempotency. I can only think of very painful solutions:

- adding a new relocation format that includes space for metadata,
- using multiple relocation entries to perform a single relocation, which implies preserving state across entries,
- having a completely separate metadata section.

In addition, we would also need something that ensures the combination of different extensions, like PAuth and Memtag, does not create conflicts. Even storing everything in the "place", as it is now, there is no 32 bits limit on the tag offset in the Memtag ABI extension, so it could technically overflow into the PAuth schema if the two are combined.

pauthabielf64/pauthabielf64.rst
```diff
@@ -1287,6 +1287,97 @@ may choose to fault an AUTH and a non-AUTH GOT generating relocation
...
+* ``ADDEND(A)`` is the result of extracting the bottom 32-bits of A.
+
+* ``SCHEMA(A)`` is the result of the dynamic linker reading the signing schema from A.
+
+.. table:: Dynamic relocations
+
+  +--------------------+------------------------------+------------------------------------+
+  | Relocation code    | Name                         | Operation                          |
+  +====================+==============================+====================================+
+  | 0x244 (580)        | R\_AARCH64\_AUTH\_ABS64      | SIGN(S + ADDEND(A), SCHEMA(A)      |
```
MaskRay: typo: Missing `)` in `SIGN(S + ADDEND(A)`

smithp35: Thank you. Now fixed.

PR #400, opened 2026-06-27
[aaelf64] Note PLT32 is always signed

jrtc27:
Unlike ABSn/PRELn, PLT32's range check is always for a signed integer.

Whilst here, improve the grammer/formatting of the comment in the table and link to the referenced section.


jrtc27:
Spotted whilst looking at https://github.com/llvm/llvm-project/pull/205963

smithp35 (approved):
Thanks for the PR. The note may need updating again if `R_<CLS>_PREL32` can be updated to match BFD's implementation, but worth having it until that's changed.

PR #404, opened 2026-07-08
[NFC] Fix mismatched quotes

tommat01:
This seems to have been copy-pasted around but I don't think it's valid to mix straight quotes and fancy quotes.

### DIFF
--- a/aadwarf32/TRADEMARK_NOTICE
+++ b/aadwarf32/TRADEMARK_NOTICE
@@ -1,6 +1,6 @@
 The text of and illustrations in this document are licensed
 under a Creative Commons Attribution–Share Alike 4.0 International
-license ("CC-BY-SA-4.0”), with an additional clause on patents.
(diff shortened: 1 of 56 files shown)


smithp35:
> This seems to have been copy-pasted around but I don't think it's valid to mix straight quotes and fancy quotes.

It also seems to be present in the ACLE https://github.com/ARM-software/acle/blob/main/main/acle.md#trademark-notice which isn't surprising given the move of the ABI and ACLE to that license.

Normally I'd just approve, but as this is touching the license text I'd like to make sure we're aligned with the ACLE.

smithp35:
Agreed with ACLE team to align on straight quotes. I'll send a separate PR for that as I've got to do one for the ABI that will supersed this one.

smithp35:
Created PR https://github.com/ARM-software/abi-aa/pull/420 with straight quotes. ACLE PR https://github.com/ARM-software/acle/pull/456

PR #408, opened 2026-07-15
[sysvabi64] Correct MOV register in large code model example

sivan-shani:
In section 7.3.1, correct the example showing how to construct an absolute address using four MOV instructions. All four instructions must use the same destination register. The example incorrectly used a different register for each `MOVK` instruction, which would not have construct the address correctly.


sivan-shani:
@smithp35

smithp35 (approved):
LGTM, thanks for spotting the mistake.

Issue #399, opened 2026-06-16
Memtag ABI Extension to ELF violates idempotency requirement


EliaGeretto:
The "ELF for the Arm® 64-bit Architecture (AArch64)" states that:

> 5.7.19   Idempotency
> All RELA type relocations are idempotent. They may be reapplied to the place and the result will be the same. This allows a static linker to preserve > full relocation information for an image by converting all REL type relocations into RELA type relocations.

(https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#idempotency)

The extended `R_AARCH64_RELATIVE`, as defined in "Memtag ABI Extension to ELF", violates the idempotency principle because it uses the place `P` to store the tag offset. Once the relocation is performed the first time, that information is lost, so it cannot be reapplied:

> | ELF64 Code | Name               | Base Operation | MemtagABI Extended Operation  |
> |------------|--------------------|----------------|-------------------------------|
> | 1027       | R_AARCH64_RELATIVE | Delta + A      | LDG(Delta + A + *P) - *P      |
(https://github.com/ARM-software/abi-aa/blob/main/memtagabielf64/memtagabielf64.rst#relocation-operations)

The interest in preserving the idempotency principle comes from baremetal targets. The only way to deploy memory tagging on a self-relocating target is relocating once with zero tags to perform the initialization required for virtual memory, apply the tags, and then relocate again with randomized tags. If the idempotency is not preserved, it is necessary store a copy of the tag offsets somewhere else and then reuse them.

Unfortunately, at a first glance, it seems impossible to preserve backwards compatibility when the tag offset is not zero, unless the tag offset is stored "off band". It seems that one bit of information would be enough to tell the loader to "fetch the tag from the previous 16 bytes granule", if end pointers are the only concern. However, this single bit of information can only be stored using a new relocation type given that both the addend field and the place cannot be touched to preserve backwards compatibility and idempotency.

It is worth noting that the "PAuth ABI Extension to ELF" has a similar issue, since it uses the place `P` to store the signature schema. The main difference with Memtag relocations is that PAuth introduces new relocations types, which provide more flexibility. The proposed workaround there is to ask the platform to decide whether to sacrifice either half the addend or idempotency, by storing the schema in `P`. See #398. This proposed workaround, however, does not address directly the issue of a PAuth relocation that also requires a tag offset (points to one off the end of a tagged object). One of the bits in the PAuth schema could be used for this. This should probably be addressed in one, or both, the documents.

[...]

PR #410, opened 2026-08-06
Design document for idempotent storage of relocation metadata.

smithp35:
https://github.com/ARM-software/abi-aa/issues/399 describes an example [1] where a combination of the pauthabielf64 and memtagabielf64 cannot be resolved in a system that resolves dynamic relocations more than once.

This design document is an exporation of the design space of possible solutions to the problem. The two most serious contenders are:

* An additional ELF section containing metadata in a one to one correspondence with the relocations. This metadata section augments the existing ELF relocation sections.

* A compressed relative relocation format that encodes the metadata in the relocation section. This would replace the existing relocation sections for relocations requiring additional metadata.

For the ABI the recommendation is to add an additional ELF metadata section to augment the existing relocation sections. This is the most general, and could apply to self-relocating dynamic linkers as well as executables.

The compressed relative relocation format is only likely to be used by self-relocating executables, this means that it probably doesn't need to be fully defined in the full ABI at this point as it can be a private contract between tool and relocation resolver.

[1] In summary both the memtagabielf64 and pauthabielf 64 require metadata in the place of the relocation. When a system can use the place then the `reserved for addend` field in the signing-schema can be used for the memtagabielf64 tag offset. A system that resolves relocations more than once, and encodes the signing-schema in the relocation addend has no more bits.

### DIFF
--- a/design-documents/elf64-relocation-place-metadata.rst
+++ b/design-documents/elf64-relocation-place-metadata.rst
@@ -257,7 +257,7 @@ cannot handle the `MEMTAGABIELF64`_ extension of
 recover the correct Tag as the additional offset and the relocation
 addend can't both use the ``reserved for addend`` field.
 
-A possible mitigations for the conflict over ``reserved for addend``
+A possible mitigation for the conflict over ``reserved for addend``
 field is to restrict the values that an offset to recover the tag can
 take. In the majority of cases the access will be to the limit of an
 array. Using the example from `MEMTAGABIELF64`_:
--- a/design-documents/elf64-relocation-place-metadata.rst
+++ b/design-documents/elf64-relocation-place-metadata.rst
@@ -203,12 +203,16 @@ be achieved by:
 * Write initialization code that does not use signed code-pointers,
   such as calling virtual-functions.
 
-* When using `MEMTAGABIELF64`_ avoid using any access that requires
-  an additional offset prior to relocation-resolution.
-
 * Use ``SHT_RELR`` which exclude any R_AARCH64_relocation with
   metadata stored in the contents of the place ``*P``. These
   relocations can be resolved multiple times.
+
+* The `MEMTAGABIELF64`_ requires tagged globals to be accessed via the
+  GOT. Global Variables used prior to relocation-resolution must be
+  untagged.
+
+* Very few high-level diagnostics available for cases that will fail
+  at runtime.
 
 Pros:
 
@@ -306,7 +310,7 @@ Extending ELF with a relocation metadata section
 ================================================
 
 Introduce a new ELF section type, ``SHT_AARCH64_REL_METADATA`` that
-contains an array of ``Elf64_Word`` values. The ``sh_link`` field
+contains an array of values containing metadata. The ``sh_link`` field
 holds the section header index of the relocation section that it
 augments. Each value corresponds one to one with a relocation and
 appears in the same order as the relocation entry in the section
@@ -314,12 +318,15 @@ containing the relocations. Use of ``SHT_AARCH64_REL_METADATA`` is not
 required if a platform can use the contents of the place to store
 metadata.
 
+The size of the metadata value is by default an ``Elf64_Word`` as that
+matches the size of the contents of the place ``*P``.
+
 If a relocation section is augmented with a
 ``SHT_AARCH64_REL_METADATA`` section (type ``0x70000009``) and the
 relocation resolver opts to use it then the relocation Operations are
 altered as follows:
 
-* ``METADATA(idx)`` extracts the ``Elf64_Word`` value from the array
+* ``METADATA(idx)`` extracts the value from the array
   at ``idx``, where ``idx`` is the index of the relocation entry in
   linked relocation section.
 
@@ -332,6 +339,7 @@ from 0. The ``SHT_RELR`` section is not expected to have a
 ``SHT_AARCH64_REL_METADATA`` section as it contains dynamic
 relocations that do not require additional metadata (the
 `MEMTAGABIELF64`_ excludes entries that require additional metadata).
+
 
 ELF files may have up to 4 dynamic relocation sections, 3 of which may
 have a linked ``SHT_AARCH64_REL_METADATA`` section:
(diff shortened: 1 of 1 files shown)


design-documents/elf64-relocation-place-metadata.rst
```diff
@@ -0,0 +1,610 @@
...
+
+If the code that does steps 1 to 5 of the `Problem Statement`_ above
+does not use a location that requires a dynamic relocation with
+metadata then these relocations only need to be resoved once. This can
+be achieved by:
+
+* Write initialization code that avoids using the GOT, which may be
+  signed. For a self-relocating executable this can be done by using
+  hidden visibility, which allows the compiler to use PC-relative
+  addressing.
```
EliaGeretto, line 201: This is very hard with `memtag-globals`. All accesses to global variables need to be routed through the GOT because all of them need to use a tagged pointer. This essentially translates to no global variables in initialization code.

design-documents/elf64-relocation-place-metadata.rst
```diff
@@ -0,0 +1,610 @@
...
+
+* No changes to ABI or existing tools.
+
+Cons:
+
+* Requires programmer effort to create and maintain. This is
+  reasonable for a self-relocating dynamic loader which can be written
+  once and reused transparently by all programs on the
+  platform. However is it too much to expect from every
+  self-relocating bare-metal project?
```
EliaGeretto, line 227: The difficulty with this is also the enforcement of this restriction at compile time, expecially in a project that uses LTO, since relocations are visible only after linking.

design-documents/elf64-relocation-place-metadata.rst
```diff
@@ -0,0 +1,610 @@
...
+*n_relocations_of_metadata* is the number of relocations before the
+ next *relocation-code-specific-metadata*
+
+*address_or_bitmap* is the ``SHT_RELR`` and ``SHT_AUTH_RELR`` encoding
+ of the relocation offsets as described in the ELF specification under
+ `RELATIVE_RELOCATION_TABLE`_.
+
+Dynamic relocations described by
+``SHT_AARCH64_META_RELR`` can be sorted by relocation code, then by
+metadata to minimise the number metadata and relocation code changes.
```
EliaGeretto, line 583: Sorting them will make the grouping more efficient, but may compromise the effectiveness of the bitmaps used in the RELR encoding. I guess we would need statistics on a project that uses RELR already to figure this out.

design-documents/elf64-relocation-place-metadata.rst
```diff
@@ -0,0 +1,610 @@
...
+  a post-link transformation that overwrites the ``SHT_RELA``
+  relocations.
+
+* With the contents of the place ``*P`` containing the static-link
+  time address, a self-relocating bare-metal system does not need to
+  resolve any relocations prior to enabling PAC and/or MTE, or
+  self-relocating itself.
+
+* in most cases will be significantly smaller than ``SHT_RELA`` as
+  many locations use the same signing-schema.
```
EliaGeretto: This may not be the case (same signing-schema) with:
- `-fptrauth-function-pointer-type-discrimination` and similar.
- manually specified discriminators (string with variable name).
- tag offsets which currently depend on the size of the object. We could tell the compiler to prefer -1 though.

Compared to RELA, the statement is true because we are at least not repeating `r_addend`, which is in the place, and `r_info`, since we state the relocation code only once.

EliaGeretto:
I think we should also discuss how much space we should reserve for metadata. Technically, if we have a full 64 bits addend, Memtag requires a 64 bits tag offset to match it (60 if we define it as a number of tag granules). This means that, to avoid any restrictions, PAuth + Memtag requires 64 + 32 bits of metadata (trimmable removing reserved bits, but still more than 64). To avoid running out of space again and also not waste metadata space in Memtag-only or PAuth-only binaries, we may consider making the metadata size dependent on the relocation code for the compressed relocation solution. Alternatively, we could have a few bits in the header of whatever metadata section we end up picking which say how many 64 bits words the metadata field will contain. In this way, when we end up wanting to do PAuth + Memtag + something else, increasing that value is enough.

design-documents/elf64-relocation-place-metadata.rst
```diff
@@ -0,0 +1,610 @@
...
+
+  [
+    <uint32: relocation-code>
+    <uint32: n-relocations-of-code>
+    [ relocation-entry ]+
+  ]*
+
+*relocation-code* is one of three supported relocations
+ ``R_AARCH64_RELATIVE``, ``R_AARCH64_AUTH_RELATIVE`` and
+ ``R_AARCH64_NONE``. The ``R_AARCH64_NONE`` is used as an optional
```
EliaGeretto, line 551: We may need an additional code for the PAuth + tag offset case, if we want a full 64 bits tag offset to match the 64 bits addend.

design-documents/elf64-relocation-place-metadata.rst
smithp35: I'll add that as an additional complicating factor. I expect that it can be worked around with care, but it is yet another example where to write the initialisation code needs expertise in two ABI extensions.

design-documents/elf64-relocation-place-metadata.rst
smithp35: It is definitely fragile. I expect that to make the "take care" work will require the footprint of the code covered by the restrictions to be a small enough and stable enough for a small team of experts to write and maintain.

design-documents/elf64-relocation-place-metadata.rst
smithp35: One option I had in mind was to make the number of additional meta-data items dependent on the relocation code, so there could be any amount of additional metadata. That extension could be made at a later date.

Given the maximum size of a small code binary is 2 GiB a 32-bit field for an offset would be enough. Even with the large code model I think there are few non-contrived use cases for a 64-bit sized tag offset.

design-documents/elf64-relocation-place-metadata.rst
smithp35: It is possible, and yes ideally we'll need some data to check the heuristic.

My expectation is that with the current encoding, changing state will almost always be more expensive than adding an additional non-bitmap address field to an existing state. There's also the problem that changing state mid-bitmap may require a new non-bitmap address field anyway.

I think it would need a different approach to make changing state cheaper, for example declaring all the metadata once, and then using a smaller index into it during the description of relocations.

```
Relocation Code
metadata 1
metadata 2
metadata 3
metadata 4
Relocations for code
Use metadata 1
Address
Bitmap
Use metadata2
Bitmap (assuming no overlap with previous bitmap
```

design-documents/elf64-relocation-place-metadata.rst
smithp35: With the default signing-schema, my C++ test application ended with 50 unique signing-schemas out of 1500 relocations. I agree that if almost all relocations have different metadata then there will be better ways to encode. There's also a balance between how complex the encoding and decoding can be. For example, would using ULEB encoding like CREL, at the expense of non-4-byte aligned fields be acceptable?

smithp35:
> I think we should also discuss how much space we should reserve for metadata. Technically, if we have a full 64 bits addend, Memtag requires a 64 bits tag offset to match it (60 if we define it as a number of tag granules). This means that, to avoid any restrictions, PAuth + Memtag requires 64 + 32 bits of metadata (trimmable removing reserved bits, but still more than 64). To avoid running out of space again and also not waste metadata space in Memtag-only or PAuth-only binaries, we may consider making the metadata size dependent on the relocation code for the compressed relocation solution. Alternatively, we could have a few bits in the header of whatever metadata section we end up picking which say how many 64 bits words the metadata field will contain. In this way, when we end up wanting to do PAuth + Memtag + something else, increasing that value is enough.

There is a possibility of altering the sizes of the metadata entries, for dynamic linkers this could be done with a `DT_` tag, as well as setting `sh_entsize`. A statically linked program would need a linker defined symbol, which may end up being user-defined rather than being set by the linker. The size of the entries could affect the encoding:
* `(4 bytes, R_AARCH64_AUTH_RELATIVE)` PAuthABI metadata only (shift left by 32) to recover the original signing-schema.
* `(4 bytes, R_AARCH64_RELATIVE)` 32-bit tag offset
* `(8 bytes, R_AARCH64_AUTH_RELATIVE)`  = PAuthABI metadata with `reserved for addend` containing a 32-bit tag offset
* `(8 bytes, R_AARCH64_RELATIVE)` 64-bit tag offset
* `(16 bytes, R_AARCH64_AUTH_RELATIVE)` = 8-bytes of PAuthABI followed by 8-bytes of tag offset.
* `(16 bytes, R_AARCH64_RELATIVE)` = 8-bytes of zeros followed by 8-bytes of tag offset.

To keep it simple I only wrote up the 8-bytes case. I think an implementation could be extended to other two cases given some careful wording.

I'm not yet convinced that there's a non-contrived use case for 64-bit offsets at the moment. The small code-model already limits program size to 2GiB (32-bit offsets in exeception index tables). If I've understood the memtagabielf64 correctly it would need an extremely large array, or some kind of pointer-integer cast with some extremely large offsets.

design-documents/elf64-relocation-place-metadata.rst
EliaGeretto: That's good news. Is your C++ test a "real" project or is it artificial? It's fair to assume that with C++ inheritance you may end up with a lot of methods with the same signature and thus the same discriminator (I am not familiar with that part of the instrumentation). Do you think that transfers to C as well? I guess so if it is "object-oriented".

EliaGeretto:
> > I think we should also discuss how much space we should reserve for metadata. Technically, if we have a full 64 bits addend, Memtag requires a 64 bits tag offset to match it (60 if we define it as a number of tag granules). This means that, to avoid any restrictions, PAuth + Memtag requires 64 + 32 bits of metadata (trimmable removing reserved bits, but still more than 64). To avoid running out of space again and also not waste metadata space in Memtag-only or PAuth-only binaries, we may consider making the metadata size dependent on the relocation code for the compressed relocation solution. Alternatively, we could have a few bits in the header of whatever metadata section we end up picking which say how many 64 bits words the metadata field will contain. In this way, when we end up wanting to do PAuth + Memtag + something else, increasing that value is enough.
> 
> There is a possibility of altering the sizes of the metadata entries, for dynamic linkers this could be done with a `DT_` tag, as well as setting `sh_entsize`. A statically linked program would need a linker defined symbol, which may end up being user-defined rather than being set by the linker. The size of the entries could affect the encoding:
> 
> * `(4 bytes, R_AARCH64_AUTH_RELATIVE)` PAuthABI metadata only (shift left by 32) to recover the original signing-schema.
> * `(4 bytes, R_AARCH64_RELATIVE)` 32-bit tag offset
> * `(8 bytes, R_AARCH64_AUTH_RELATIVE)`  = PAuthABI metadata with `reserved for addend` containing a 32-bit tag offset
> * `(8 bytes, R_AARCH64_RELATIVE)` 64-bit tag offset
> * `(16 bytes, R_AARCH64_AUTH_RELATIVE)` = 8-bytes of PAuthABI followed by 8-bytes of tag offset.
> * `(16 bytes, R_AARCH64_RELATIVE)` = 8-bytes of zeros followed by 8-bytes of tag offset.
> 
> To keep it simple I only wrote up the 8-bytes case. I think an implementation could be extended to other two cases given some careful wording.
> 
> I'm not yet convinced that there's a non-contrived use case for 64-bit offsets at the moment. The small code-model already limits program size to 2GiB (32-bit offsets in exeception index tables). If I've understood the memtagabielf64 correctly it would need an extremely large array, or some kind of pointer-integer cast with some extremely large offsets.

As long as the new design leaves the possibility for future expansion open without forcing extremely complex solutions, we should be fine. What you are proposing sounds reasonable to me.

design-documents/elf64-relocation-place-metadata.rst
smithp35: Artificial; it is just a trivial program to test that I can throw and catch an exception. This drags in quite a lot libunwind (C) and libcxxabi (C++) in particular the itanium demangler for error messages which has quite a few virtual functions. In practice the vtables dominate the example, with a large .data.rel.ro, small .got and even smaller .data

I think C++ vtables would translate reasonably well to C function pointers with type discrimination enabled. Many function pointers will share the same type signature.

PR #412, opened 2026-08-10
[MEMTAGABIELF64] Rewrite LDG as a logical operation

### LINKED ISSUE #411: Memtag ABI: Clarify the semantics of `LDG(S)` in relocation processing
Section **8.4 Relocation Operations** could benefit from clarifying the semantics of the `LDG(S)` operation used in the extended relocation definitions.  In particular, it would be helpful to clarify:

- whether `LDG(S)` is an **abstract semantic operation** ("materialize the correct logical tag for `S`") or whether it normatively requires execution of the architectural `ldg` instruction;
    
- the expected observable behavior of `LDG(S)` when `S` refers to an allocation that is **not allocation-tagged** (i.e. whether the resulting pointer is expected to carry logical tag `0`).

The following wording is proposed for the definition of `LDG(pointer)`:

>**`LDG(pointer)` denotes the operation of materializing a pointer whose address equals `pointer` and whose logical tag equals the allocation tag associated with the allocation containing `pointer`. If the referenced allocation is not allocation-tagged the resulting logical tag is zero.** 

## Context and details

Section **8.4 Relocation Operations** extends the following relocation types:

| ELF64 Relocation | Base semantics | MemtagABI semantics |
| --- | --- | --- |
| `R_AARCH64_ABS64` | `S + A` | `LDG(S) + A` |
| `R_AARCH64_GLOB_DAT` | `S + A` | `LDG(S) + A` |
| `R_AARCH64_RELATIVE` | `Delta + A` | `LDG(Delta + A + *P) - *P` |

The specification also defines:

>*"`LDG(pointer)` is an instruction for the run-time environment to use the `ldg` instruction on pointer to materialize the correct logical tag for a symbol."*

It is not entirely clear whether `LDG(S)` is intended to be interpreted as an abstract semantic operation or as a normative requirement to execute the architectural `ldg` instruction.

These relocation types are also generated for function pointers. For example:

```cpp
using FnT = void();

extern "C" [[gnu::visibility("default")]] void defaultFunction() {}

constinit FnT* defaultFunctionPtr = &defaultFunction; // R_AARCH64_ABS64
```

In this example, the relocation symbol `S` is the address of `defaultFunction` in the executable `.text` section.

The [ARM Architecture Reference Manual](https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/LDG--Load-Allocation-Tag-) defines `ldg` as a load operation:

```
accdesc = CreateAccDescLDGSTG(MemOp_LOAD, stzgm, t);  
...  
tag = AArch64_MemTag(address, accdesc);
```

At the same time, [the architecture specifies](https://developer.arm.com/documentation/ddi0602/2026-03/Shared-Pseudocode/aarch64-functions-memory?lang=en#func_AArch64_MemTagRead_2) that reading the allocation tag of a non-allocation-tagged address yields allocation tag `0`:

```
if accdesc.tagaccess && memtagtype == MemTag_AllocationTagged then  
    (memstatus, tag) = PhysMemTagRead(memaddrdesc, accdesc);  
    return (tag, memaddrdesc, memstatus);  
...  
else  
    // Otherwise read the tag as zero  
    tag = '0000';  
    return (tag, memaddrdesc, memstatus);  
end;
```

The ambiguity arises because `ldg` cannot be executed on every symbol address. The instruction is defined as a load operation and therefore performs read permission checks. Consequently, executing `ldg` on an execute-only (XOM) mapping faults.

On the other hand, the library pseudocode for `ldg` specifies that reading the allocation tag of a non-allocation-tagged address returns allocation tag `0`. Therefore, for symbols known not to be allocation-tagged, the observable result of `LDG(S)` can be obtained without executing the `ldg` instruction.

It is currently in the spec implicitly, as the spec does not require to always match the `LDG(pointer)` operation to the `ldg` instruction.
However defining `LDG(S)` explicitly as a semantic operation, together with its observable behavior (including the result for non-allocation-tagged memory), would remove any possible ambiguity while allowing implementations to produce the required result in an equivalent manner.

smithp35:
The `LDG(pointer)` operation is defined in terms of the runtime environment using the `ldg` instruction. In most cases using the ldg instruction will work on allocation tagged memory and untagged memory. However if the untagged memory cannot be read, such as when that memory is execute-only then `ldg` cannot be used.

Rewrite `LDG(pointer) in terms of the underlying requirement which is that after the operation, `pointer` has the same logical tag as the allocation tag for *pointer. With memory that is not allocation tagged having a tag of 0000.

fixes: https://github.com/ARM-software/abi-aa/issues/411

### DIFF
--- a/memtagabielf64/memtagabielf64.rst
+++ b/memtagabielf64/memtagabielf64.rst
@@ -631,7 +631,7 @@ The relocations reference the following mnemonics:
 - ``LDG(pointer)`` returns ``pointer`` with the correct logical tag
   for the memory pointed to by ``pointer``. Memory without the
   architectural memory attribute ``MemTag_AllocationTagged`` has a
-  logcal tag of ``0000``. For memory that is readable by the run-time
+  logical tag of ``0000``. For memory that is readable by the run-time
   environment the operation is "as if" the ``ldg`` instruction
   operating on the pointer after it has been aligned down to the
   closest tag granule.
--- a/memtagabielf64/memtagabielf64.rst
+++ b/memtagabielf64/memtagabielf64.rst
@@ -628,13 +628,16 @@ non-MemtagABI binary or non-MTE device is the same as the non-extended variant.
 
 The relocations reference the following mnemonics:
 
-- ``LDG(pointer)`` is an instruction for the run-time environment to use the
-  ``ldg`` instruction on ``pointer`` to materialize the correct logical tag for
-  a symbol. This operation should also align the pointer down to the closest tag
-  granule before executing the ``ldg`` instruction.
+- ``LDG(pointer)`` returns ``pointer`` with the correct logical tag
+  for the memory pointed to by ``pointer``. Memory without the
+  architectural memory attribute ``MemTag_AllocationTagged`` has a
+  logcal tag of ``0000``. For memory that is readable by the run-time
+  environment the operation is "as if" the ``ldg`` instruction
+  operating on the pointer after it has been aligned down to the
+  closest tag granule.
 
-- For all the relocation types listed below, the loader should use the ``ldg``
-  instruction on the target address before writing into the target field, as the
+- For all the relocation types listed below, the loader should use the ``LDG``
+  operation on the target address before writing into the target field, as the
   target field may be inside of a tagged region.
 
 .. table:: Relocations with extended semantics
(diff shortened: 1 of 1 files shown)


memtagabielf64/memtagabielf64.rst
```diff
@@ -628,13 +628,16 @@ non-MemtagABI binary or non-MTE device is the same as the non-extended variant.
...
-  ``ldg`` instruction on ``pointer`` to materialize the correct logical tag for
-  a symbol. This operation should also align the pointer down to the closest tag
-  granule before executing the ``ldg`` instruction.
-
-- For all the relocation types listed below, the loader should use the ``ldg``
-  instruction on the target address before writing into the target field, as the
+- ``LDG(pointer)`` returns ``pointer`` with the correct logical tag
+  for the memory pointed to by ``pointer``. Memory without the
+  architectural memory attribute ``MemTag_AllocationTagged`` has a
+  logcal tag of ``0000``. For memory that is readable by the run-time
```
DanielKristofKiss:
```suggestion
  logical tag of ``0000``. For memory that is readable by the run-time
```

DanielKristofKiss (approved):
just a typo, otherwise LGTM

memtagabielf64/memtagabielf64.rst
smithp35: Thanks for the spot, will fix.

Issue #369, opened 2026-02-24
Remove ILP32 from AArch64 documents, replacemed with referall to previous releases for ILP32 content


smithp35:
ILP32 support for ELF platforms in the documents has been in perpetual Beta, with new additions to the ABI not considering ILP32. Now that the one complete implementation in GCC has been deprecated (https://gcc.gnu.org/gcc-15/changes.html) we will be removing ILP32 from the ABI documentation.

All documents that had ILP32 content will contain a statement that ILP32 support has been removed, please refer to a previous release of the document for ILP32 documentation.

smithp35:
All PRs associated with this issue now merged. Closing.

PR #413, opened 2026-08-13
[AAPCS64] Remove ILP32 Support

smithp35:
ILP32 support for ELF platforms in the documents has been in perpetual Beta, with new additions to the ABI not considering ILP32. Now that the one complete implementation in GCC has been deprecated (https://gcc.gnu.org/gcc-15/changes.html) we will be removing ILP32 from the ABI documentation.

Part of https://github.com/ARM-software/abi-aa/issues/369

### DIFF
--- a/aapcs64/aapcs64.rst
+++ b/aapcs64/aapcs64.rst
@@ -180,8 +180,6 @@ The following support level definitions are used by the Arm ABI specifications:
 
 Parts related to SME are at **Beta** release quality.
 
-The ILP32 variant is at **Beta** release quality.
-
 All other content in this document is at the **Release** quality level.
 
 Change history
@@ -274,6 +272,9 @@ changes to the content of the document for that release.
 +------------+--------------------+------------------------------------------------------------------+
 | 2025Q4     | 3\ :sup:`rd`       | - Standardize on "caller-saved" and "callee-saved".              |
 |            | March 2026         |                                                                  |
++------------+--------------------+------------------------------------------------------------------+
+| 2026Q3     | 13\ :sup:`th`      | - Remove ILP32 from the ABI. Refer to previous ABI releases for  |
+|            | August 2026        |   **Beta** ILP32 documentation.                                  |
 +------------+--------------------+------------------------------------------------------------------+
 
 References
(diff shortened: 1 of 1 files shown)


aapcs64/aapcs64.rst
```diff
@@ -833,7 +823,7 @@ Homogeneous Aggregates:
 The Base Procedure Call Standard
 ================================
 
-The base standard defines a machine-level calling standard for the A64 instruction set. It assumes the availability of the vector registers for passing floating-point and SIMD arguments. Application code is expected to conform to one of three data models defined in this standard; ILP32, LP64 or LLP64.
+The base standard defines a machine-level calling standard for the A64 instruction set. It assumes the availability of the vector registers for passing floating-point and SIMD arguments. Application code is expected to conform to one of three data models defined in this standard; LP64 or LLP64.
```
stuij:
```suggestion
The base standard defines a machine-level calling standard for the A64 instruction set. It assumes the availability of the vector registers for passing floating-point and SIMD arguments. Application code is expected to conform to one of two data models defined in this standard; LP64 or LLP64.
```

aapcs64/aapcs64.rst
stuij: We're not referencing this at all anymore except for the changelog and the note that we removed it which itself includes a quick summary, so remove here as well?

aapcs64/aapcs64.rst
```diff
@@ -486,6 +485,8 @@ This standard is presented in four sections that, after an introduction, specify
 
 This specification does not standardize the representation of publicly visible C++-language entities that are not also C language entities (these are described in `CPPABI64`_) and it places no requirements on the representation of language entities that are not visible across public interfaces.
 
+**(Beta)** suppport for ILP32 SysV-like variant where int, long int and pointer are 32-bit, has been removed. Refer to a previous binary release of the ABI for **(Beta)** ILP32 documentation.
```
stuij: I think describing the beta feature with beta markers is confusing here. At first I read it as removal of the feature being beta.

```suggestion
Beta suppport has been removed for the ILP32 SysV-like variant where int, long int and pointer are 32-bit. Refer to a previous binary release of the ABI for Beta ILP32 documentation.
```

TBH I would remove this line altogether. We already mention it in the change log and I feel we don't need to keep a changelog in the text of the things we removed.

stuij (changes requested):

aapcs64/aapcs64.rst
smithp35: I've added the details to the commit message, as I've removed the note.

aapcs64/aapcs64.rst
smithp35: I have removed it in favour of the changelog.

aapcs64/aapcs64.rst
smithp35: Thanks; adopted.

smithp35 (commented):
I've uploaded a patch with review comments addressed. Will update the other docs to match.

stuij (approved):
LGTM!

PR #415, opened 2026-08-13
[VFABI64] Remove ILP32 Support

smithp35:
ILP32 support for ELF platforms in the documents has been in perpetual Beta, with new additions to the ABI not considering ILP32. Now that the one complete implementation in GCC has been deprecated (https://gcc.gnu.org/gcc-15/changes.html) we will be removing ILP32 from the ABI documentation.

Part of https://github.com/ARM-software/abi-aa/issues/369


stuij:
ILP32 is still referenced in the glossary.

stuij (approved):
LGTM

PR #418, opened 2026-08-17
[AAELF64] Add section types from ELF extension documents

smithp35:
The PAuthABIELF64 and MemtagABIELF64 documents add additional ELF section types like SHT_AARCH64_AUTH_REL. Add these to the table of reserved section types.

The build attributes section has now been defined by buildattr64 add a reference in the table.

PR #419, opened 2026-08-17
 [AAELF64] Add relocation metadata section

### LINKED ISSUE #399: Memtag ABI Extension to ELF violates idempotency requirement

smithp35:
The PAuthABI and MemtagABI have relocations that encode metadata in the place of the relocation. For some self-relocating systems this can cause a problem as the code running prior to or during relocation may need to access to the partial application of that relocation. For example accessing a global variable, required to be in the GOT because it can be tagged.

Introduce an optional relocation metadata section containing an array of metadata entries that have a one to one correspondence to the relocation entries in its linked dynamic relocation section.

Fixes: https://github.com/ARM-software/abi-aa/issues/399


smithp35:
Includes PR #418 . This is a write up of the relocation metadata section mentioned in #410

PR #420, opened 2026-08-24
Standardise on straight quotes in CC-BY-SA-4.0 notices. [NFC]

smithp35:
The ABI and ACLE mix two different comment characters in several documents "CC-BY-SA-4.0”. Standardise on straight quotes "CC-BY-SA-4.0"


tommat01 (approved):
LGTM, thanks for picking this up.

PR #421, opened 2026-09-07
Fix typo in expansion of "GOT"

sivan-shani:
Replace "Global Object Table" with the correct term "Global Offset Table" in:

-  aaelf64
-  addenda32
-  bpabi32


smithp35 (approved):
LGTM. Thanks for spotting this.

PR #426, opened 2026-09-14
aaelf64: link relocation entries to their notes

sivan-shani:
Replace ambiguous “see notes below” text with named RST footnote references.


aaelf64/aaelf64.rst
```diff
@@ -1272,23 +1272,22 @@ The following tables record single instruction relocations and relocations that
   +------------+------------+------------------------------+------------+---------------------------------------------------------------------+
   | ELF64 Code | ELF32 Code | Name                         | Operation  | Comment                                                             |
   +============+============+==============================+============+=====================================================================+
-  | 287        | 22         | R\_<CLS>\_MOVW\_PREL\_G0     | S+A-P      | Set a MOV[NZ]immediate field to bits [15:0] of X (see notes below)  |
+  | 287        | 22         | R\_<CLS>\_MOVW\_PREL\_G0     | S+A-P      | Set a MOV[NZ]immediate field to bits [15:0] of X [#movw-prel]_      |
```
smithp35: In the text all we see is [1] which is not obvious it is a link to a note. Can we make this `( see note [#movw-prel]_)`? 

Same for all of the changes.

aaelf64/aaelf64.rst
```diff
@@ -1272,23 +1272,22 @@ The following tables record single instruction relocations and relocations that
...
   +------------+------------+------------------------------+------------+---------------------------------------------------------------------+
   | 292        | \-         | R\_<CLS>\_MOVW\_PREL\_G2\_NC | S+A-P      | Set a MOVK immediate field to bits [47:32] of X. No overflow check  |
   +------------+------------+------------------------------+------------+---------------------------------------------------------------------+
-  | 293        | \-         | R\_<CLS>\_MOVW\_PREL\_G3     | S+A-P      | Set a MOV[NZ]immediate value to bits [63:48] of X (see notes below) |
+  | 293        | \-         | R\_<CLS>\_MOVW\_PREL\_G3     | S+A-P      | Set a MOV[NZ]immediate value to bits [63:48] of X [#movw-prel]_     |
   +------------+------------+------------------------------+------------+---------------------------------------------------------------------+
 
-.. note::
-
+.. [#movw-prel]
```
smithp35, line 1296: If we are able to add text, it would be good to be able to see `note [1]` rather than just [1].

Not sure if that is possible.

smithp35 (commented):
I think if we are going to do this change, then it needs to be done to all the relocation tables so the document is stylistically consistent. I can see a few tables with the (see notes below) interleaved with your suggested change. Can you go through all the relocation tables?

aaelf64/aaelf64.rst
sivan-shani: Done

aaelf64/aaelf64.rst
sivan-shani: `note [1]`
Does not seems possible.
Changed to:
`[1] **Note:**`

smithp35 (approved):
Thanks for the update. That looks good to me, although I'd like to keep this open for a while to give other reviewers the opportunity to comment/object.

ilinpv:
> Before: before-table-line before-footnote After: after-table-line after-footnote

Thanks for the readability improvements! I think these numbered backlinks 1,2,3,4 can be a bit confusing. Could we use the relocation names they refer back to instead or some other text that makes it clear what each link points back to?

stuij:
Or remove the backlinks altogether if that's possible.

sivan-shani:
Commit updated to include named commits with no backlinks.
It seems as rst does not implement named backlinkes.

Example:
image

aaelf64/aaelf64.rst
```diff
@@ -1120,20 +1120,25 @@ The following tables record single instruction relocations and relocations that
   +------------+-----------------------------+------------+-----------------------------------------------------------------------------------------------------------------+
   | ELF64 Code | Name                        | Operation  | Comment                                                                                                         |
   +============+=============================+============+=================================================================================================================+
-  | 270        | R\_AARCH64\_MOVW\_SABS\_G0  | S + A      | Set a MOV[NZ] immediate field using bits [15:0] of X (see notes below); check -2\ :sup:`16` <= X < 2\ :sup:`16` |
+  | 270        | R\_AARCH64\_MOVW\_SABS\_G0  | S + A      | Set a MOV[NZ] immediate field using bits [15:0] of X (see `Signed MOVW`_);@ -2\ :sup:`16` <= X < 2\ :sup:`16`   |
```
smithp35: I think the `@` character in these 3 references isn't needed. It is showing up raw in the text. I think these are the only references with the @

smithp35 (commented):
Thanks for the updates. Even though it loses the back-references it should remove the confusion over the numbers.

Link to the rendered doc https://github.com/sivan-shani/abi-aa/blob/f7a8f1f613d62f669e168bad23707e030af01dc8/aaelf64/aaelf64.rst 

@stuij @ilinpv if Sivan's can you let us know if these changes are OK for you? Once the `@` characters have been removed I can approve.

aaelf64/aaelf64.rst
stuij: Yea this looks fine to me.

stuij: The way the links look in general I mean of course.

ilinpv:
Thanks @sivan-shani for the patch updates, named commits even without backlinks look good to me.

sivan-shani:
Redundant '@' removed, reviewers approve, 
@smithp35 Could you please merge? Thank you.

PR #427, opened 2026-09-18
sysvabi64: Fix typos and spell preemptible consistently

yury-khrustalev:
Just fixing a few typos. Also `preemptible` was written a few different ways: changed it to one consistent spelling.


sysvabi64/sysvabi64.rst
```diff
@@ -472,13 +472,13 @@ table (GOT).
 
 A GOT generating relocation is used to inform the static
 linker to create the GOT entry. The address of symbol definitions that
-cannot be pre-empted at dynamic link time can have their address
+cannot be preempted at dynamic link time can have their address
```
smithp35: pre-empted is the most common spelling in the remainder of the ABI, albeit that is mostly coming from the 32-bit ABI.

Looking through the differences, it seems like pre-empt is more British English and preempt more US English. Given that the Arm convention is to favour (he says using British English) the US spellings I think this is reasonable.

Sorry to ask, but could you update the spellings in the other documents so we're globally consistent. Can do that in a separate PR.

stuij:
Regarding pre-emptible and its siblings, in the ABI doc corpus as a whole, the overwhelming variant is with hyphen (87 to 6 apparently), so I would  convert the unhyphenated form to hyphenated.

smithp35:
> Regarding pre-emptible and its siblings, in the ABI doc corpus as a whole, the overwhelming variant is with hyphen (87 to 6 apparently), so I would convert the unhyphenated form to hyphenated.

I remember using the hyphenated form to match the existing ABI conventions. There is an argument that unhyphenated is more of a US English spelling, so more appropriate, however I think being consistent one way or the other across the ABI is preferable.

yury-khrustalev:
> Regarding pre-emptible and its siblings, in the ABI doc corpus as a whole, the overwhelming variant is with hyphen (87 to 6 apparently), so I would convert the unhyphenated form to hyphenated.

Thank, I've changed `pre-empt*` to `preempt*` across all .rst files. I hope it's more consistent now.

smithp35 (approved):
LGTM, thanks for updating all the spellings to be consistent. Please give some time before merging to see if there are any further comments from other reviewers.

